PR #1, opened 2015-12-11
[Editline] Redesign automatic indentation fixing for robustness

k8stone:
The FixIndentationCommand implementation has proven to be fragile across various libedit iterations. Reworked to use the same basic strategy as moving between multiple lines. When doing indentation changes are required the new approach exits line editing completely and restarts with amended content. This approach *should* prove to be more reliable.

Intended to address [SR-77](https://bugs.swift.org/browse/SR-77)

### DIFF
--- a/source/Host/common/Editline.cpp
+++ b/source/Host/common/Editline.cpp
@@ -863,37 +863,50 @@ Editline::NextLineCommand (int ch)
 unsigned char
 Editline::FixIndentationCommand (int ch)
 {
-    if (!m_fix_indentation_callback) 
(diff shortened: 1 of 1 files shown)


tfiala:
Hi Kate,

This code looks like it belongs in llvm.org lldb, svn trunk.  It is common code that we really should get upstream.  Can you please go through the standard llvm.org lldb submission process?  We can then cherry pick this into Swift LLDB to get it here quickly.

k8stone:
... and that's exactly how the process is intended to work.  Thanks!

PR #2, opened 2015-12-25
Addressing issues with 32-bit systems

hpux735:
There were a few minor issues with this codebase that prevented compilation on 32-bit systems.  These issues have been addressed.  Also, multiple warnings arising from the difference in size_t as used in printf were promoted to unsigned long's to make the %ul match the argument size.

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -265,7 +265,7 @@ public:
             return m_num_type_params;
         }
         
-        size_t
+        uint64_t
         GetNumWitnessesAtIndex (size_t idx) const
         {
             if (idx >= m_num_type_params)
--- a/packages/Python/lldbsuite/test/issue_verification/Makefile
+++ b/packages/Python/lldbsuite/test/issue_verification/Makefile
@@ -1,0 +1,4 @@
+LEVEL = ../make
+CXX_SOURCES := inline_rerun_inferior.cpp
+CXXFLAGS += -std=c++11
+include $(LEVEL)/Makefile.rules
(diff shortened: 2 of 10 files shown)


jasonmolenda:
Hi, thanks for the suggested changes.  The printf casts are a good idea, but please use (uint64_t) instead of (unsigned long) - I think a few of those lines being modified were already using (uint64_t) for some of their arguments.  Unsigned long may be different width on different machines / word sizes.

We shouldn't be changing LLDB_INVALID_ADDRESS - lldb stores addresses as 64-bits even when it is working with a 32-bit architecture.  We generally work fine when running on a 32-bit arm system, for instance, and lldb used to run on 32-bit i386 on mac (although it's been years since anyone has built that, I suspect).

jasonmolenda:
Looking these over, they're all in generic lldb code.  Could you please post the diffs (I'd recommend just the printf casts ;) over on lldb.llvm.org?  This github repository is for tracking the addition of swift code to the lldb source base -- we should make generic changes in lldb.llvm.org and then migrate them over to this repository.

hpux735:
Thanks for those suggestions.  I agree, I’ll make those changes.  There are a few cases where LLDB_INVALID_ADDRESS is being assigned to a 32-bit type. I’ll just have to changes those types, I assume.

jasonmolenda:
There's some sloppiness about using our 64-bit "not a value" value in 32-bit places -- definitely worth looking at.  There's a bunch of 32x64 mess down in debugserver and we ignore that - it's being used on mac/ios etc but eventually we plan to drop it and switch over to lldb-gdbserver.

jasonmolenda:
If those LLDB_INVALID_ADDRESS misuses are in Swift files, please do a pull request here on github.  But for any other parts of lldb, lldb.llvm.org is the place to go with.

hpux735:
Grr, that rebase totally messed it up...  I'll make a new PR.

tfiala:
Ah I see now the archeology of the flow of the patch change over to llvm.org.  Reading email in reverse time order is not my friend here :-)

We'll get that patch in llvm.org moving if it hasn't already gone in.  I had some questions on the casts but mostly because I wasn't in front of a computer to look at it more carefully.

PR #3, opened 2015-12-26
Fixed some issues on 32-bit systems

hpux735:
This is all that's needed in the swift-specific parts of LLDB.

### DIFF
--- a/source/Host/common/HostInfoBase.cpp
+++ b/source/Host/common/HostInfoBase.cpp
@@ -456,7 +456,7 @@ HostInfoBase::ComputeHostArchitectureSupport(ArchSpec &arch_32, ArchSpec &arch_6
 
     arch_32.Clear();
     arch_64.Clear();
-
+ 
     switch (triple.getArch())
     {
         default:
--- a/source/Host/common/HostInfoBase.cpp
+++ b/source/Host/common/HostInfoBase.cpp
@@ -456,7 +456,7 @@ HostInfoBase::ComputeHostArchitectureSupport(ArchSpec &arch_32, ArchSpec &arch_6
 
     arch_32.Clear();
     arch_64.Clear();
- 
+
     switch (triple.getArch())
     {
         default:
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -6234,7 +6234,7 @@ VisitNodeQualifiedArchetype (SwiftASTContext *ast,
     {
         swift::Demangle::Node::iterator end = cur_node->end();
         VisitNodeResult type_result;
-        uint64_t index = LLDB_INVALID_ADDRESS;
+        lldb::addr_t index = LLDB_INVALID_ADDRESS;
         for (swift::Demangle::Node::iterator pos = cur_node->begin(); pos != end; ++pos)
         {
             switch (pos->get()->getKind())
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -6234,7 +6234,7 @@ VisitNodeQualifiedArchetype (SwiftASTContext *ast,
     {
         swift::Demangle::Node::iterator end = cur_node->end();
         VisitNodeResult type_result;
-        size_t index = LLDB_INVALID_ADDRESS;
+        uint64_t index = LLDB_INVALID_ADDRESS;
         for (swift::Demangle::Node::iterator pos = cur_node->begin(); pos != end; ++pos)
         {
             switch (pos->get()->getKind())
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -265,7 +265,7 @@ public:
             return m_num_type_params;
         }
         
-        uint64_t
+        lldb::addr_t
         GetNumWitnessesAtIndex (size_t idx) const
         {
             if (idx >= m_num_type_params)
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -265,7 +265,7 @@ public:
             return m_num_type_params;
         }
         
-        size_t
+        uint64_t
         GetNumWitnessesAtIndex (size_t idx) const
         {
             if (idx >= m_num_type_params)


jasonmolenda:
Looks good to me.  Most of the swift lldb team is away on vacation for the next week, let's let them give it the final approval when they're back - but I think these changes are good.

hpux735:
Thanks, Jason. :)

tfiala:
@clayborg, can you have a look?

I think the arm spec additions need to happen in llvm.org lldb and pulled through there.  We shouldn't be taking them in swift-lldb.

tfiala:
After looking a bit more at cd4416a and 634bd9d, we definitely want those considered at the llvm.org lldb svn trunk branch.  Most of swift-lldb is llvm.org lldb, with just the swift-specific bits added to the github swift-lldb repository.   (We cover more details on this [here](https://swift.org/contributing/#llvm-and-swift).)

@hpux735 and/or @iachievedit, can you put up patches for cd4416a and 634bd9d against llvm.org lldb?  Those can then be pulled through to swift-lldb once they've been accepted, and we can adjust this here.

In the interim, the original small change here could be broken back out again and looked fine (the original change specified in e7c16daf9d2f92c00d3d609f7e03b02ddc9ca686).  In general we prefer to take small changes to larger ones.  I'd be happy to pull that in if we get back to it.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -6234,7 +6234,7 @@ VisitNodeQualifiedArchetype (SwiftASTContext *ast,
     {
         swift::Demangle::Node::iterator end = cur_node->end();
         VisitNodeResult type_result;
-        size_t index = LLDB_INVALID_ADDRESS;
+        uint64_t index = LLDB_INVALID_ADDRESS;
```
clayborg: This should be "lldb::addr_t" instead of "uint64_t" if we are initializing this with LLDB_INVALID_ADDRESS.

clayborg: As Todd Fiala stated, this is all generic stuff that should go directly into llvm.org, then it will be merged into swift-lldb by the auto merger.

hpux735:
Unfortunately, after looking at it more closely, I think that there is no way around including v6l and v7l within LLDB itself.  Swift is using ```armv6-unknown-linux-gnueabihf``` as the target when it launches LLDB, and LLDB itself is coming up with amrv6l on its own.  Therefore, for the architecture logic to succeed it has to be here.  That said, I know that the changes need to go to the lldb.llvm.org repo.  In the mean time, it's a lot easier to review code here, rather than on the mailing list, so what comments are there relating to the arm architectures.  Specifically, is there a reference for the meaning of field three (sub) in the ```ArchDefinitionEntry``` structure? They're just uint32's, but I don't see where they map to something meaningful, and many of them are repeated.  My instinct is to have v6l and v7l map to a copy of v6 and v7 respectively.

jasonmolenda:
I think what you're seeing are the CPU subtype enumerations used on Darwin systems.  The problem is that there are lots of things called "armv7" that may have incompatible instruction sets, so we use the subtype to distinguish.  For instance, armv7 refers to a generic ARMV7-A ISA, armv7s is used for one of the Swift architecture (different from swift language) cores where we have an idiv instruction and a couple others.  Other examples are the low-power thumb-instruction-only Cortex M0-4 chips (the "M-profile", or "microcontroller profile" chips).  Cortex-M0 ("ARMv6-M") is armv6m, Cortex-M3 ("ARMv7-M") is armv7m, Cortex-M4 ("ARMv7e-M") is armv7em.

It's been years since I worked with an armv6 core - I think it lacks NEON / VFP instructions, and it doesn't have thumb?  Or maybe it has thumb (T16) but not thumb-2 (T32).  It's definitely a different architecture - you wouldn't want to jit armv7 code and try to execute it on an armv6 cpu.

lldb has an idea of "compatible architectures".  An armv7s system can run armv7 code -- armv7s is a superset of armv7's ISA, so if you're running on an armv7s cpu and the user tries to load an armv7 solib, that is fine.  On the other hand, if you're running on an armv7 cpu, you must disassemble with armv7, you must jit for armv7, you ignore any binaries with armv7s cpu subtypes.

I'm not sure what the armv7l and armv7l cpu subtypes would be meant to indicate.  I don't know a lot about the ARM product map outside what is used at apple, so this could just be my ignorance.  I don't think you generally want to consider armv6 and armv7 as compatible ...I suppose an armv7 process running on an armv7 core may be able to load an armv6 solib if the ABI are compatible, I don't really remember the details.

Anyway, tl;dr: "armv7" means many different ISAs, and we use the cpu subtype to disambiguate between them & ensure we don't run illegal code or disassemble instructions incorrectly.

jasonmolenda:
BTW apologies if this is all obvious already - just wanted to jot it down so we're both using the same terms.

jasonmolenda:
One last thought.  Adding cpu subtypes is a real pain, I would avoid it unless you really need to distinguish between different ISA subsets / features that are being used.

clayborg:
Some file formats (ELF, MachO) have differing ways of specifying which architecture is contained within a file. In ELF we use the e_machine from the ELF header. This isn't much help as all it tells us is EM_ARM. Thanks for the help ELF... So then we have to root through the ELF notes looking for table scraps to help us identify the actual ARM architecture. In MachO we have a CPU type that says "CPU_TYPE_ARM", but then we have a subtype that helps us determine which actual ARM variant it is in the CPU subtype field. This makes it very easy for us to decode what the architecture is. We use a table of ArchDefinitionEntry objects in ArchSpec.cpp to help us fill in the ArchSpec object from a set of ELF or MachO defines. 

So entries for any architecture you are going to specify by name must be added to the main table (g_core_definitions in ArchSpec.cpp), but you must also provide ways for the ObjectFileMachO and ObjectFileELF to correctly take an executable and extract a compatible ArchSpec from the binary file (fill in the ArchDefinitionEntry tables as needed).

More comments below.

> On Jan 4, 2016, at 6:09 PM, William Dillon <notifications@github.com> wrote:
> 
> Unfortunately, after looking at it more closely, I think that there is no way around including v6l and v7l within LLDB itself. Swift is using armv6-unknown-linux-gnueabihf as the target when it launches LLDB, and LLDB itself is coming up with amrv6l on its own. Therefore, for the architecture logic to succeed it has to be here. That said, I know that the changes need to go to the lldb.llvm.org repo. In the mean time, it's a lot easier to review code here, rather than on the mailing list, so what comments are there relating to the arm architectures. Specifically, is there a reference for the meaning of field three (sub) in the ArchDefinitionEntry structure? They're just uint32's, but I don't see where they map to something meaningful, and many of them are repeated. 

They are solely there to help you take an object file and extract a ArchSpec given an object file (ELF, MachO, COFF). Since each object file format has its own defines that describe the architecture in the executable  (ELF has e_machine in the ELF header + notes, MachO has the CPU type and subtype, COFF has just a machine type somewhere), the ArchDefinitionEntry is designed to allow each file format make these mean what they need to. We can add fields if necessary. ArchDefinitionEntry is currently tailored to MachO and ELF, but can be expanded if needed.

> My instinct is to have v6l and v7l map to a copy of v6 and v7 respectively.

So can you explain what v6l and v7l are?

hpux735:
Armv6l and Armv7l are a linux-ism for armv7(a), little endian, for example.  There is a debate currently at the llvm-dev mailing list relating to this issue and my patch there.  I've been searching for the exact place where LLDB actually gets armv7l to see if I can shunt it away there, but haven't had any luck yet.  At a time, source/Plugins/Platform/Linux/PlatformLinux.cpp:445 seemed promising, but I don't think it is any more.  It appears that LLDB is getting the process triple from LLVM, and therefore, appropriate changes to LLVM would fix these issues as well. source/Host/common/HostInfoBase.cpp:455 is where that happens.

clayborg:
Then you should only need to add entries into the "g_core_definitions" for these. 

Any object files that run into armv7 and armv6 binaries will never resolve to the armv7l or armv6l since they should just be armv7 and armv6 respectively, but we will add them to the table in case anyone specifies them by name with any "--arch" options. And Yes LLVM would need to be modified to accept triples with these in them if it doesn't already.

Again, why are these needed? Is someone trying to specify them at the command line? Seems like we should just use "armv6" or "armv7".

hpux735:
The solution I just pushed works well on ARMv7l, and v6l is compiling currently (It's a 1-2 day process).  The theory of operation is that the Triple constructor in LLVM preserves the string that it's provided unless it's changed using an accessor method.  Additionally, LLVM maps armv6l and armv7l to armv6k and armv7 respectively (see https://github.com/hpux735/swift-llvm/blob/stable/lib/Support/Triple.cpp#L506 ).  Therefore, if v6l or v7l is provided the subarch will be matched to v6k or v7 and re-set to the same, which will remove the trailing 'l' while maintaining all the semantic meaning and current architecture tables.

hpux735:
Ok, I just removed the code that needs to go to llvm.org, and addressed clayborg's suggestions.  I hope that this is fairly non controversial at this point. :)

source/Host/common/HostInfoBase.cpp
```diff
@@ -456,7 +456,7 @@ HostInfoBase::ComputeHostArchitectureSupport(ArchSpec &arch_32, ArchSpec &arch_6
 
     arch_32.Clear();
     arch_64.Clear();
-
+
```
tfiala: Can we get rid of this whitespace change?

tfiala:
Hi William,

If you can get rid of the pure whitespace change, I'll merge this in.

Thanks!

hpux735:
Fixed it.  Thanks, Todd!

tfiala:
Thanks for addressing this, William!  Merged.

PR #5, opened 2016-01-28
[SR-88] PlaygroundQuickLook tests now test 'reflecting:' initializer,…

austinzheng:
… not custom quick look functionality

@dabrahams Let me know if this is what you had in mind. I made sure each expression evaluated correctly in the REPL.


austinzheng:
The original code had a fifth `struct S { }` test case, which was intended to return `nil`. The new API would return the text `"REPL.S()"`; however, I haven't yet been able to figure out what the name of the test module is. As well, this would be testing something different than the original impl (whether or not the `_Mirror` API properly indicated that no quicklook implementation existed). Let me know if this is something you feel strongly about one way or another.

dabrahams:
@austinzheng yes, that’s what I had in mind. @egranata could someone from the lldb team please confirm this looks right and merge if so?

austinzheng:
Since it's in, I think this PR can be closed. Many thanks!

PR #13, opened 2016-04-08
fix swig binding generation to pass through #ifdef  __APPLE__ in typemap

tfiala:
Addresses the underlying cause of:
https://bugs.swift.org/browse/SR-1192

and prevents future static Python bindings to break on environments where __APPLE__ is not defined.


tfiala:
@swift-ci Please test

tfiala:
This was primarily a non-Apple issue.  CI for Linux completed.  Merging in.

PR #17, opened 2016-04-12
fix git clone command line by reverting upstream patch

tfiala:
I reviewed a change up in LLVM.org LLDB to support checking out git faster with a --depth=1 flag.  This change had bad ramifications in GitHub swift-lldb, where checking out LLVM/clang/swift/etc. through this script for normal workflow is commonplace.  The depth-limited checkout is not a useful style of repository for those longer-lived use cases.  This change reverts that commit.


tfiala:

PR #18, opened 2016-04-13
temporarily increase the test suite timeout

tfiala:
This is a temporarily measure until the following issue is resolved:


tfiala:

PR #19, opened 2016-04-14
fix a race is the LLDB test suite results collection

tfiala:
The race boiled down to this:

If a test worker queue is able to run the test inferior and
clean up before the dosep.py listener socket is spun up, and
the worker queue is the last one (as would be the case when
there's only one test rerunning in the rerun queue), then
the test suite will exit the main loop before having a chance
to process any test events coming from the test inferior or
the worker queue job control.

I found this race to be far more likely on fast hardware.
Our Linux CI is one such example.

Addresses an issue (maybe the only issue) causing the
following bug:


tfiala:

PR #20, opened 2016-04-17
normalize test file extension for test filenames

tfiala:
Something changed so that a test's filename can
report as the .pyc rather than the .py file in some flows.  This
causes confusion, particularly in cases where a test runner
queue has to synthesize a test event, which uses the .py extension that everything was previously using.  Then the rerun result event has a different
key than the timeout test run, and the timeout test run does
not get overwritten.

This change introduces a normalization step that converts all
test filenames ending in .pyc to end in .py.

Fixes:

### DIFF
--- a/packages/Python/lldbsuite/test/result_formatter.py
+++ b/packages/Python/lldbsuite/test/result_formatter.py
@@ -64,7 +64,7 @@ def create_results_formatter(config):
     def create_socket(port):
         """Creates a socket to the localhost on the given port.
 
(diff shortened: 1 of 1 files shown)


tfiala:

tfiala:
Merging this in - this is a fix for the test running infrastructure.  I hit the same failing tests locally, but these are now due to the Swift #line/#sourceLocation change.  I don't blame anyone for not catching that given all the noise that the LLDB test suite was generating.

PR #23, opened 2016-04-27
Sketch data formatters for new range types, introduced in SE-0065 "A New Model for Collections and Indices"

gribozavr:


gribozavr:
@egranata @tfiala Could you review, please?

tfiala:
Looks reasonable to me.  This is Enrico's code, so his opinion counts far more.

Let's give it a test run as well.

tfiala:
@swift-ci Please test

gribozavr:
This code depends on https://github.com/apple/swift/pull/2108, so it won't work as-is.

tfiala:
Okay this build looks like it was stuck behind the triple breakage.

tfiala:
> This code depends on apple/swift#2108, so it won't work as-is.

I see, okay.

egranata:
I am not familiar with the details of the new collection model, but the code looks very reasonable - assuming it passes our test suite whenever the main change gets pulled in, this looks good to me.

PR #25, opened 2016-04-29
Update MemoryReader::readBytes API to return int

bitjammer:
This function needs to be compatible with C programs, so we can't
use the bool return type.


bitjammer:
This PR is paired with https://github.com/apple/swift/pull/2335.

bitjammer:
@egranata can you review please?

bitjammer:
Never mind! I'll just make the change on the C side.

PR #28, opened 2016-05-31
Add gmodules testing

tfiala:
Cherry-pick LLVM.org LLDB svn trunk r270848 to enable gmodules testing.

This change also xfails a few tests in GitHub that are not failing upstream.  Radars track those issues.

### DIFF
--- a/packages/Python/lldbsuite/support/gmodules.py
+++ b/packages/Python/lldbsuite/support/gmodules.py
@@ -1,0 +1,30 @@
+from __future__ import absolute_import
+from __future__ import print_function
+
+# System modules
+import os
+import re
+
+
+GMODULES_SUPPORT_MAP = {}
+GMODULES_HELP_REGEX = re.compile(r"\s-gmodules\s")
+
+
+def is_compiler_clang_with_gmodules(compiler_path):
+    # Before computing the result, check if we already have it cached.
+    if compiler_path in GMODULES_SUPPORT_MAP:
+        return GMODULES_SUPPORT_MAP[compiler_path]
+
+    def _gmodules_supported_internal():
+        compiler = os.path.basename(compiler_path)
+        if "clang" not in compiler:
+            return False
+        else:
+            # Check the compiler help for the -gmodules option.
+            clang_help = os.popen("%s --help" % compiler_path).read()
+            return GMODULES_HELP_REGEX.search(clang_help, re.DOTALL) is not None
+
+    GMODULES_SUPPORT_MAP[compiler_path] = _gmodules_supported_internal()
+    return GMODULES_SUPPORT_MAP[compiler_path]
+
+
--- a/packages/Python/lldbsuite/test/decorators.py
+++ b/packages/Python/lldbsuite/test/decorators.py
@@ -543,15 +543,3 @@ def skipUnlessThreadSanitizer(func):
             return "Compiler cannot compile with -fsanitize=thread"
         return None
     return skipTestIfFn(is_compiler_clang_with_thread_sanitizer)(func)
-
-def skipUnlessClangModules():
-    """Decorate the item to skip test unless Clang -gmodules flag is supported."""
-    def is_compiler_clang_with_gmodules(self):
-        compiler_path = self.getCompiler()
-        compiler = os.path.basename(compiler_path)
-        if compiler != "clang":
-            return "Test requires clang as compiler"
-        clang_help = os.popen("%s --help" % (compiler_path)).read()
-        match = re.match(".* -gmodules ", clang_help, re.DOTALL)
-        return "Clang version doesn't support -gmodules flag" if not match else None
-    return skipTestIfFn(is_compiler_clang_with_gmodules)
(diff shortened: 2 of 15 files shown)


tfiala:
@swift-ci Please test

Geforce8472:
Not sure why but this seems to break my swift build. This could be because it seems to be calling g++ rather than clang.

```
os command: make clean
with pid: 21153
stdout: rm -f  main.o main.d main.d.tmp main.dwo
rm -f "a.out"

stderr:
retcode: 0



os command: make MAKE_DSYM=NO MAKE_GMODULES=YES ARCH=x86_64  CC=/home/git/Swift/build/buildbot_linux/llvm-linux-x86_64/bin/clang SWIFTCC=/home/git/Swift/build/buildbot_linux/swift-linux-x86_64/bin/swiftc
with pid: 21170
stdout: g++  -std=c++11  -g -O0 -fno-builtin -m64   -I/home/git/Swift/lldb/packages/Python/lldbsuite/test/make/../../../../../include -include /home/git/Swift/lldb/packages/Python/lldbsuite/test/make/test_common.h   -gmodules -c -o main.o main.cpp
../../../make/Makefile.rules:595: recipe for target 'main.o' failed

stderr: cc1plus: error: unrecognised debug output level ‘modules’
cc1plus: error: unrecognised debug output level ‘modules’
make: *** [main.o] Error 1

retcode: 2


ERROR

Traceback (most recent call last):
  File "/home/git/Swift/lldb/packages/Python/lldbsuite/test/lldbinline.py", line 153, in __test_with_gmodules
    self.buildGModules()
  File "/home/git/Swift/lldb/packages/Python/lldbsuite/test/lldbtest.py", line 1340, in buildGModules
    if not module.buildGModules(self, architecture, compiler, dictionary, clean):
  File "/home/git/Swift/lldb/packages/Python/lldbsuite/test/plugins/builder_base.py", line 166, in buildGModules
    lldbtest.system(commands, sender=sender)
  File "/home/git/Swift/lldb/packages/Python/lldbsuite/test/lldbtest.py", line 431, in system
    raise cpe
CalledProcessError: Command 'make MAKE_DSYM=NO MAKE_GMODULES=YES ARCH=x86_64  CC=/home/git/Swift/build/buildbot_linux/llvm-linux-x86_64/bin/clang SWIFTCC=/home/git/Swift/build/buildbot_linux/swift-linux-x86_64/bin/swiftc ' returned non-zero exit status 2
Config=x86_64-/home/git/Swift/build/buildbot_linux/llvm-linux-x86_64/bin/clang-3.9
Session info generated @ Fri Jun  3 16:36:36 2016
To rerun this test, issue the following command from the 'test' directory:

./dotest.py -A x86_64 -C /home/git/Swift/build/buildbot_linux/llvm-linux-x86_64/bin/clang-3.9 -v -t -N dsym lang/cpp/extern_c
```

Any idea how to fix this? I'm happy to help.

tfiala:
Hello!

What is the full command line you’re using to run the tests?  (And if you have any build-related environment variables set, what are those?).

There is a check against the C compiler (I believe) that scans its help output to see if it professes to support -gmodules as an option.  If the C++ compiler you’re using is not the matching one for the C compiler (e.g. If you’re using clang for C but g++ for C++), I could see that tripping things up.  The LLDB test suite assumes it can figure out the right C++ compiler just from knowing the C compiler.  (In fact, looking at your make line, it looks like you’re dealing with that scenario).

Is there a particular reason you’re not wanting to use clang++ for C++ when you’re using clang (CC=/home/git/Swift/build/buildbot_linux/llvm-linux-x86_64/bin/clang) for the C side?  The test suite currently isn’t designed to deal with that type of split setup.  It will assume the corresponding C++ compiler for the given C compiler.

You might not be attempting to make this be the case, in which case it sounds like something in the determination of the C++ compiler is failing.  You can try to rule that out by doing something like this:

cd /your/lldb/src/root
python test/dotest.py —executable /path/to/your/just/built/lldb -C /home/git/Swift/build/buildbot_linux/llvm-linux-x86_64/bin/clang —swift-compiler /home/git/Swift/build/buildbot_linux/swift-linux-x86_64/bin/swiftc

And see if that works.  If it does, that implies something about the environment or the configuration on that bot is confusing the normal C++ compiler determination logic.

Anything wrong with that, btw, would need to go over to LLVM.org <http://llvm.org/> LLDB, as this isn’t specific to Swift.

Let me know what you find!

Thanks,
Todd

tfiala:
I’m able to reproduce this over here.

Something does appear to be busted with the c++-finding code.  It’s incorrectly choosing g++ on Ubuntu (at least, on 14.04).

I’ll look into this.

-Todd

tfiala:
I think that issue is fixed here now:

commit 947486c66b7ca3a8649650c846cede7009d9a65e (HEAD, origin/master, origin/HEAD, master)
Author: Todd Fiala <tfiala@apple.com>
Date:   Fri Jun 3 11:35:35 2016

    fixup Makefile to match source files in the test

A test was converted from C to C++, but the Makefile was not adjusted properly.  This caused some built-in rules to get invoked, which caused the wrong C++ compiler to be selected, manifesting in the issue I was reproducing.

Let me know if you still see the issue after that change makes it to your bot.

-Todd

Geforce8472:
Sorry for the slow response, I was away over the weekend. That commit seems to have fixed the issue for me as well. Thanks for your help.

tfiala:
Glad to hear it!

-Todd

PR #29, opened 2016-06-13
prepend XFAIL/XPASS to the xUnit-reported test method name

tfiala:
If the xUnit formatter is instructed to mark XFAIL/XPASS results as
xUnit successes, then prepend XFAIL: or XPASS: to the method name.

Prior to this change and the corresponding change in
swift/utils/build-script-impl to mark XPASS and XFAIL results
as successes, the pretty-ified output from the Jenkins XUnit
plugin did not know anything about XFAILs or XPASSes.
With these two changes, we will see XPASS and XFAIL results
as "passing" in the xUnit results, but we'll be able to distinguish
them by the prefix.

rdar://26769429


tfiala:
@swift-ci Please test

tfiala:
I'm seeing the behavior I was hoping to see on the Linux CI build.  Bringing this in now.

PR #35, opened 2016-07-21
Update test for for SE-0036

ahoppen:
Updates a test for the changes in [SE-0036](https://github.com/apple/swift-evolution/blob/master/proposals/0036-enum-dot.md).


tfiala:
Thanks, Doug!

PR #36, opened 2016-07-21
educate Linux LLDB how to find the resource dir

tfiala:
Non-Apple platforms would fail to find the Swift resource dir when
running LLDB from the build dir.  This issue was not present when
running LLDB from the install dir.

This change makes one more fallback attempt to synthesize the
Swift resource dir by doing what would be a valid regex-based
manipulation of the LLDB lib path for a build-script-driven
style build.  That style build is the prevailing build style for
everywhere except for the Xcode-driven LLDB build.

This change also restores running the LLDB Swift REPL tests
on Linux that run from the build layout.

Fixes:
bugs.swift.org/SR-1300
bugs.swift.org/SR-2136

### DIFF
--- a/packages/Python/lldbsuite/test/lldbrepl.py
+++ b/packages/Python/lldbsuite/test/lldbrepl.py
@@ -27,7 +27,6 @@ class REPLTest(PExpectTest):
     
     @swiftTest
     @no_debug_info_test
-    @skipIfLinux  # https://bugs.swift.org/browse/SR-1300
     def testREPL(self):
         # setup the regexp for the prompt
         self.prompt = re.compile('\\d+>')
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -15,6 +15,7 @@
 // C++ Includes
 #include <mutex> // std::once
 #include <queue>
+#include <regex>
 #include <set>
 
 #include "clang/Basic/TargetInfo.h"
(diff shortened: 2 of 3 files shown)


tfiala:
@swift-ci Please test

tfiala:
This change also adds a ton more logging on the '(lldb) enable lldb types' log channel regarding the process that searches for the Swift resource dir.

tfiala:
Our master-next LLDB PR tester bot looks totally busted.  That build didn't even do the build phase.

I'll need to get with Mishal on that when he's back.

Given that, I'm going to merge this into master-next and watch the bots there.

tfiala:
The solution here worked fine for Ubuntu 15.10 and Ubuntu 16.04; however, it looks like a potentially busted C++ stdlib and or compiler on Ubuntu 14.04 wasn't finding/evaluating the call to std::regex_replace() using the const char* variant.  I don't know if that variant --- which the docs claim is in C++11 --- is just missing in the C++ stdlib on Ubuntu 14.04 or if the template selection logic is faulty.  But in any event, the template signature it tried to match against was the std::string variant.

I just committed a change which forces the use of the std::string variant, which hopefully fixes the Ubuntu 14.04 issue (which I'll know shortly).

Once that clears, I'll be taking this change into swift-lldb/master and the current Swift-3 preview branch.

tfiala:
Okay that wasn't good enough.  It builds now, but on Ubuntu 14.04 x86_64 the std::regex engine is throwing a regex exception where Ubuntu 15.10 and 16.04 are not.  Grrr, this is supposed to be a *standard* library with consistent results...

So it looks like I may end up manually implementing the little bit I need to parse because Ubuntu 14.04's C++ stdlib can't do regex replace and replace this code here.

PR #38, opened 2016-07-29
[Do not merge yet][SE-0101] Migration

rintaro:
Depends on https://github.com/apple/swift/pull/3854

[SE-0101](https://github.com/apple/swift-evolution/blob/master/proposals/0101-standardizing-sizeof-naming.md) removes `sizeof`, `sizeofValue` and related functions from the stdlib.
This PR is migrating them to newly introduced `MemoryLayout<T>` values.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/private_var/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/private_var/main.swift
@@ -13,7 +13,7 @@
 private var a = 1
 
 func doSomething(b: Int) {
-	a += strideof(b.dynamicType) //% self.expect("expr a", substrs=['Int', '= 1'])
+	a += MemoryLayout.of(b).stride //% self.expect("expr a", substrs=['Int', '= 1'])
 }
 
 doSomething(b:2)


packages/Python/lldbsuite/test/lang/swift/private_var/main.swift
```diff
@@ -13,7 +13,7 @@
 private var a = 1
 
 func doSomething(b: Int) {
-	a += strideof(b.dynamicType) //% self.expect("expr a", substrs=['Int', '= 1'])
+	a += MemoryLayout.of(b).stride //% self.expect("expr a", substrs=['Int', '= 1'])
```
rintaro: Actually this `stride` value means nothing here.
We might want to use simpler expression? e.g. `a += 42`

gribozavr: I asked Jim and Sean about this test, and they suggested that we use `a += b`, as long as the test continues to pass.

gribozavr:
Continuing in #39.

PR #45, opened 2016-08-27
Don't crash when trying to capture persistent variables in a block.

scallanan:
Don't crash when trying to capture persistent variables in a block.

Reports an error instead.  We can fix this later to make persistent variables work, but right now we hit an LLVM assertion if we get this wrong.  This patch also makes the error stream easier to use, and ensures it is never NULL.  It also fixes some of the logic in ClangExpressionParser to allow expressions to error out if IRForTarget fails.

This fixes <rdar://problem/27770298>


scallanan:
@jimingham Please review

jimingham:
This looks fine to me.

PR #49, opened 2016-09-06
Remove unnecessary use of ArchetypeType::isSelfDerived()

slavapestov:


slavapestov:
@egranata Does this look reasonable?

slavapestov:
@swift-ci Please test

egranata:
@slavapestov Assuming that true is the right answer for the question isSelfDerived() was asking before, then, yeah, sure.

Keep in mind that we are undergoing a huge reformatting of LLDB today, so you may want to hold off on committing this; @tfiala might have a more exact timeline for you

tfiala:
I don't think there's an issue committing on swift-lldb/master.  We've done what I expect to be the final merge from swift-lldb/master => swift-lldb/master-next before the Great Code Formatting tsunami washes over us (likely tomorrow).

tfiala:
The failure on the Linux build is a non-issue.  (Timeout accrued in no method, so rerun of timed out file isn't clearing the error properly.  There's a bug on that).

This should be fine to go in.

PR #50, opened 2016-09-13
Update use of swift::GenericDWARFVersion for swift API change.

adrian-prantl:
rdar://problem/28231982


adrian-prantl:
https://github.com/apple/swift/pull/4697

PR #51, opened 2016-09-13
Swift 3.0 branch

egranata:
Add VALID_OR_RETURN checks across the boards. Fixes rdar://27902313

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -282,6 +282,8 @@ GetEnumInfoCache (const swift::ASTContext *a)
 CachedMemberInfo *
 SwiftASTContext::GetCachedMemberInfo (void* type)
 {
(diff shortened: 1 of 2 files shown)


egranata:
@scallanan Please review

source/Symbol/SwiftASTContext.cpp
```diff
@@ -8543,6 +8627,8 @@ SwiftASTContext::GetIndexOfChildMemberWithName (void* type, const char *name,
 uint32_t
 SwiftASTContext::GetIndexOfChildWithName (void* type, const char *name, bool omit_empty_base_classes)
 {
+    VALID_OR_RETURN(0);
```
scallanan: It'd be nice if there were an invalid value we could use here, even just `UINT32_MAX`.

scallanan:
@egranata I have one suggestion, looks good otherwise.

PR #54, opened 2016-09-17
Add back "added tests for SR-2637"

tfiala:
This commit is intended to address the Objective-C header path issue that showed up on the CI configuration for the new test for SR-2637.  It brings in more of our normal LLDB test inferior logic for setting swiftc and clang flags.

This reverts commit a4630701c013d8e497266ab48b3d19d6f39758d8.


tfiala:
@swift-ci Please test OS X platform

PR #59, opened 2016-09-23
[CMake] Fix LLDB in-tree build

llvm-beanz:
This is a bunch of small fixes that enable swift-lldb to build in-tree. This change is NFC for build-script builds. With this change in-tree builds do not depend on all Clang and Swift static archives, instead the build only depends on a minimal set. Additionally the in-tree builds add many required dependencies on clang and swift targets that are not needed in the build-script builds.


tfiala:
Looks good!

tfiala:
Err, I thought I saw a build check on that but I'm not seeing it now.  If this breaks the build, I'll undo it and we get get it back up on a new PR.

tfiala:
@swift-ci Please test

(Checking to see if this can work retroactively).

PR #61, opened 2016-09-27
[CMake] For Darwin rpath $ORIGIN should be @loader_path

llvm-beanz:
This allows lldb builds on darwin with CMake to have the correct rpath.

Without this patch all binaries were getting $ORIGIN-based rpaths, which are invalid on Darwin.

### DIFF
--- a/cmake/modules/AddLLDB.cmake
+++ b/cmake/modules/AddLLDB.cmake
@@ -101,7 +101,13 @@ endmacro(add_lldb_library)
 macro(add_lldb_executable name)
   add_llvm_executable(${name} DISABLE_LLVM_LINK_LLVM_DYLIB ${ARGN})
   set_target_properties(${name} PROPERTIES FOLDER "lldb executables")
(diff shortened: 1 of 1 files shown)


cmake/modules/AddLLDB.cmake
```diff
@@ -101,7 +101,13 @@ endmacro(add_lldb_library)
 macro(add_lldb_executable name)
   add_llvm_executable(${name} DISABLE_LLVM_LINK_LLVM_DYLIB ${ARGN})
   set_target_properties(${name} PROPERTIES FOLDER "lldb executables")
-  set_target_properties(${name} PROPERTIES INSTALL_RPATH "$ORIGIN/../lib")
+  # $ORIGIN is only
```
jrose-apple: Incomplete comment?

llvm-beanz: Doh! I will push a fix momentarily.

tfiala:
LGTM.  Merging in.

PR #63, opened 2016-09-30
Fix for TypePreservingNSNumber

egranata:


jckarter:
A comment in the Swift implementation that LLDB relies on its layout would be helpful. Is there anything we could do to help isolate you from changes in the future? We don't want the implementation details of these classes to be ABI.

egranata:
@jckarter I'd like to keep this pull request minimal since it's already a fair chunk of magic and it's for the 3.0 branch. I am happy to put all the comments you want on master though!

As for insulating us from the internals, we do need to know the type of the data and where to read it from. If you can expose that to us in a way you're comfortable with, then by all means. Maybe magic symbols we can look up and an agreed-upon encoding?

jckarter:
Sure, this looks good for 3.0. Offset symbols seem like a reasonable way to expose the layout if you need it.

scallanan:
I can honestly say that I have never been more pleased to approve a patch than today.
Looks Good To Me.

PR #65, opened 2016-10-12
[CMake] Search for llvm-config in LLDB_PATH_TO_LLVM_BUILD first

vivkong:
Add NO_DEFAULT_PATH to cmake's find_program() so we don't pick up
llvm-config in cmake or system paths.

My local build as been failing as it picked up llvm-config in my build compiler since it is on the $PATH.

### DIFF
--- a/cmake/modules/LLDBStandalone.cmake
+++ b/cmake/modules/LLDBStandalone.cmake
@@ -17,7 +17,7 @@ if (CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
     set(FIND_PATHS "${LLDB_PATH_TO_LLVM_BUILD}/bin")
   endif()
   find_program(LLVM_CONFIG "llvm-config"
-    PATHS ${FIND_PATHS})
+    HINTS ${FIND_PATHS})
 
   if(LLVM_CONFIG)
     message(STATUS "Found LLVM_CONFIG as ${LLVM_CONFIG}")


tfiala:
@swift-ci Please test

tfiala:
@llvm-beanz, any concerns?

llvm-beanz:
This is wrong and will break building against installed copies of LLVM.

llvm-beanz:
Instead of specifying `NO_DEFAULT_PATHS` try using the `HINTS` option to search `LLDB_PATH_TO_LLVM_BUILD` before the system paths.

tfiala:
>  This is wrong and will break building against installed copies of LLVM.

Okay, that's what I was thinking.  Glad I had you check :-)

> Instead of specifying NO_DEFAULT_PATHS try using the HINTS option to search LLDB_PATH_TO_LLVM_BUILD before the system paths.

@vivkong, want to give that a shot?

vivkong:
Thanks for the review and I appreciate the suggestions!

I found out that I can override the search order by calling `find_program` multiple times as suggested in the documentation.  Please see my updated commit and let me know if I should make other changes.  Thanks!

llvm-beanz:
You shouldn't need to call `find_program` multiple times. You should be able to just call `find_program(LLVM_CONFIG "llvm-config" HINTS ${FIND_PATHS})`

Is there a reason act doesn't work? `find_program` searches multiple paths until it finds the program you're looking for. To get what you want we just need to ensure that the paths are searched in the right order. The current implementation searches the system path before it searches the specified paths because the `PATHS` option is the lowest priority. Using `HINTS` instead should solve the problem because it is higher priority than the system paths.

The search order is documented here: https://cmake.org/cmake/help/v3.4/command/find_program.html

vivkong:
I've updated the fix using the suggested `HINTS` options.  Thanks.

tfiala:

tfiala:

tfiala:
Incidentally, this really should go into LLVM.org LLDB.  (Which should then be reviewed at reviews.llvm.org).  Anything in LLDB that isn't specific to Swift should happen up in LLVM.org's svn trunk for LLDB.  We only insert content into the GitHub layer of the code when it is specific to Swift.

@vivkong, sorry to make more work for you, but can you head over to reviews.llvm.org, put the diff up there against LLDB svn trunk, add lldb-commits as a subscriber, and put @llvm-beanz and @labath as reviewers?  Thanks!

vivkong:
Hi @tfiala, just want to verify before I submit a patch to upstream LLDB - it seems that I'm making a change on top of your commit back in May: 
https://github.com/apple/swift-lldb/commit/9e86f44b57ce66962604ea43cb183759f5557738#diff-fe3aa12ad9ba304eb29b00bb04a5e1b3

I didn't see this fix in LLDB's trunk.  How should I proceed?  

Thanks!

tfiala:
Hi, @vivkong!

> I didn't see this fix in LLDB's trunk. 

LLDB's build configuration for Linux is a bit different in GitHub than it is in LLVM.org.  I think the biggest diff is that I believe we use .so files in places where LLVM.org LLDB Linux doesn't (or maybe vice versa), so changes upstream are sometimes breaking the .so support.  This particular colliding change is around llvm-config.

I'd say get it in upstream, and we should try to make the support here look identical on the Linux end.  If/where we find issues that fail to build and test on GitHub swift-lldb with the pattern used in LLVM.org, we need to go back to LLVM.org's LLDB and fix it there, then bring it into GitHub swift-lldb.  (i.e. the approach should be to get it right in LLVM.org LLDB svn trunk, and use it here).

So we're fine if it collides with my old change, and we should use what's up in LLVM.org as much as possible, so long as we don't break the build here.  And if we do break it, we need to figure out the cause, and get it fixed upstream.

Make sense?

tfiala:
> Or can we just make the change in LLDBStandalone.cmake (this PR)?

We're going to do this here.  (We temporarily couldn't reach GitHub or I would have kicked this off sooner).

Sorry for the back and forth on this, it is a reflection of our mismatched approaches to CMake between llvm.org and swift.org code that exists at the moment.

vivkong:
Thank you @tfiala!

PR #73, opened 2016-11-10
Fix the SwiftSILManipulator for upstream changes.

gottesmm:


gottesmm:
@swift-ci Please test

gottesmm:
I am just going to wait for the linux test to finish. If it passes, then I think it is safe to say that the OS X will as well, given how simple the change is.

rudkx:
@gottesmm Can you add a test? Is there a reason this only hit under LLDB testing and not in the compiler tests?

gottesmm:
No, I just forgot to update this given upstream changes. The assert that was hit here was the "test" that invariants were never broken. So this shows that the invariant asserts are working.

PR #78, opened 2016-11-15
PrintDiagnostics can't call VALID_OR_ERROR, since it needs to print

jimingham:
diagnostics when there is an error.

<rdar://problem/29226777>


jimingham:
@swift-ci Please test

PR #81, opened 2016-11-19
Swift 3.1 cherry picks - round 1

tfiala:
These are changes from upstream or master-next that we want to show up in Swift 3.1.

@swift-ci Please test


tfiala:
@swift-ci Please test

PR #87, opened 2016-11-30
[Swift] Fix copy/paste error

Jnosh:
These lines look like a simple copy/paste oversight.

I tried the following:
```
lldb --repl
> :settings append target.swift-module-search-paths <path to folder containing Module.swiftmodule>
> import Module
```

Without this patch the `import` will fail with `error: repl.swift:1:8: error: no such module 'Module'`, with this patch, the import succeeds.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/module_search_paths/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/module_search_paths/Makefile
@@ -1,0 +1,11 @@
+LEVEL = ../../../make
+
+SWIFT_SOURCES := main.swift
+
+include $(LEVEL)/Makefile.rules
+
+module:
+	$(SWIFTCC) -g -Onone -emit-module -module-name Module module.swift
+
+cleanup:
+	rm -rf *.swiftdoc *.swiftmodule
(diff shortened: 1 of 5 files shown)


jrose-apple:
Test case?

Jnosh:
Finally managed to get a working test case. I hope it's at least somewhat idiomatic.

I first tried to add a REPL test case to avoid the need for a dummy process but I just couldn't get the import to work there. I eventually tracked the problem down.

Turns out the REPL tests start the REPL with custom compiler arguments which causes the SwiftAST to be initialized on launch. So any import paths added through the target setting are never seen by the AST since it loads them during initialization.

Is that the intended behavior or would it make sense for the search paths to be added later?

tfiala:
@swift-ci Please test

tfiala:
@swift-ci Please test

tfiala:
> Is that the intended behavior or would it make sense for the search paths to be added later?

@jimingham or @scallanan, could either of you comment on the question from @Jnosh quoted above?

Jnosh:
> Is that the intended behavior or would it make sense for the search paths to be added later?

Just reread that question and I added one _the_ too many. Should read:

> ...would it make sense for search paths to be added later?

as in add (and potentially remove?) module/framework search paths through `target.swift-(module, framework)-search-paths` even after the AST has been initialized.

Hopefully that is a bit clearer.

tfiala:
Argh, I thought I was committing your other PR, @Jnosh.  We can do a post-commit review on this one.  I do want @jimingham or @scallanan to provide input on this, though.

PR #88, opened 2016-11-30
[Gardening] Reformat comments in SwiftASTContext

Jnosh:
* Clean up some weird line-breaks probably caused by automatic adjustment to 80 column width.
* Remove some double spaces.
* Fix a few typos and such.


tfiala:
Hi @Jnosh,

Thanks for putting energy into some gardening!

Just running through the test machinery as a precaution.

@swift-ci Please test

PR #90, opened 2016-12-01
Remove special handling for undefined LLDB_REVISION

llvm-beanz:
Returning empty string instead of nullptr is unnecessary with the refactored code here, and it actually causes some odd version printing.


llvm-beanz:

llvm-beanz:

PR #91, opened 2016-12-01
Updated the Swift expression parser to handle NULL IR modules

scallanan:
Updated the Swift expression parser to handle NULL IR modules.
    
    IRGen in rare cases can produce a NULL module without reporting an error.
    LLDB should report an error here but it should not crash.  This fixes the
    problem.
    
    Swift is still looking into what scenarios this occurs in.  Once we know that,
    we can add a testcase.
    
    <rdar://problem/28979704>


scallanan:
@swift-ci please test

scallanan:
@swift-ci please test

scallanan:
@swift-ci please test

scallanan:
Jim Ingham approved this offline.  Merging.

PR #93, opened 2016-12-01
Stop format changes

jimingham:
This change adds a thread format specifically for the stop printing, since then format you want for "thread list" is different from the thread header you want in stop printing.


jimingham:
@swift-ci please test

jimingham:
@swift-ci Please test

PR #99, opened 2016-12-02
[swift, gardening] Update code headers

Jnosh:
* Updated copyright time frame
* Use https:// links to swift.org
* Added a few missing header

I noticed that the headers in the main Swift repository are being migrated to use https:// links to swift.org. I did a quick pass to update all the Swift-project related headers for swift-lldb.
I added a few missing headers to Swift related tests but otherwise only changed files that already had a Swift project header.

Feel free to reject if this is too much noise without any clear benefit or if I should split it up into smaller groups.


tfiala:
Looks good, thanks @Jnosh!

PR #102, opened 2016-12-07
Fix typo in documentation

alblue:
fraomeworks -> frameworks


jimingham:
Thanks for catching this.

PR #105, opened 2016-12-12
Port swift-lldb to Windows

hughbe:
This requires some CMake stuff, some upstream cherry-picking, and some modification of swift-specific source code.

Note: the changes in "Fix invalid namespace qualified UTF conversion helpers on Windows" changes lldb code. However: other actual lldb code was changed from `llvm::UTF8` to `UTF8` in `swift-clang`, `swift-llvm` and `swift-lldb`. This just manifests itself now, because we never built these files on non-Windows platforms.

/cc @compnerd @jrose-apple

### DIFF
--- a/source/Core/IOHandler.cpp
+++ b/source/Core/IOHandler.cpp
@@ -28,7 +28,9 @@
 #include "lldb/Core/State.h"
 #include "lldb/Core/StreamFile.h"
 #include "lldb/Core/ValueObjectRegister.h"
+#ifndef LLDB_DISABLE_LIBEDIT
 #include "lldb/Host/Editline.h"
+#endif
 #include "lldb/Interpreter/CommandCompletions.h"
 #include "lldb/Interpreter/CommandInterpreter.h"
 #include "lldb/Symbol/Block.h"
@@ -273,12 +275,16 @@ IOHandlerEditline::IOHandlerEditline(
     IOHandlerDelegate &delegate)
     : IOHandler(debugger, type, input_sp, output_sp, error_sp, flags),
 #ifndef LLDB_DISABLE_LIBEDIT
-      m_editline_ap(), m_delegate(delegate), m_prompt(),
-      m_continuation_prompt(), m_current_lines_ptr(nullptr),
-      m_base_line_number(line_number_start), m_curr_line_idx(UINT32_MAX),
-      m_multi_line(multi_line), m_color_prompts(color_prompts),
-      m_interrupt_exits(true), m_editing(false) {
+      m_editline_ap(),
+#endif
+      m_delegate(delegate), m_prompt(), m_continuation_prompt(),
+      m_current_lines_ptr(nullptr), m_base_line_number(line_number_start),
+      m_curr_line_idx(UINT32_MAX), m_multi_line(multi_line),
+      m_color_prompts(color_prompts), m_interrupt_exits(true),
+      m_editing(false) {
   SetPrompt(prompt);
+
+#ifndef LLDB_DISABLE_LIBEDIT
   bool use_editline = false;
 
   use_editline = m_input_sp->GetFile().GetIsRealTerminal();
@@ -306,7 +312,11 @@ IOHandlerEditline::IOHandlerEditline(
   SetContinuationPrompt(continuation_prompt);
 }
 
-IOHandlerEditline::~IOHandlerEditline() { m_editline_ap.reset(); }
+IOHandlerEditline::~IOHandlerEditline() {
+#ifndef LLDB_DISABLE_LIBEDIT
+  m_editline_ap.reset();
+#endif
+}
 
 void IOHandlerEditline::Activate() {
   IOHandler::Activate();
@@ -319,9 +329,11 @@ void IOHandlerEditline::Deactivate() {
 }
 
 bool IOHandlerEditline::GetLine(std::string &line, bool &interrupted) {
+#ifndef LLDB_DISABLE_LIBEDIT
   if (m_editline_ap) {
     return m_editline_ap->GetLine(line, interrupted);
   } else {
+#endif
     line.clear();
 
     FILE *in = GetInputFILE();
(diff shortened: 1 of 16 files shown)


hughbe:
I'm fairly sure I've found a bug in `ExpressionSourceCode.cpp` on non-Windows platforms.
Take a look at this code: https://github.com/hughbe/swift-lldb/blob/10d41feb230bf7c1c0137aef071aaddff0d9df8a/source/Expression/ExpressionSourceCode.cpp#L232-L251

Notice that we append `.cpp` or `.swift` to the end of the file template. However, this goes against the documentation for `mkstemp`

> The last six characters of template must be "XXXXXX" and these are
    replaced with a string that makes the filename unique.

However, surely if we append `.cpp` or `.swift` to the file path passed to `mkstemp`, then the last six characters won't be `XXXXXX`. Not sure what's going on here.

jrose-apple:
We need LLDB reviewers in here more than us. Calling @scallanan and @clayborg.

source/Expression/ExpressionSourceCode.cpp
```diff
@@ -236,8 +239,16 @@ bool ExpressionSourceCode::SaveExpressionTextToTempFile(
     break;
   }
 
+#if defined(_WIN32)
```
clayborg: If making temp files needs abstraction for other systems we should look and see if there is a LLVM equivalent. We don't want every place in the LLVM/Clang/LLDB source base to do any "#if defined(_WIN32)"

source/Core/IOHandler.cpp
```diff
@@ -28,7 +28,9 @@
 #include "lldb/Core/State.h"
 #include "lldb/Core/StreamFile.h"
 #include "lldb/Core/ValueObjectRegister.h"
+#ifndef LLDB_DISABLE_LIBEDIT
```
clayborg: Does this file differ from the LLDB llvm.org SVN top of tree and the changes here are just syncing with top of tree SVN?

source/Expression/ExpressionSourceCode.cpp
```diff
@@ -252,6 +263,10 @@ bool ExpressionSourceCode::SaveExpressionTextToTempFile(
     }
     if (!success)
       FileSystem::Unlink(FileSpec(expr_source_path.c_str(), true));
+
+#if defined(_WIN32)
```
clayborg: this line and the two below should go away if we abstract temp filenames into llvm or the LLDB host layer.

source/Host/common/Socket.cpp
```diff
@@ -363,7 +363,7 @@ Error Socket::Close() {
   Log *log(lldb_private::GetLogIfAnyCategoriesSet(LIBLLDB_LOG_CONNECTION));
   if (log)
     log->Printf("%p Socket::Close (fd = %i)", static_cast<void *>(this),
-                m_socket);
+                static_cast<int>(m_socket));
```
clayborg: Isn't m_socket a SOCKET on windows?:

SOCKET WSAAPI socket(
  _In_ int af,
  _In_ int type,
  _In_ int protocol
);

How does casting this to an integer help here?

source/Plugins/ExpressionParser/Clang/ClangExpressionDeclMap.h
```diff
@@ -362,7 +362,7 @@ class ClangExpressionDeclMap : public ClangASTSource {
   //----------------------------------------------------------------------
   class ParserVars {
   public:
-    ParserVars(ClangExpressionDeclMap &decl_map) : m_decl_map(decl_map) {}
+    ParserVars() {}
```
clayborg: I'll let Sean Callanan comment on all expression parser changes.

source/Host/common/Socket.cpp
hughbe: Socket is defined as:
```
typedef UINT_PTR        SOCKET;
```
The warning we therefore get is trying to convert `unsigned long long` to `int`

source/Core/IOHandler.cpp
hughbe: I basically fixed the build errors that I presume were caused by a bad merge - many syntax errors occured as a result - e.g. missing `{`, unhandled identifiers etc.

Take a look at this commit here: https://github.com/hughbe/swift-lldb/commit/8d544a47ba95533c369949fa3bf0ad01dadea0a3
It demonstrates what the diff looks like when I copy-and-paste the remaining bits of `IOHandler.cpp` from upstream. Basically, its changes in API that I didn't want to modify right now because I don't really wan't to break all the things!

source/Expression/ExpressionSourceCode.cpp
hughbe: Yeah, this is ugly. Let me know about this function: http://llvm.org/docs/doxygen/html/namespacellvm_1_1sys_1_1fs.html#ae3cdedf952d5eeb089d175c61d9f6fee

I think it does what we want

tfiala:
No harm in testing.

@swift-ci Please test

tfiala:
@swift-ci Please test

scallanan:
It looks like the expression parser changes are just cleanup that removes unused state.  I love it (assuming the tests succeed)!

hughbe:
I had to rebase to fix a merge conflict, but I recall the tests passing before. Probably a good idea to retest :D

hughbe:
Friendly ping!

hughbe:
@swift-ci please test

hughbe:
I get the feeling this is targeting the wrong branch... Which branch should I submit this PR to?

kubamracek:
I believe you should just target the master branch.

hughbe:
Thanks! I know that trying to change the branch of this PR will end up notifying and sending emails to several hundred people as GitHub attempts to push their commits from the old branch to the new branch - see https://github.com/apple/swift-clang/pull/44

So unfortunately, I'll have to close this PR and create a new one: #125

Sorry for losing the discussion history

PR #111, opened 2016-12-20
Cmake framework install

llvm-beanz:
Fixing LLDB's CMake to support installing the LLDB framework.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,12 @@ if (NOT LLDB_DISABLE_PYTHON)
     # Ensure we do the python post-build step when building lldb.
     add_dependencies(lldb finish_swig)
 
+    if(LLDB_BUILD_FRAMEWORK)
+      # The target to install libLLDB needs to depend on finish swig so that the
+      # framework build properly copies over the Python files.
+      add_dependencies(install-liblldb finish_swig)
+    endif()
+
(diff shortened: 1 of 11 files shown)


llvm-beanz:
@swift-ci please test

dcci:
I spoke with Chris offline and he's not working on this anymore. I think Tim will pick his work & when that happens he can resurrect this PR.

keith:
@dcci Do you have a sense of how much more there is to do here? I might be interested in picking this up, although I'm not super familiar with CMake. Being able to run the tests that rely on LLDB.framework without having to do a full xcodebuild would be great. I've been developing with ninja locally because the rebuilds are so much faster, but then when I have to run the tests I have to re-build with xcodebuild.

llvm-beanz:
@keith It mostly works as is today. There are some test failures when you build the framework with CMake and nobody has had time to track them down. I'm fairly certain the are caused by CMake's intermediate directory being structured like an install directory instead of how Xcode puts built products in a flat directory.

If you want to go down that path I also recommend ditching build-script and use CMake to configure Swift as part of LLVM, Clang, LLDB. That dramatically speeds up build times particularly for iteration cycles (i.e. running `check-swift` or `check-lldb` is WAY faster because you build way less).

dcci:
@Keith I think @llvm-beanz already commented on how much work will be.
I'll be happy if you consider picking this up & I'll review your work (definitely looking forward to having an alternative to `xcodebuild` to build)

PR #112, opened 2016-12-21
Updated the XCTest support to handle the correct executable names.

scallanan:
Previously LLDB was looking for the wrong test runner binary, leading
to problems loading the Swift module for the test bundle.  I have updated
the logic, and the corresponding test.

Also moved the logic into PlatformDarwin where it belongs, rather than
having it spread between PlatformDarwin and SwiftASTContext.

<rdar://problem/28412550>


scallanan:
@swift-ci please test

PR #113, opened 2016-12-23
Makes VALID_OR_* failed asserts print error before dying.

gmilos:
Some associated cleanups, especially separating diagnostics
gathering (CopyDiagnostics) and mutating state (PrintDiagnostics).

rdar://problem/29769303

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -99,34 +99,31 @@
 #include "Plugins/SymbolFile/DWARF/DWARFASTParserSwift.h"
 
 #ifdef LLDB_CONFIGURATION_DEBUG
-#define VALID_OR_RETURN(value)                                                 \
-  do {                                                                         \
-    lldbassert(!HasFatalErrors());                                             \
-    if (HasFatalErrors()) {                                                    \
-      return (value);                                                          \
-    }                                                                          \
-  } while (0)
-#define VALID_OR_RETURN_VOID()                                                 \
-  do {                                                                         \
-    lldbassert(!HasFatalErrors());                                             \
-    if (HasFatalErrors()) {                                                    \
-      return;                                                                  \
-    }                                                                          \
-  } while (0)
-#else
-#define VALID_OR_RETURN(value)                                                 \
+#define _VALID_OR_RETURN(_return)                                              \
   do {                                                                         \
     if (HasFatalErrors()) {                                                    \
-      return (value);                                                          \
+      DiagnosticManager diagnostic_manager;                                    \
+      CopyDiagnostics(diagnostic_manager);                                     \
+      llvm::errs() << diagnostic_manager.GetString();                          \
+      lldbassert(false);                                                       \
+      _return;                                                                 \
     }                                                                          \
   } while (0)
-#define VALID_OR_RETURN_VOID()                                                 \
+
+#define VALID_OR_RETURN(value) _VALID_OR_RETURN(return(value))
+#define VALID_OR_RETURN_VOID() _VALID_OR_RETURN(return)
+#else /* LLDB_CONFIGURATION_DEBUG */
+#define _VALID_OR_RETURN(_return)                                              \
   do {                                                                         \
     if (HasFatalErrors()) {                                                    \
-      return;                                                                  \
+      _return;                                                                 \
     }                                                                          \
   } while (0)
-#endif
+
+#define VALID_OR_RETURN(value) _VALID_OR_RETURN(return(value))
+#define VALID_OR_RETURN_VOID() _VALID_OR_RETURN(return)
+#endif /* LLDB_CONFIGURATION_DEBUG */
+
 
 using namespace lldb;
 using namespace lldb_private;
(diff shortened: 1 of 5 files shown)


gmilos:
Could someone kick off CI tests on this please?

gmilos:
@jimingham ^

jimingham:
@swift-ci please test

source/Symbol/SwiftASTContext.cpp
```diff
@@ -99,34 +99,32 @@
...
+#define _VALID_OR_RETURN(_return)                                              \
   do {                                                                         \
-    lldbassert(!HasFatalErrors());                                             \
     if (HasFatalErrors()) {                                                    \
-      return (value);                                                          \
+      DiagnosticManager diagnostic_manager;                                    \
+      CopyDiagnostics(diagnostic_manager);                                    \
+      llvm::errs() << diagnostic_manager.GetString();                          \
+      lldbassert(!HasFatalErrors());                                           \
+      lldbassert(false);                                                       \
```
jimingham: Why are there two lldbasserts here?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -99,34 +99,32 @@
...
 #ifdef LLDB_CONFIGURATION_DEBUG
-#define VALID_OR_RETURN(value)                                                 \
+#define _VALID_OR_RETURN(_return)                                              \
   do {                                                                         \
-    lldbassert(!HasFatalErrors());                                             \
     if (HasFatalErrors()) {                                                    \
-      return (value);                                                          \
+      DiagnosticManager diagnostic_manager;                                    \
+      CopyDiagnostics(diagnostic_manager);                                    \
+      llvm::errs() << diagnostic_manager.GetString();                          \
```
jimingham: I'm worried that dumping this to an error channel will just cause this to get lost.  For instance, when lldb is used in an IDE stderr tends to be the system console.  It might be better to make it go to one of lldb's log channels (as well if not instead.)  Also, VALID_OR_RETURN can get called a lot of times as we back out of trying to do something with an AST context that had errors.  So you are likely to really spam whatever output channel you are sending this to.  It would be good to try this out and make sure that this doesn't get really obnoxious.  If it does we might want to have some way to emit the diagnostics once per SwiftASTContext.

gmilos:
@jimingham @scallanan This slipped through the cracks earlier, I'd like to close a loop on this (whichever way we decide to take it).

I'm responding to Jim's comments inline.

Some background to save you having to trawl through radar history + make this public: the primary motivation for this change is getting the debug builds of lldb to print useful information before dying. In debug build `lldbassert()` turns into `assert()`. On non-debug build, `lldbassert()` turns into a print to `llvm::errs()`. Also, my usecase was to develop lldb from command-line rather than from Xcode.

source/Symbol/SwiftASTContext.cpp
gmilos: It was just to get the right `assert` message (stringified `HasFatalError()`) and at the same time being guaranteed that we are going to die (hence the `lldbassert(false)`), but of course on non-debug builds we don't die, so `lldbassert` would end up printing 2x.

Retaining just the `lldbassert(false)`.

source/Symbol/SwiftASTContext.cpp
gmilos: Ok, so there are two separate issues:

1. Spamming `errs()`: I'm not too sure this is such a big deal as every failed `lldbassert()` already generates lots of output: https://github.com/apple/swift-lldb/blob/765c9aab8edc3b2e30bb0505aee8464c62bbfe30/source/Utility/LLDBAssert.cpp#L26-L31.
2. stderr getting lost: I'm looking at this patch as as attempt to produce more verbose `lldbassert()` for `SwiftASTContext.cpp`, hence logging to the same place where `lldbassert()` logs to. Logging _just_ to a log channel will split some of the information from the rest, and doesn't seem right. Logging to both `errs()` and a log channel is OK, but produces even more spam (and works against 1.)

Ultimately I'm happy with whatever you think is the best here. I just want a way of seeing this information without maintaining a fork. Please let me know which way you want to go, and I'll implement it so.

alblue:
@swift-ci please test

alblue:
@jimingham can you close the review so that we can kick off a build please?

alblue:
@jimingham can you close the review so that we can kick off a build please?

jimingham:
Sorry for the delay.  This is fine.  This change touches the DiagnosticManager.{h, cpp} files that are shared with the llvm.org lldb sources, so to prevent future merging conflicts, can you apply those changes to llvm.org & then cherry-pick them to swift?

alblue:
@gmilos do you have the ability to upstream the changes to LLVM?

gmilos:
@jimingham I split up the change into two commits, I also submitted the LLVM/LLDB upstream one here: https://reviews.llvm.org/D32078. I'm not totally sure this fits within the accepted process (if only because the patch couldn't be attached to lldb repository, and it's therefore missing context).

Please let me know if I need to do anything more on the upstream change. Once accepted, I'll re-do this PR whichever way you think is best (is there an accepted cherry-picking process?).

jimingham:
Gregor,

Not sure what you mean by "the patch couldn't be attached to the lldb repository"?  DiagnositicManager exists in the llvm.org version of lldb.  But anyway, that bit looks fine so you should be able to check it into llvm.org, or if you don't have permission to do that I will.

I've never done a pull request that is actually a cherry-pick, so I don't know if there is a formal way to do that.  I don't think we need to be that formal for a patch that just adds a function you intend to use elsewhere.  The only thing the pull requests add over the llvm.org review is test gating but it's hard to see how adding a function could affect the testsuite, so.

Then the swift specific parts can go as a regular pull request.


Jim

dcci:
I see this was committed to LLVM and will be picked by the next merge. Closing this one.

PR #115, opened 2017-01-06
[SR-2757][Swift] Use new capture list VarDecl initializer

modocache:
Addresses SR-2757.

https://github.com/apple/swift/pull/6490 modified the VarDecl to add a boolean parameter to specify whether the variable was an element in a closure capture list, in order to improve diagnostics in the Swift compiler. Use this new parameter.


CodaFi:
⛵️

PR #116, opened 2017-01-07
Updated SwiftASTContext to find a single module with Swift flags.

scallanan:
This makes LLDB get flags from the following sources (in order of preference):

  - the main executable;
  - a unit test bundle; or
  - the first module that has a Swift AST.

We used to take the union of all the Swift flags if the first two cases didn't apply.
This was a bad approach and caused a lot of user frustration (mostly of the form
"po doesn't work.")

<rdar://problem/29906378>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1254,6 +1254,48 @@ static std::string &GetDefaultResourceDir() {
   return s_resource_dir;
 }
 
+static bool HasASTData(Module &module) {
+  if (SymbolVendor *sym_vendor = module.GetSymbolVendor()) {
+    // We could do more validation here, but the validation requires that
+    // this mutate a CompilerInstance, which is not appropriate for this
+    // function.  Instead, we let the caller do the rest of the validation.
+    return !(sym_vendor->GetASTData(eLanguageTypeSwift).empty());
+  }
+
+  return true;
+}
+
+static lldb::ModuleSP GetSwiftModule(Target &target) {
+  ModuleSP candidate_sp(target.GetExecutableModule());
+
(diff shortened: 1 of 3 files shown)


jrose-apple:
The approach sounds good to me, except that I think the unit test bundle ought to win over the main executable. I get that that's trickier, though.

scallanan:
@swift-ci please test

gottesmm:
@scallanan A request: Can you rebase on your branch rather than merging in? It makes a nicer git history when you merge.

scallanan:
@gottesmm If that's standard procedure, I'm happy to follow suit.

gottesmm:
@scallanan Just to be clear I am only talking on this branch. You should merge it in normally (not squash). That is the procedure we are using in general terms.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1670,48 +1720,38 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
...
-      bool read_options_from_ast = false;
-      ModuleSP exe_module_sp(target->GetExecutableModule());
-
-      // If we're debugging a testsuite, then treat the main test bundle as the
-      // executable.
-      if (exe_module_sp &&
-          PlatformDarwin::IsUnitTestExecutable(*exe_module_sp)) {
-        ModuleSP unit_test_module =
-            PlatformDarwin::GetUnitTestModule(target->GetImages());
+      ModuleSP exe_module_sp(GetModuleForCanonicalCompilerFlags(*target));
```
jimingham: Trivial, but exe_module_sp is no longer the correct name for this, since it isn't the main executable.  Maybe flags_module_sp?

dcci:
Close as nobody is working actively on this. We can reopen if we need to.

PR #117, opened 2017-01-08
[Swift] Completely replace Module with ModuleDecl

modocache:
The typedef `swift::Module` was a temporary solution in the Swift codebase that allowed `swift::Module` to be renamed to `swift::ModuleDecl` without requiring every single callsite to be modified.

Completely stop using `swift::Module`, in order to then remove the typedef in the Swift codebase.


CodaFi:
⛵️

modocache:
Thanks, @CodaFi! 🙇

PR #118, opened 2017-01-11
Merge CMake patches from upstream

llvm-beanz:
This merges a bunch of patches from LLVM.org into Swift-LLDB.


llvm-beanz:
@swift-ci please test

PR #119, opened 2017-01-23
Mark ASan and TSan Swift tests as "skipUnlessSwift...Sanitizer".

kubamracek:
Mark ASan and TSan Swift tests as "skipUnlessSwift...Sanitizer".  This will detect whether the Swift compiler can successfully build with -fsanitize=address/-fsanitize=thread and skip the tests if not.  This should fix the tests when Swift is checked out without the compiler-rt repo.


kubabrecka:
@swift-ci please test

PR #122, opened 2017-01-26
Eliminated repl_swift's dependency on the Darwin overlay.

scallanan:
repl_swift only uses a few functions in Darwin that can very easily be brought in as symbols.
It does not use the Darwin overlay at all, and the overlay constitutes an extra dependency.
This patch eliminates that dependency, creating a module that defines the necessary functions.


scallanan:
@swift-ci please test

tools/repl/swift/REPLTime/REPLTime.h
```diff
@@ -0,0 +1,38 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+// -----------------------------------------------------------------------------
+
+#ifndef REPL_TIME_H
+#define REPL_TIME_H
+
+typedef unsigned long long uint64_t;
```
jrose-apple: Ah, redefining these types is likely to cause problems; if you really need to use them, you should name them something different. In retrospect that means this isn't going to work because you can't redefine the real `mach_absolute_time` in the header; you'd have to make it a mixed-source target (i.e. compile in a .c file to forward the calls for you).

jrose-apple: It *might* work just because you literally don't import anything else, but it's pretty brittle.

scallanan: It is possible to import other modules interactively via the REPL.  Is that going to cause collisions?

scallanan: The repl_swift built purely by Xcode doesn't have debug information so we will not load it.
The answer here may simply be to make sure the binary is always stripped of its debug info.

scallanan:
Updated to reflect Jordan's comments.

scallanan:
@swift-ci please test

PR #123, opened 2017-01-27
Support for programmatically disabling the REPL build

scallanan:
Some configurations may require building the LLDB `swift_repl` binary to be disabled.
Because Xcode doesn't support conditional targets, we instead have the binary built and installed by scripts.


scallanan:
@swift-ci please test

PR #124, opened 2017-01-28
Made repl_swift build with the C compiler under Xcode.

scallanan:
In some environments the Swift compiler is unusable at the time LLDB is
being built, and the Swift compiler LLDB builds is also not usable.  In this
case making repl_swift a simple C binary is preferable.

It still needs to load the core Swift library so that LLDB can find the
necessary symbols, so I've ensured that the Swift library is linked in and
actually loaded by the dynamic loader.


scallanan:
@swift-ci please test

scallanan:
@swift-ci please test

scallanan:
@swift-ci please test

scallanan:
@swift-ci please test

scallanan:
@swift-ci please test

scallanan:
Checks have passed.  I will leave the scripts and sources used to build repl_swift with the Swift compiler in place, for cmake and in case some Xcode build configurations turns out to need  to do it the old way.

PR #128, opened 2017-01-30
Print Double.self instead of a double instance

scallanan:
This fixes test/testcases/lang/swift/tmd_symbol by forcing the metadata for Double to be generated in a more reliable way.

PR #129, opened 2017-01-31
XFail symbol-checking test that uses string interpolation

DougGregor:
This test is fairly brittle, and breaks with some simple changes in the way string interpolation is compiled, and the commentary in https://github.com/apple/swift-lldb/pull/128 indicates that it's not important. Xfail it for now to unblock https://github.com/apple/swift/pull/7146.

(cherry picked from commit dbb730e8ca79a468e87c49c302c31d1d0fe96f22)


DougGregor:
@swift-ci please test

DougGregor:
Yay Linux build succeeded

PR #130, opened 2017-01-31
Verify that LLDB can debug source files compiled with older -swift-versions.

scallanan:
LLDB's expression parser will operate in the mode of the latest Swift, but it needs to interoperate with code with different `-swift-version` settings.  I have added a test that checks that this continues to work as the Swift language and compiler evolve.


scallanan:
@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/swift_version/TestSwiftVersion.py
```diff
@@ -0,0 +1,93 @@
...
+import unittest2
+
+
+def execute_command(command):
+    # print '%% %s' % (command)
+    (exit_status, output) = commands.getstatusoutput(command)
+    # if output:
+    #     print output
+    # print 'status = %u' % (exit_status)
+    return exit_status
```
jimingham: TestExec.py and TestObjCiVarIMP.py have pretty much this same function defined, except that they do:

    (exit_status, output) = seven.get_command_status_output(command)

seven is a wrapper to handle py2 and py3, so you probably want to use it.  And three copies of this is too many, can you move this up into lldbtest or somewhere central and share it?

packages/Python/lldbsuite/test/lang/swift/swift_version/TestSwiftVersion.py
```diff
@@ -0,0 +1,93 @@
...
+        # Create the target
+        target = self.dbg.CreateTarget(exe_path)
+        self.assertTrue(target, VALID_TARGET)
+
+        for t in tests:
+          source_name = t['file']
+          source_spec = lldb.SBFileSpec(source_name)
+
+          breakpoint = target.BreakpointCreateBySourceRegex(t['source_regex'], source_spec)
+          self.assertTrue(breakpoint.GetNumLocations() > 0, VALID_BREAKPOINT)
```
jimingham: Since you're running a couple of tests in a for loop, if one of the instances fails three will be no breadcrumbs telling us which one.  Can you replace VALID_BREAKPOINT with some string that indicates while breakpoint you were trying to set?

packages/Python/lldbsuite/test/lang/swift/swift_version/mod3.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+// -----------------------------------------------------------------------------
+
+struct S3 {
+  var i = 3
+}
+
+public func f3() -> Int {
+  return 3 // break 3
+}
```
jimingham: I think you are missing a new-line here.

packages/Python/lldbsuite/test/lang/swift/swift_version/TestSwiftVersion.py
```diff
@@ -0,0 +1,93 @@
...
+
+        process = target.LaunchSimple(None, None, os.getcwd())
+        self.assertTrue(process, PROCESS_IS_VALID)
+
+        for t in tests:
+          self.expect("expr " + t['expr'], DATA_TYPES_DISPLAYED_CORRECTLY, substrs=[t['substr']])
```
jimingham: I tend to use frame.EvaluteExpression so I can print the error when it fails.  That really helps for transient failures on bots.

packages/Python/lldbsuite/test/lang/swift/swift_version/TestSwiftVersion.py
scallanan: Will do.

packages/Python/lldbsuite/test/lang/swift/swift_version/mod3.swift
scallanan: Will fix.

packages/Python/lldbsuite/test/lang/swift/swift_version/TestSwiftVersion.py
scallanan: Will do.

packages/Python/lldbsuite/test/lang/swift/swift_version/TestSwiftVersion.py
scallanan: Will do.

scallanan:
@swift-ci please test

scallanan:
@swift-ci please test

scallanan:
I would prefer to do that as a part of a separate PR.  Would that be all right with you?

jimingham:
Sure, provided you do it, do it at your convenience.

PR #132, opened 2017-02-03
Handle swift "indirect enums" correctly.

jimingham:
A swift "indirect enum" is equivalent to an enum where all the payload
elements are indirect.  But swift::EnumElementDecl::isIndirect returns
false in this case and you have to check that the parent is indirect
as well to get the right answer.

<rdar://problem/29953436>


jimingham:
@swift-ci please test

jckarter:
Thanks @jimingham!

PR #136, opened 2017-02-13
Weak link libswiftCore.dylib from repl_swift

scallanan:
This severs a build dependency on the swift standard library, simplifying the Mac OS X build.


scallanan:
@swift-ci please test

jrose-apple:
:-( This will impact REPL launch times, which are already pretty bad. Are there any alternatives here?

scallanan:
Would RTLD_LAZY be better?

jrose-apple:
Sean talked to me in person and relayed that dlopen would not be significantly slower than static linking here.

PR #138, opened 2017-02-14
Disable swift returns test case

aschwaighofer:
Needed after enabling the swift calling convention (https://github.com/apple/swift/pull/7237).


aschwaighofer:
No this is not right ... wrong branch probably. closing.

PR #151, opened 2017-02-22
Prepare for Demangle::Node::getText() will return a StringRef and not a std::string.

eeckstein:
…


eeckstein:
@swift-ci Please smoke test

eeckstein:
@swift-ci Please test

PR #156, opened 2017-03-02
Normalize the LLVM cmake path before appending it to the module path

hughbe:
`LLVM_CMAKE_PATH` has backlashes in it. CMake then tries to append this to `CMAKE_MODULE_PATH` but gets confused and errors out as it thinks we're providing an escape sequence (that's unknown, causing generation to fail)

E.g "C:\Users\hugh\..."


hughbe:
@swift-ci please test

hughbe:
@jimingham PTAL, should this go to upstream LLDB?

hughbe:
and @llvm-beanz 'cause you're the CMake guy, at least looking at commits

llvm-beanz:
This change looks fine, but should definitely go to LLVM.org. The CMake logic being changed here is identical upstream.

jimingham:
By default, all commits that are in non-swift code should go into llvm.org and get merged back to the swift lldb.  That way they everybody gets the benefit of the changes, and they won't cause later merge conflicts if they get fixed differently upstream.

hughbe:
@swift-ci please test

hughbe:
I've updated the PR to cherry-pick my commit from upstream lldb

llvm-beanz:
LGTM!

PR #157, opened 2017-03-02
Fix IOHandler.cpp for systems without libedit

hughbe:
Fixes multiple build errors on Windows. This never occured on most systems as they have libedit.

The fix is to copy-paste from upstream lldb


hughbe:
@swift-ci please test

hughbe:
@swift-ci please test Linux Platform

PR #159, opened 2017-03-02
Fix control path warnings in swift specific LLDB code

hughbe:


hughbe:
@swift-ci please test

PR #160, opened 2017-03-02
Include header instead of cpp in SwiftLanguage.cpp

hughbe:
This caused linker errors on Windows


hughbe:
@swift-ci please test

PR #161, opened 2017-03-02
Fix MSVC signed/unsigned conversion and size_t conversion warnings

hughbe:


hughbe:
@swift-ci please test

jimingham (changes requested):
The first of these two is in generic lldb code, and it is better to make this change in llvm.org and cherry-pick it to here or it might later cause merge conflicts when we sync up again.

hughbe:
@swift-ci please test

hughbe:
@swift-ci please test

hughbe:
Thanks for taking a look. I've opened https://reviews.llvm.org/D30926

The upstream LLVM contribution process it pretty tricky so I might not have done it correctly - please let me know

PR #163, opened 2017-03-02
Use __FUNCSIG__ with MSVC instead of __PRETTY_FUNCTION

hughbe:


hughbe:
@swift-ci please test

PR #170, opened 2017-03-16
Fix a problem with new symbol mangling in the repl.

eeckstein:
It's not sufficient to grep for the "lldb_result" variable in a mangled name,
because in the new mangling scheme parts of the name may be substituted.
Instead the name has to be de-mangled.


eeckstein:
@swift-ci Please test

PR #178, opened 2017-04-12
Add @discardableResult annotation to repl_main()

ianpartridge:
This fixes a compilation warning:

```
lldb/tools/repl/swift/main.swift:57:1: warning: result of call to 'repl_main()' is unused
repl_main()
^        ~~
```

I hope `swift-4.0-branch` is the right home for this.

### DIFF
--- a/tools/repl/swift/main.swift
+++ b/tools/repl/swift/main.swift
@@ -17,6 +17,7 @@
 import Darwin
 #endif
 
+@discardableResult
 func repl_main() -> Int
 {
     return 0


ianpartridge:
Hi @jimingham you seem to be a reviewer around here 😄 - this is a tiny thing, wondering if you could take a look?

jimingham:
Thanks for the detailing!  

This should go on the "stable" branch of lldb.  That gets periodically merged onto the swift 4.0 branch, but not the other way, so it's best to start with stable.  You can also cherry-pick it to 4.0 from stable if you don't want to wait till the next time we merge.

jimingham:
But don't worry about that for this commit, I'll cherry-pick it back to stable.

ianpartridge:
Thanks!

PR #190, opened 2017-04-22
Re-enable some tests

slavapestov:


slavapestov:
@jimingham Any objections to re-enabling these?

PR #196, opened 2017-05-04
    xfail the test TestSwiftPartiallyGenericFuncContinuation.py.

jimingham:
This isn't actually a failure in this test, but rather the debug info has
    changed so that each of the lines in the example have two regions, one of
    which has locals, and one of which has no locals.  The inline tests aren't
    flexible enough for me to work around this in the test.

    <rdar://problem/31975108>


jimingham:
Explanation: A debug info change is causing an lldb test to fail.  xfail the test while waiting for the swift change that will allow it to pass again.
Scope: Just a test xfail to get the testsuite green again.
Radar: <rdar://problem/31994513>
Risk: The bug that is causing the failure is not related to the thing being tested.  It might be possible to rework the test to still test the issue at hand w/o falling afoul of the bug, but that would have involved totally rewriting the test.  But that does mean there is a window where somebody could break the functionality the test actually tests and we won't notice till the blocking bug is fixed.
Testing: With the xfail, the failure is turned to an expected failure.

PR #197, opened 2017-05-05
Adjust to return type change of getBaseName in main swift repo

ahoppen:
In the main swift repo, the return type of getBaseName has
changed to `DeclBaseName` so we need to call getBaseIdentifier
now to get a declaration's `Identifier`

This goes hand-in-hand with apple/swift#9321


jrose-apple:
apple/swift-lldb#197
@swift-ci Please smoke test

PR #198, opened 2017-05-08
xfail TestSwiftBridgedArray.py

jimingham:
Something changed in the way Swift Numeric types are stored in
bridged NS Arrays.

<rdar://problem/32024572> LLDB test_swift_bridged_array failing

This just xfails the test to get the bots green.


jimingham:
Explanation: A test started failing in lldb (no corresponding lldb change.)  This fix resets the bot to green while we figure out what actually went wrong.
Scope: This is just xfailing a test.  We have not yet found the root cause of the issue.
Radar (and possibly SR Issue): <rdar://problem/32024572>
Risk: This is just xfailing a test.
Testing: The test now returns an expected failure not a real failure.

PR #199, opened 2017-05-08
Be a little more permissive in DynamicLoaderMacOS::CanLoadImage

jimingham:
If we can't find the "is dyld locked" symbol, assume it is safe
to load the image unless we only have 1 image loaded - in which case
we are in _dyld_start and it is definitely NOT safe.

Also add a little better errors to that function, and better logging
in SBProcess.cpp.

<rdar://problem/30174817>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@302327 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/source/API/SBProcess.cpp
+++ b/source/API/SBProcess.cpp
@@ -1135,22 +1135,34 @@ uint32_t SBProcess::LoadImage(lldb::SBFileSpec &sb_remote_image_spec,
 uint32_t SBProcess::LoadImage(const lldb::SBFileSpec &sb_local_image_spec,
                               const lldb::SBFileSpec &sb_remote_image_spec,
                               lldb::SBError &sb_error) {
+  Log *log(lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_API));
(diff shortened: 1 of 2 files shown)


jimingham:
@swift-ci please test

jimingham:
Explanation: When lldb goes to load a dylib into a process it is debugging, it first checks that dyld isn't in a bad state.  That involves checking the value of a dyld static.  In cases where the shared cache upload has failed, that variable isn't available.  In that case, instead of refusing to load the library, we fall back on the more permissive test to ensure we have loaded some libraries, and otherwise let the load go forward.
Scope: This was breaking Xcode features that use lldb to load dylibs into the target.
Radar (and possibly SR Issue): <rdar://problem/30174817>
Risk: We might not protect the user against loading a library when it was dangerous to do so, but only in the case where their symbol upload from the device failed for some reason.
Testing: You have to have a bad exploded shared cache to trigger this failure.  You can do that by deleting libdyld.dylib from the exploded shared cache, then observe that SBProcess::LoadImage still succeeds.

PR #200, opened 2017-05-09
Fix the summary string for change to representation of Foundation.Ind…

jimingham:
…exPath.

<rdar://problem/32040811>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/foundation_value_types/indexpath/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/foundation_value_types/indexpath/main.swift
@@ -13,7 +13,13 @@ import Foundation
 
 func main() {
   var path = IndexPath(indexes: [1,2,3,4,5])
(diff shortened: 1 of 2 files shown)


jimingham:
Explanation: The internal representation of Foundation.IndexPath changed, and lldb's summary formatter needs to accommodate the change.
Scope: Only affects gathering the summary string for variables of type Foundation.IndexPath.
Radar (and possibly SR Issue): <rdar://problem/32040811>
Risk: Very low, this is just gathering a string to present to the user for values of this type.  The previous string was wrong after the Foundation changes.
Testing: The test case TestSwiftFoundationTypeIndexPath.py demonstrated the failure and with this fix it passes.  I added a few more examples to the test case as well.

jimingham:
@swift-ci please test macOS

jimingham:
@swift-ci please test platform MacOS

jimingham:
@swift-ci please test

PR #201, opened 2017-05-11
Swift 4.0 branch with exclusivity suppression

lhames:
Explanation:

These two commits teach LLDB to temporarily suppress enforcement of exclusivity during evaluation of swift expressions. This is necessary to allow debugger expressions to access variables that are exclusively owned in the original program in the context where the expression is to be evaluated.

Scope: Affects evaluation of swift expressions in LLDB. Will disable exclusivity enforcement by the swift runtime while each expression is executed, then re-enable it afterwards. This will suppress exclusivity errors in expressions, but should not affect expression evaluation otherwise.  

Radar (and possibly SR Issue): <rdar://problem/31715465>

Risk: Affects evaluation of swift expressions under LLDB only. Low risk.

Testing: Some testing by hand, plus test cases for:
(1) a single expression accessing an exclusively owned variable, and
(2) two nested expressions accessing an exclusively owned variable

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -386,6 +386,9 @@ public:
   lldb::SyntheticChildrenSP
   GetBridgedSyntheticChildProvider(ValueObject &valobj);
 
+  void WillStartExecutingUserExpression();
+  void DidFinishExecutingUserExpression();
+
 protected:
   //------------------------------------------------------------------
   // Classes that inherit from SwiftLanguageRuntime can see and modify these
@@ -444,6 +447,7 @@ protected:
   AppleObjCRuntimeV2 *GetObjCRuntime();
 
   void SetupSwiftError();
+  void SetupExclusivity();
 
   const CompilerType &GetBoxMetadataType();
 
@@ -484,6 +488,12 @@ protected:
       m_bridged_synthetics_map;
 
   CompilerType m_box_metadata_type;
+
+  std::mutex m_active_user_expr_mutex;
+  uint32_t m_active_user_expr_count = 0;
+  llvm::Optional<lldb::addr_t> m_dynamic_exclusivity_flag_addr =
+    llvm::Optional<lldb::addr_t>();
+  bool m_original_dynamic_exclusivity_flag_state = false;
 
 private:
   DISALLOW_COPY_AND_ASSIGN(SwiftLanguageRuntime);
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -489,6 +489,12 @@ protected:
 
   CompilerType m_box_metadata_type;
 
+
+  // These members are used to track and toggle the state of the "dynamic
+  // exclusivity enforcement flag" in the swift runtime. This flag is set to
+  // true when an LLDB expression starts running, and reset to its original
+  // state after that expression (and any other concurrently running
+  // expressions) terminates.
   std::mutex m_active_user_expr_mutex;
   uint32_t m_active_user_expr_count = 0;
   llvm::Optional<lldb::addr_t> m_dynamic_exclusivity_flag_addr =
--- a/packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
+++ b/packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
@@ -90,7 +90,6 @@ class TestExclusivitySuppression(TestBase):
 
         self.assertTrue(len(threads) == 1)
         thread = threads[0]
-        self.assertTrue(frame, "Frame 0 is valid.")
 
         opts = lldb.SBExpressionOptions()
         opts.SetIgnoreBreakpoints(False)
--- a/packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
+++ b/packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
@@ -1,0 +1,145 @@
+# TestExclusivitySuppression.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""
+Test suppression of dynamic exclusivity enforcement
+"""
+import commands
+import lldb
+from lldbsuite.test.lldbtest import *
+import lldbsuite.test.decorators as decorators
+import lldbsuite.test.lldbutil as lldbutil
+import os
+import unittest2
+
+def execute_command(command):
+    (exit_status, output) = commands.getstatusoutput(command)
+    return exit_status
+
+class TestExclusivitySuppression(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
+    # Test that we can evaluate w.s.i at Breakpoint 1 without triggering
+    # a failure due to exclusivity
+    @decorators.swiftTest
+    def test_basic_exclusivity_suppression(self):
+        """Test that exclusively owned values can still be accessed"""
+
+        self.buildAll()
+
+        target = self.create_target()
+
+        # Set the breakpoints
+        bp1 = target.BreakpointCreateBySourceRegex('Breakpoint 1',
+                                                   self.main_source_spec)
+        self.assertTrue(bp1.GetNumLocations() > 0, VALID_BREAKPOINT)
+
+        # Launch the process, and do not stop at the entry point.
+        process = target.LaunchSimple(None, None, os.getcwd())
+
+        self.assertTrue(process, PROCESS_IS_VALID)
+
+        # Frame #0 should be at our breakpoint.
+        threads = lldbutil.get_threads_stopped_at_breakpoint(process, bp1)
+
+        self.assertTrue(len(threads) == 1)
+        thread = threads[0]
+        frame = thread.frames[0]
+        self.assertTrue(frame, "Frame 0 is valid.")
+
+        self.check_expression(frame, "w.s.i", "8", use_summary=False)
+
+    # Test that we properly handle nested expression evaluations by:
+    # (1) Breaking at breakpoint 1
+    # (2) Running 'expr get()' (which will hit breakpoint 2)
+    # (3) Evaluating w.s.i at breakpoint 2 (this is a nested evaluation)
+    # (4) Contininging the evaluation of 'expr get()' to return to bp 1
+    # (5) Evaluating w.s.i again to check that finishing the nested expression
+    #     did not prematurely re-enable exclusivity checks.
+    @decorators.swiftTest
+    def test_exclusivity_suppression_for_concurrent_expressions(self):
+        """Test that exclusively owned values can still be accessed"""
+        self.buildAll()
+
+        target = self.create_target()
+
+        bp1 = target.BreakpointCreateBySourceRegex('Breakpoint 1',
+                                                   self.main_source_spec)
+        self.assertTrue(bp1.GetNumLocations() > 0, VALID_BREAKPOINT)
+
+        bp2 = target.BreakpointCreateBySourceRegex('Breakpoint 2',
+                                                   self.main_source_spec)
+        self.assertTrue(bp2.GetNumLocations() > 0, VALID_BREAKPOINT)
+
+        # Launch the process, and do not stop at the entry point.
+        process = target.LaunchSimple(None, None, os.getcwd())
+
+        self.assertTrue(process, PROCESS_IS_VALID)
+
+        # Frame #0 should be at our breakpoint.
+        threads = lldbutil.get_threads_stopped_at_breakpoint(process, bp1)
+
+        self.assertTrue(len(threads) == 1)
+        thread = threads[0]
+        frame = thread.frames[0]
+        self.assertTrue(frame, "Frame 0 is valid.")
+
+        self.check_expression(frame, "w.s.i", "8", use_summary=False)
+
+
+    @decorators.swiftTest
+    def test_exclusivity_suppression_on_expression_bailout(self):
+        """Test that exclusively owned values can still be accessed"""
+        self.buildAll()
+
+
+    def setUp(self):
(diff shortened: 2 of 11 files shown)


lhames:
@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
```diff
@@ -0,0 +1,145 @@
...
+        # Frame #0 should be at our breakpoint.
+        threads = lldbutil.get_threads_stopped_at_breakpoint(process, bp1)
+
+        self.assertTrue(len(threads) == 1)
+        thread = threads[0]
+        frame = thread.frames[0]
+        self.assertTrue(frame, "Frame 0 is valid.")
+
+        self.check_expression(frame, "w.s.i", "8", use_summary=False)
+
```
jimingham, line 97: I don't think you finished this test.   It's pretty much identical to the first test (except for setting a second breakpoint.)

packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @decorators.swiftTest
+    def test_exclusivity_suppression_on_expression_bailout(self):
+        """Test that exclusively owned values can still be accessed"""
+        self.buildAll()
+
```
jimingham: Ditto for this test.

packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
```diff
@@ -0,0 +1,145 @@
...
+            print value.GetSummary()
+            print value.GetValue()
+        if use_summary:
+            answer = value.GetSummary()
+        else:
+            answer = value.GetValue()
+        report_str = "%s expected: %s got: %s" % (
+            expression, expected_result, answer)
+        self.assertTrue(answer == expected_result, report_str)
+
```
jimingham, line 132: We now have 10 cut and paste copies of this function around.  We should move it to lldbutil.py and use it from there.

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
```diff
@@ -55,6 +55,20 @@ SwiftUserExpression::SwiftUserExpression(
 
 SwiftUserExpression::~SwiftUserExpression() {}
 
+void SwiftUserExpression::WillStartExecuting() {
+  // Can we assert that the process weak-ptr is non-null here?
+  // What's the 'retry_if_fail' argument to GetSwiftLanguageRuntime?
+  // Can the runtime ever come back as null?
+  if (auto process = m_jit_process_wp.lock())
```
jimingham: It should be fine to assert if no process here.  You can run lldb's expression parser in "calculator" mode without a process.  But you can only do that if the expression is IR interpretable.  This code only gets run once we decide we can't IR interpret the expression and are actually going to run it.  For that you have to have a process.

You might be debugging a C program, and somebody tries to run a Swift expression.  There is no swift loaded into your program so no swift language runtime.  The lookup is going to fail, and cache that there's no swift language runtime.  Then you load some library that has swift code.  If retry_if_null is true (the default) we will look for it again - finding it this time.  I actually can't think of a good reason to set this to false off the top of my head.

As above, the runtime can come back as null, so it's probably good to check here.  I can't see how you would be able to get to here without finding a language runtime however, so again it might be appropriate to assert.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -3872,6 +3890,83 @@ SwiftLanguageRuntime::GetBridgedSyntheticChildProvider(ValueObject &valobj) {
...
+    size_t bytes_read =
+      m_process->DoReadMemory(*m_dynamic_exclusivity_flag_addr,
+                              &m_original_dynamic_exclusivity_flag_state,
+                              sizeof(m_original_dynamic_exclusivity_flag_state),
+                              error);
+
+    if (bytes_read != sizeof(m_original_dynamic_exclusivity_flag_state)) {
+      if (log)
+        log->Printf("SwiftLanguageRuntime: Unable to read "
+                    "disableExclusivityChecking flag state");
```
jimingham: Add the error's message to the log message (error.GetCString() will get it for you).

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -3872,6 +3890,83 @@ SwiftLanguageRuntime::GetBridgedSyntheticChildProvider(ValueObject &valobj) {
...
+      size_t bytes_written =
+        m_process->DoWriteMemory(
+                              *m_dynamic_exclusivity_flag_addr,
+                              &new_flag_state,
+                              sizeof(m_original_dynamic_exclusivity_flag_state),
+                              error);
+      if (bytes_written != sizeof(m_original_dynamic_exclusivity_flag_state)) {
+        if (log)
+          log->Printf("SwiftLanguageRuntime: Unable to set "
+                      "disableExclusivityChecking flag state");
```
jimingham: Same here, put the error's message into the log output.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -132,6 +133,23 @@ void SwiftLanguageRuntime::SetupSwiftError() {
       target, g_SwiftNativeNSError, eSymbolTypeObjCClass);
 }
 
+void SwiftLanguageRuntime::SetupExclusivity() {
+  Target &target(m_process->GetTarget());
+
+  ConstString g_disableExclusivityChecking("_swift_disableExclusivityChecking");
+
+  m_dynamic_exclusivity_flag_addr = FindSymbolForSwiftObject(
+      target, g_disableExclusivityChecking, eSymbolTypeData);
```
jimingham, line 142: That API (FindSymbolForSwiftObject) makes me sad.  It does nothing swift specific at all, it just looks up a symbol by name globally.  So no way does it belong in SwiftLanguageRuntime (probably Target instead).  And it doesn't actually return the symbol, it finds the symbol's address.

This function is really the same as the first half of:

Target::FindLoadAddrForNameInSymbolsAndPersistentVariables

Probably better to break that into Target::FindLoadAddrForNameInSymbols and Target::FindLoadAddrForNameInSymbolsAndPersistentVariables.  Then you can use the former for your purposes.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -3872,6 +3890,83 @@ SwiftLanguageRuntime::GetBridgedSyntheticChildProvider(ValueObject &valobj) {
...
+                    "disableExclusivityChecking flag state");
+    } else {
+      bool new_flag_state = true;
+      size_t bytes_written =
+        m_process->DoWriteMemory(
+                              *m_dynamic_exclusivity_flag_addr,
+                              &new_flag_state,
+                              sizeof(m_original_dynamic_exclusivity_flag_state),
+                              error);
+      if (bytes_written != sizeof(m_original_dynamic_exclusivity_flag_state)) {
```
jimingham: Same comment as when reading.  You need to get the sizeof(bool) as above.  Then you should make a Scalar with the value of 1 (scalars don't know about bools, but a C++ bool true is an int with value 1) and you can use WriteScalarToMemory to get the byte swapping right.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -3872,6 +3890,83 @@ SwiftLanguageRuntime::GetBridgedSyntheticChildProvider(ValueObject &valobj) {
...
+
+  if (m_active_user_expr_count == 0 &&
+      m_dynamic_exclusivity_flag_addr) {
+    // We're executing the first user expression. Toggle the flag.
+    Error error;
+    size_t bytes_read =
+      m_process->DoReadMemory(*m_dynamic_exclusivity_flag_addr,
+                              &m_original_dynamic_exclusivity_flag_state,
+                              sizeof(m_original_dynamic_exclusivity_flag_state),
+                              error);
```
jimingham: You are assuming that the size of a Host bool and the size of a Target bool are the same here.  That's not guaranteed to be true.  You can get the TypeSystem from the Target (GetScratchTypeSystemForLanguage is good, that's the expression context.) then get call GetBuiltinTypeByName("bool") then get the size from the type you got back.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -3872,6 +3890,83 @@ SwiftLanguageRuntime::GetBridgedSyntheticChildProvider(ValueObject &valobj) {
...
+    size_t bytes_written =
+      m_process->DoWriteMemory(
+                              *m_dynamic_exclusivity_flag_addr,
+                              &m_original_dynamic_exclusivity_flag_state,
+                              sizeof(m_original_dynamic_exclusivity_flag_state),
+                              error);
+    if (bytes_written != sizeof(m_original_dynamic_exclusivity_flag_state)) {
+      if (log)
+        log->Printf("SwiftLanguageRuntime: Unable to reset "
+                    "disableExclusivityChecking flag state");
```
jimingham: Again, add the error string here.

jimingham (changes requested):
The overall structure is fine.  See the detailed comments below.

packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
lhames: Oops - this was meant to be left out of this commit and added in a follow-up.

I have a "complete" version of this test but it fails due to unexpected behavior in SBProcess::continue - I'll see if I can get that sorted out along with the other issues you've flaged and post the full test.

packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
lhames: As above - I'll see if I can figure out the issue I was encountering with Process continue, then post the full test.

packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
lhames: Sounds good to me. Is that something that would be better done in a separate commit though?

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
lhames: Ok. I'll add the appropriate asserts and remove those comments.

source/Target/SwiftLanguageRuntime.cpp
lhames: Ok. I'll file a bug to refactor this in a follow-up.

source/Target/SwiftLanguageRuntime.cpp
lhames: Sounds good. I'll add the bool size query (and log error messages suggested below).

lhames: Is the type system returned by GetScratchTypeSystemForLanguage going to be C++ or Swift? We need the C++ type size here (though I suspect they're the same in practice).

lhames: Oh - the language is an argument to GetScratchTypeSystemForLanguage. Ignore that then. :)

lhames:
@swift-ci please test

include/lldb/Target/SwiftLanguageRuntime.h
```diff
@@ -485,6 +489,12 @@ class SwiftLanguageRuntime : public LanguageRuntime {
 
   CompilerType m_box_metadata_type;
 
+  std::mutex m_active_user_expr_mutex;
+  uint32_t m_active_user_expr_count = 0;
+  llvm::Optional<lldb::addr_t> m_dynamic_exclusivity_flag_addr =
+    llvm::Optional<lldb::addr_t>();
+  bool m_original_dynamic_exclusivity_flag_state = false;
+
```
jimingham, line 503: Be nice to have a comment describing what these are for.

packages/Python/lldbsuite/test/lang/swift/expression/exclusivity_suppression/TestExclusivitySuppression.py
jimingham: Yes, that's fine.

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
```diff
@@ -55,6 +55,29 @@ SwiftUserExpression::SwiftUserExpression(
 
 SwiftUserExpression::~SwiftUserExpression() {}
 
+void SwiftUserExpression::WillStartExecuting() {
+  // Can we assert that the process weak-ptr is non-null here?
+  // What's the 'retry_if_fail' argument to GetSwiftLanguageRuntime?
+  // Can the runtime ever come back as null?
```
jimingham: You don't need this comment any more, do you?

jimingham (commented):
I don't see the SwiftLangaugeRuntime.cpp changes in the latest diff here.

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
```diff
@@ -59,14 +59,23 @@ void SwiftUserExpression::WillStartExecuting() {
...
-  // Same questions as above.
-  if (auto process = m_jit_process_wp.lock())
-    process->GetSwiftLanguageRuntime()->DidFinishExecutingUserExpression();
+  if (auto process = m_jit_process_wp.lock()) {
+    if (auto swift_runtime = process->GetSwiftLanguageRuntime())
+      swift_runtime->DidFinishExecutingUserExpression();
+    else
+      llvm_unreachable("Can't execute a swift expression without a runtime");
+  } else
+    llvm_unreachable("Can't execute an expression without a process");
```
jimingham: An expression could cause a process to exit.  So while you can't START an expression executing without a process, it is possible that it not exist when you exit.  I am pretty sure you won't get called to here in that case, but it's probably best to just give up quietly on the way out if you don't have a process.

jimingham (commented):
Ah, okay, I found the SwiftLanguageRuntime.cpp changes by going to the latest hash directly.  That part of the diff looks fine.

lhames:
@swift-ci please test

jimingham:
That looks fine.  Please remember to do the cleanups to the FindSymbolForSwiftObject and check_expression in subsequent commits.

jimingham (approved):

PR #202, opened 2017-05-23
Swift returns

jimingham:
The swift calling conventions have changed for swift 4.0, both for regular returns and for error returns.  These changes get lldb to understand the new calling convention.

<rdar://problem/25471028>

### DIFF
--- a/include/lldb/API/SBFunction.h
+++ b/include/lldb/API/SBFunction.h
@@ -54,6 +54,8 @@ public:
   lldb::LanguageType GetLanguage();
 
   bool GetIsOptimized();
+  
+  bool GetCanThrow();
 
   bool operator==(const lldb::SBFunction &rhs) const;
 
--- a/include/lldb/Symbol/Function.h
+++ b/include/lldb/Symbol/Function.h
@@ -346,10 +346,14 @@ public:
   ///
   /// @param[in] range
   ///     The section offset based address for this function.
+  ///
+  /// @param[in] can_throw
+  ///     Pass in true if this is a function know to throw
   //------------------------------------------------------------------
   Function(CompileUnit *comp_unit, lldb::user_id_t func_uid,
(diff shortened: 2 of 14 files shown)


jimingham:
@swift-ci please test

jimingham:
@swift-ci please test

jimingham:
Explanation: The swift calling convention changed for swift 4.0.  In general, lldb doesn't need to know the calling convention because we use the expression parser for most tasks.  But we do display the return value when we "step out", so we need to know how to fetch return types at that point.  The  two changes were that the error return is now in a register, and on x86_64, swift uses 4 registers for complex returns rather than the two the SysV ABI mandates.  This patch adjusts lldb's return fetching mechanism for these cases, and adds some tests. 
Scope: Xcode prints the return or error value in the Locals view on "step out" so this is a visible bit of information.  If we don't get the calling convention right, that UI will present incorrect data.  This is also printed out in the command-line lldb stop message when you step out.  But this is really the only place where this functionality is used in lldb.  For instance, it isn't used to gather the result of expression evaluation.
Radar (and possibly SR Issue): <rdar://problem/25471028>
Risk: Low.  The swift error return code is pretty isolated, and wrong at present.  So this change would at worse make that differently wrong.  The regular return passing code is shared with the C family of languages on x86_64, since those ABI's are mostly the same.  So this could potentially affect that.  But we have tests for C returns which pass after the change, and the code adding additional registers are clearly swift only.  So the chance we messed that up is low.
Testing: The lldb testsuite has tests for gathering swift returns (swift/lang/return).  I added a few tests to that test case.  And there were no changes in a full testsuite run.

jimingham:
The OS X test failure is because I added an SB API to allow the code to be tested, but the CI's don't rebuild the SWIG files, so you have to hand-update them.  I'll do that but the test passes locally so I'll do it as a separate commit.

PR #203, opened 2017-05-23
Handle the UnicodeScalar type rename to Unicode.Scalar.

jimingham:
Have to adjust the name of the type the formatter matches.  It looks
like you can still spell this UnicodeScalar so I left the old match
in place.

<rdar://problem/32149438>

### DIFF
--- a/source/Plugins/Language/Swift/SwiftLanguage.cpp
+++ b/source/Plugins/Language/Swift/SwiftLanguage.cpp
@@ -576,6 +576,10 @@ static void LoadSwiftFormatters(lldb::TypeCategoryImplSP swift_category_sp) {
                 "Swift.UnicodeScalar summary provider",
                 ConstString("Swift.UnicodeScalar"), summary_flags);
   AddCXXSummary(swift_category_sp,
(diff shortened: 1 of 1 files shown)


jimingham:
@swift-ci please test

jimingham:
Explanation: lldb's data formatters match against type name.  The type name for UnicodeScalar was changed to Unicode.Scalar, so the formatter no longer fires.  This fix adds the new type name to the formatter match list.
Scope: Without the change Unicode.Scalar will print as:

  3> let c : Unicode.Scalar = "\u{1112}"
c: Unicode.Scalar = {
  _value = 4370
}

With the patch:

  4> let c : Unicode.Scalar = "\u{1112}" 
c: Unicode.Scalar = U'ᄒ'

Radar (and possibly SR Issue): <rdar://problem/32149438>
Risk: Very low.  If somebody had their own Unicode.Scalar type, this formatter would match it and if it didn't work, they would have to turn it off.  
Testing: The UnicodeScalar formatter was added without a test.  I hand-tested that it does the right thing.  I'll add a test, but I want to get the fix in now.

PR #206, opened 2017-05-30
Preparations for removal of getName on ValueDecl

ahoppen:
This PR goes hand-in-hand with apple/swift#9975


ahoppen:
@jrose-apple I would need this PR merged ASAP as I already merged apple/swift#9975 which relies on this but don't have push access to the swift-lldb repo

PR #207, opened 2017-06-15
not sure how to remove a pr

jasonmolenda:
When lldb is debugging a process on a remote system, it may need to look in the target.exec-search-paths setting to find a binary on the remote host which is not present on the debug host system.  If the binary is in a Darwin bundle, like com.apple.sbd.xpc/com.apple.sbd, and lldb is looking for a binary called "com.apple.sbd", it needs to know to (1) look for "com.apple.sbd.xpc" in the target.exec-search-paths, and (2) to look for "com.apple.sbd.xpc.dSYM" next to that bundle to get the debug information correctly.

The old algorithm for testing the parent directories was using the dot's in the filenames to strip off filename components.

<rdar://problem/31825940>


jasonmolenda:
@swift-ci please test

PR #209, opened 2017-06-15
Correctly find a dSYM for a file with dots in the filename and using target.exec-search-paths

jasonmolenda:
When lldb is debugging a process on a remote system, it may need to look in the target.exec-search-paths setting to find a binary on the remote host which is not present on the debug host system. If the binary is in a Darwin bundle, like com.apple.sbd.xpc/com.apple.sbd, and lldb is looking for a binary called "com.apple.sbd", it needs to know to (1) look for "com.apple.sbd.xpc" in the target.exec-search-paths, and (2) to look for "com.apple.sbd.xpc.dSYM" next to that bundle to get the debug information correctly.

The old algorithm for testing the parent directories was using the dot's in the filenames to strip off filename components.

rdar://problem/31825940


jasonmolenda:
@swift-ci please test

source/Host/common/Symbols.cpp
```diff
@@ -104,38 +103,41 @@ static bool LocateDSYMInVincinityOfExecutable(const ModuleSpec &module_spec,
...
+          // and
+          // /S/L/F/Foundation.framework.dSYM
+          //
+          // so we'll need to start with /S/L/F/Foundation.framework/Versions/A,
+          // add the .dSYM part to the "A", and if that doesn't exist, strip off
+          // the "A" and try it again with "Versions", etc., until we find a dSYM
+          // bundle or we've stripped off enough path components that there's no
+          // need to continue.
+
+          for (int i = 0; i < 4; i++) {
```
scallanan, line 123: It looks to me like all these operations are supported by `FileSpec`, e.g. `CopyByRemovingLastPathComponent` for your `rfind`-based approach and `AppendPathComponent` instead of your `+=`.  In fact, it looks like the FindBundleBinaryInExecSearchPaths uses these APIs.  Can we avoid doing this as `std::string` processing?

source/Plugins/Platform/MacOSX/PlatformDarwin.cpp
```diff
@@ -1945,3 +1946,79 @@ PlatformDarwin::LaunchProcess(lldb_private::ProcessLaunchInfo &launch_info) {
   // Let our parent class do the real launching.
   return PlatformPOSIX::LaunchProcess(launch_info);
 }
+
+lldb_private::Error
+PlatformDarwin::FindBundleBinaryInExecSearchPaths (const ModuleSpec &module_spec, Process *process,
```
scallanan, line 1951: This is largely copied from `PlatformRemoteDarwinDevice.cpp`.  It looks like we're just making sure `PlatformMacOSX` has access to it too; is that correct?  If so, I'm not going to review this logic closely.

scallanan (changes requested):
I have some concerns about the `std::string` implementation style that I mentioned inline below.

I'm also wondering whether this will hold up for sub-frameworks.  A cursory search of the sdk using
```
find . -name "*.framework"
```
shows this pattern commonly used.  4 levels won't be enough there; I suggest repeating until `.framework` no longer appears in your path.

source/Host/common/Symbols.cpp
jasonmolenda: Will look at using a FileSpec instead, thanks.

source/Plugins/Platform/MacOSX/PlatformDarwin.cpp
jasonmolenda: Yeah, exactly.

jasonmolenda:
re. embedded frameworks, that's not a problem.  If you're debugging /S/L/F/Foundation.framework/Versions/A/Frameworks/OtherStuff.framework/OtherStuff, the dSYM will be a sibling of OtherStuff.framework, you don't need to ascend up to the containing bundle.  Also there's no guarantee what the bundle extension will be -- some kext bundles have embedded ("PlugIns") kexts as well.  I haven't seen any Darwin bundle that required going up more than 4 levels.

jasonmolenda:
@swift-ci please test

scallanan (approved):
Thanks, that looks a lot cleaner.  And thanks for the explanation about .dSYM locations for nested frameworks.

clayborg:
Was this first done in in SVN top of tree?

jasonmolenda:
Yeah this is r304520 + r305441 + r305547.

jasonmolenda:
@swift-ci please test

PR #216, opened 2017-06-22
[stable] Swift Runtime Reporting: Change the ABI of the breakpoint function to stop using a boolean argument

kubamracek:
Swift Runtime Reporting: Change the ABI of the breakpoint function to stop using a boolean argument (and replace it with uintptr_t 'flags'). There's apparently problems with retrieving boolean arguments via GetArgumentValues.


kubamracek:
@swift-ci please test

vedantk (approved):
Looks safe to me. It would be worth adding in a comment explaining what exactly goes wrong when using eBasicTypeBool.

kubamracek:
I'll submit an assert() with a comment upstream.  This seems to be limitation/bug of GetArgumentValues, which apparently was never meant to understand all possible types.

vedantk:
Sounds great, thanks.

PR #218, opened 2017-06-23
[4.0] Improve reporting of Swift @objc inference issues, include details about method declaration and fix-its.

kubamracek:


kubamracek:
@swift-ci please test

PR #219, opened 2017-06-28
Main Thread Checker: When in Swift, and the report is describing a call to an setter or getter of a property, use the property name instead of the method name

kubamracek:
Main Thread Checker: When in Swift, and the report is describing a call to an setter or getter of a property, use the property name instead of the method name.  We currently report things like `NSView.setFrame(_:)`, which isn't correct:  The call is user's code looks like `myview.frame = ...`.  Let's look up the actual property name and use it in the message instead.

### DIFF
--- a/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
+++ b/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
@@ -123,8 +123,8 @@ static std::string TranslateObjCNameToSwiftName(std::string className,
         // We need to loadAllMembers(), otherwise 'isAccessor' returns false.
         if (auto extension = clang::dyn_cast<swift::ExtensionDecl>(funcCtx)) {
           extension->loadAllMembers();
-        }
-        if (auto nominal = clang::dyn_cast<swift::NominalTypeDecl>(funcCtx)) {
+        } else if (auto nominal =
+                       clang::dyn_cast<swift::NominalTypeDecl>(funcCtx)) {
           nominal->loadAllMembers();
         }
 
--- a/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
+++ b/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
@@ -109,7 +109,7 @@ static std::string TranslateObjCNameToSwiftName(std::string className,
                            swift::DeclVisibilityKind Reason) {
       if (result)
         return; // Take the first result.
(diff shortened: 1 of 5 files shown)


source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
```diff
@@ -109,12 +109,32 @@ static std::string TranslateObjCNameToSwiftName(std::string className,
...
+      swift::AbstractFunctionDecl *decl = funcs.front();
+      if (auto func = clang::dyn_cast<swift::FuncDecl>(decl)) {
+        swift::DeclContext *funcCtx = func->getParent();
+        // We need to loadAllMembers(), otherwise 'isAccessor' returns false.
+        if (auto extension = clang::dyn_cast<swift::ExtensionDecl>(funcCtx)) {
+          extension->loadAllMembers();
+        }
+        if (auto nominal = clang::dyn_cast<swift::NominalTypeDecl>(funcCtx)) {
+          nominal->loadAllMembers();
+        }
```
vedantk: It looks like ExtensionDecl and NominalTypeDecl are mutually non-inheriting, so it should be fine to use an 'else if' here?

kubamracek:
@swift-ci please test

vedantk (approved):
This looks good to me.

As far as I know, we don't yet have CI available which lets us test this change, so I we'll have to wait for a test case.

source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
```diff
@@ -109,12 +109,32 @@ static std::string TranslateObjCNameToSwiftName(std::string className,
                            swift::DeclVisibilityKind Reason) {
       if (result)
         return; // Take the first result.
-      if (swift::ClassDecl *cls = clang::dyn_cast<swift::ClassDecl>(VD)) {
-        auto funcs = cls->lookupDirect(selectorToLookup, true);
-        if (funcs.size() > 0) {
-          result = funcs.front()->getFullName();
+      swift::ClassDecl *cls = clang::dyn_cast<swift::ClassDecl>(VD);
```
jrose-apple: Nitpick: these should all be llvm::dyn_cast.

jrose-apple (approved):
Looks good to me as well.

kubamracek:
I'm now adding testcases for MTC, see <https://github.com/apple/swift-lldb/pull/224>.

kubamracek:
Added a testcase.

kubamracek:
@swift-ci please test

kubamracek:
@swift-ci please test

PR #220, opened 2017-06-28
Fix up swift stepping in the new mangling

jimingham:
I got the swift stepping test running with an intermediate version of the new mangling, but something changed on the swift side and broke these tests.  This gets them running again.

The stepping out also ran into some additional difficulties because we have debug info for the swift standard library, which the stepping tests also have to successfully get out of.  I fixed a few places in the step out from here generation to make this more robust.

### DIFF
--- a/include/lldb/Target/LanguageRuntime.h
+++ b/include/lldb/Target/LanguageRuntime.h
@@ -164,6 +164,9 @@ public:
   virtual bool GetIRPasses(LLVMUserExpression::IRPasses &custom_passes) {
     return false;
   }
+  
+  static bool
+  IsSymbolAnyRuntimeThunk(lldb::ProcessSP process, Symbol &symbol);
 
 protected:
   //------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


jimingham:
Explanation: lldb relies on the demangled forms of various inter-call thunks to figure out how to follow them into the actual function being called.  The logic for doing this was change for Swift 4.0.  lldb was changed to accommodate those changes but the swift side hadn't settled down, and subsequent changes broke the stepping test cases we had.  This PR gets those tests back to passing.
Scope: This change has two parts.  One was to accommodate the new mangling.  That code just makes correct the decoding of the thunks & figuring out what to do with them.  The other part is to make the procedure used to walk out of places we decide we don't want to stop in simpler when possible, so that it would be more robust.
Radar (and possibly SR Issue): <rdar://problem/31515444>
Risk: This is a complex area of lldb, so touching it is never going to be low risk, but these changes are pretty straightforward changes, and there are a lot of tests for stepping in the lldb testsuite.  So this is a low-medium risk change.  
Testing: The failing tests now pass, and none of the other lldb stepping tests were affected.

jimingham:
@swift-ci please test

PR #221, opened 2017-06-29
[Swift] Update VarDecl to use specifiers

CodaFi:
Complements apple/swift#10695


jimingham:
Wouldn't it be better to convert VariableInfo::GetIsLet to VariableInfo::GetVarSpecifier and then just pass that to the VarDecl?  That is currently the only use of this function, and if you did it that way, when you did add more variable types, there's a good chance they will just work...

CodaFi:
Sure, I can make that change here.  I just wanted something minimal for now because LLDB is only creating `VarDecl`s, and the more immediate addition of specifiers will be to `ParamDecl`s.

CodaFi:
⛵️

PR #223, opened 2017-07-05
[4.0] Add a testcase for MainThreadCheckerRuntime plugin

kubamracek:
Cherry-pick of r 307170 .  This adds a simple testcase for MainThreadCheckerRuntime. The tool (Main Thread Checker) is only available on Darwin, so the test also detects the presence of libMainThreadChecker.dylib and is skipped if the tool is not available.


kubamracek:
@swift-ci please test

vedantk (approved):
LGTM

PR #226, opened 2017-07-05
Bool summary should ignore spare bits

jimingham:
The swift Bool only considers the LSB significant.  So the formatter should mask off all the other bits before interpreting the value.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/variables/bool/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/variables/bool/Makefile
@@ -1,0 +1,5 @@
+LEVEL = ../../../../make
+
+SWIFT_SOURCES := main.swift
+
+include $(LEVEL)/Makefile.rules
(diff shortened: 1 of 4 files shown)


jimingham:
@swift-ci please test

jasonmolenda (approved):
Looks good, nice test case.

jimingham:
Explanation: Only the LSB of a Swift Bool is significant.  The lldb data summary for Swift Bool's didn't know that so sometimes it would report a Bool as "invalid".  Need to mask the value before reporting "true" or "false".
Scope: This is a very isolated change that only affects the Swift Bool summary provider.  It prevents us presenting a confusing summary for Swift Bool's.
Radar (and possibly SR Issue): <rdar://problem/29605924> ObjCBool: boolValue displays as (Bool) <invalid> (0xfe)
Risk: Very low, this is an isolated change, and I've added a test case so if things change on the swift side we should catch it either by the test failing or the test failing to compile or assert.
Testing: We didn't have a test case for the Swift Bool summary provider.  This change adds one.

PR #227, opened 2017-07-06
Don't treat DW_AT_const_value as meaning "static" in Swift.

scallanan:
This is a heuristic that works for code compiled with Clang, but the
Swift language is different (as explained in an added comment).

I've also added a test that checks this.

rdar://31332056

### DIFF
--- a/source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
+++ b/source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
@@ -4269,7 +4269,7 @@ VariableSP SymbolFileDWARF::ParseVariableDIE(const SymbolContext &sc,
           }
         } else {
           // The heuristic for inferring static variables works for Clang's
-          // behavior on C-like languages, which generallydoes n ot emit
+          // behavior on C-like languages, which generally does not emit
           // AT_const_value for locals.
           //
           // However, the Swift compiler can and does emit AT_const_value for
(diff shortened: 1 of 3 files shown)


scallanan:
@swift-ci please test

source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
```diff
@@ -4268,7 +4268,15 @@ VariableSP SymbolFileDWARF::ParseVariableDIE(const SymbolContext &sc,
             }
           }
         } else {
-          if (location_is_const_value_data)
+          // The heuristic for inferring static variables works for Clang's
+          // behavior on C-like languages, which generallydoes n ot emit
```
jasonmolenda: your keyboard had some sync issues with the space bar.

jimingham:
What happens with module-level const variables?  In:

```
let g_int : Int = 10

func use_g_int()
{
    let local_int : Int = 20
    print(g_int, local_int)
}

use_g_int()

```

When stopped at the print above, does g_int show up in "frame var"?

scallanan:
I will add that to the test, thanks Jim.  I will also fix the misplaced space in the comment, thanks Jason!

scallanan:
Updated.  The code actually is only distinguishing between locals and statics; globals follow a different path which this doesn't touch at all.  I've added a global to the test and made sure it doesn't sneak its way into the list, though, so that this assumption is documented.

jimingham (approved):
Great.  Then this looks good to me.

PR #229, opened 2017-07-06
Restore mutable variables as the default

CodaFi:
To get the tests going again.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.h
+++ b/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.h
@@ -63,7 +63,7 @@ public:
 
     VariableInfo(CompilerType &type, swift::Identifier name,
                  VariableMetadataSP metadata, 
-                 swift::VarDecl::Specifier specifier = swift::VarDecl::Specifier::Let,
+                 swift::VarDecl::Specifier specifier, 
                  bool is_capture_list = false)
         : m_type(type), m_name(name), m_var_specifier(specifier),
           m_is_capture_list(is_capture_list), m_metadata(metadata) {}
@@ -82,7 +82,7 @@ public:
     CompilerType m_type;
     swift::Identifier m_name;
     swift::VarDecl *m_decl = nullptr;
-    swift::VarDecl::Specifier m_var_specifier = swift::VarDecl::Specifier::Let;
+    swift::VarDecl::Specifier m_var_specifier = swift::VarDecl::Specifier::Var;
     bool m_is_capture_list = false;
 
   public:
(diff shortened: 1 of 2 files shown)


CodaFi:
@swift-ci please test

CodaFi:
It seems like the REPL-related tests pass at least.

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.h
```diff
@@ -63,7 +63,7 @@ class SwiftASTManipulatorBase {
 
     VariableInfo(CompilerType &type, swift::Identifier name,
                  VariableMetadataSP metadata, 
-                 swift::VarDecl::Specifier specifier = swift::VarDecl::Specifier::Let,
+                 swift::VarDecl::Specifier specifier = swift::VarDecl::Specifier::Var,
```
jimingham: Why are you defining a variable in order to pass a constant to this constructor?

jimingham (changes requested):
The change is right, we always want to make var's not let's.  But the first change looks odd.  Am I missing some subtle reason for doing it this way?

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.h
CodaFi: The code that used to be here was

```
bool is_let = false
```

I just translated it to the new representation.

jimingham: Ah, we have the convention that if you are calling a function with a bunch of arguments, instead of doing:

`ComplexFunction(false, false, true, false);
`
we do

```
bool does_thing_a = true;
bool does_thing_b = false;
...
ComplexFunction(does_thing_a, does_thing_b, ...)

```

To make the code easier to read.  Somebody was being lazy and inlining the variable definition.  That was an awkward way to write that in the first place, but also, since you are passing this enum and not a "true" or "false" it documents itself so the variable isn't necessary.

Can you remove that and then this is fine.

CodaFi:
@swift-ci please test

CodaFi:
@swift-ci please test

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.h
```diff
@@ -63,7 +63,7 @@ class SwiftASTManipulatorBase {
 
     VariableInfo(CompilerType &type, swift::Identifier name,
                  VariableMetadataSP metadata, 
-                 swift::VarDecl::Specifier specifier = swift::VarDecl::Specifier::Let,
+                 swift::VarDecl::Specifier specifier,
```
jimingham, line 66: Aren't you just passing an uninitialized variable now?

CodaFi: Uninitialized?  This is a parameter to a constructor.

jimingham: Eh, sorry, for some reason I thought you had changed this as requested, and the window elided the "= ..." so I though you had removed the wrong part.

Anyway, please remove the variable before committing.

CodaFi:
After consulting in person, :shipit:

PR #231, opened 2017-07-07
[4.0] Main Thread Checker: When in Swift, and the report is describing a call to an setter or getter of a property, use the property name instead of the method name

kubamracek:
This is cherry-pick of  28ad4cf into swift-4.0-branch:  Main Thread Checker: When in Swift, and the report is describing a call to an setter or getter of a property, use the property name instead of the method name. We currently report things like NSView.setFrame(_:), which isn't correct: The call is user's code looks like myview.frame = .... Let's look up the actual property name and use it in the message instead.


kubamracek:
@swift-ci please test

PR #232, opened 2017-07-10
X86 unwind buffer limit checks

jasonmolenda:
The x86 instruction unwinder can be asked to disassemble non-instruction
blocks of memory, and if the final bytes of that block look like a long
x86 instruction, it can cause the llvm disassembler to read past the end
of the buffer. Use the maximum allowed instruction length that we pass
to the llvm disassembler as a way to limit this to the size of the buffer.
An example of how to trigger this is when lldb does a function call, it
puts a breakpoint on the beginning of main() and uses that as the return
address from the function call. When we stop at that location, lldb may
try to find the first frame up the stack. Because this is on the first
instruction of a function, it will get the word-size value at the stack
pointer and assume that this was the caller's pc value. But this is random
stack memory and could point to anything - an object in memory, something
in the data section, whatever. And if we have a symbol for that thing,
we'll try to disassemble it.
This was leading to infrequent crashes in customer scenarios; figured out
what was happening with address sanitizer.
rdar://problem/30463256


jasonmolenda:
@swift-ci please test

source/Plugins/UnwindAssembly/x86/x86AssemblyInspectionEngine.cpp
```diff
@@ -577,9 +577,10 @@ uint32_t x86AssemblyInspectionEngine::extract_4(uint8_t *b) {
 }
 
 bool x86AssemblyInspectionEngine::instruction_length(uint8_t *insn_p,
-                                                     int &length) {
+                                                     int &length, 
+                                                     uint32_t buffer_remaining_bytes) {
 
-  const uint32_t max_op_byte_size = m_arch.GetMaximumOpcodeByteSize();
+  uint32_t max_op_byte_size = std::min(buffer_remaining_bytes, m_arch.GetMaximumOpcodeByteSize());
```
jimingham, line 583: Did you need to remove the const here?  It doesn't matter much, but it shouldn't have been necessary.

jimingham (approved):
This looks good, thanks for adding the test.  I had one trivial inline comment.

PR #233, opened 2017-07-10
Swift 4.0 branch stringref perf

llvm-beanz:
This back ports a fix for a severe performance regression in LLDB that was introduced in LLVM 4.0.

### DIFF
--- a/source/Symbol/Symtab.cpp
+++ b/source/Symbol/Symtab.cpp
@@ -265,8 +265,8 @@ void Symtab::InitNameIndexes() {
         continue;
 
       const Mangled &mangled = symbol->GetMangled();
-      entry.cstring = mangled.GetMangledName().GetStringRef();
-      if (!entry.cstring.empty()) {
+      entry.cstring = mangled.GetMangledName();
+      if (entry.cstring) {
         m_name_to_index.Append(entry);
 
         // Now try and figure out the basename and figure out if the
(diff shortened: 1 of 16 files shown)


llvm-beanz:
@swift-ci please test

source/Symbol/Symtab.cpp
```diff
@@ -265,46 +265,48 @@ void Symtab::InitNameIndexes() {
...
         const SymbolType symbol_type = symbol->GetType();
         if (symbol_type == eSymbolTypeCode ||
             symbol_type == eSymbolTypeResolver) {
-          if (name.size() >= 2 && name[0] == '_' && name[1] == 'Z' &&
-              ((name.size() < 3) || (name[2] != 'T' && // avoid virtual table, VTT structure, typeinfo
-                                 // structure, and typeinfo name
-               name[2] != 'G' && // avoid guard variables
-               name[2] != 'Z'))) // named local entities (if we eventually handle
-                                // eSymbolTypeData, we will want this back)
+          llvm::StringRef entry_ref(entry.cstring.GetStringRef());
```
jimingham, line 287: This actually changes the behavior from the original.  In the original, we did:

1. Store the mangled name in `entry.cstring`
2. Make a string ref for the original mangled name (`name`).
3. Strip the linker annotations, put the stripped version back in "entry.cstring" and add that to the index
4. Check `name` (still the unstripped mangled name) for _, _Z, etc...

In the new version, since you are checking `entry_ref` which you get from `entry.cstring` on line 287 rather than`name`, you're now applying the logic of step 4. to the name with linker symbol annotations stripped.

Is that intentional?

jimingham (changes requested):
This looks good in general.  It seems like it is really three changes:

1) The StringRef->ConstString for the UniqueCStringMap.  That's mostly formal and looks correct.
2) The return value for StripLinkerSymbolAnnotations went from string->StringRef.  This avoids a copy which is good, but only for symbols that have linker annotations, so it's probably a minor thing.  
I have one question about the effects of this conversion inline.  It also seems a little odd not to change the argument back to a ConstString since that's how it gets used in all its use sites, but that's not a big deal and not worth deviating from SVN for.
3) Some CMake changes.  I don't see how 1 & 2  necessitated this change?

source/Symbol/Symtab.cpp
llvm-beanz: This is how the code is in llvm.org, and StripLinkerSymbolAnnotations strips off leading annotations that occur before the actual start of the mangled symbol. I suspect that was a bug fixed along the way, but since StripLinkerSymbolAnnotations only actually does anything on ELF, we it should be relatively safe to do this.

llvm-beanz:
The CMake changes are not strictly necessary for this patch, but without them you can't build properly using CMake on Darwin, so I think they are worth pulling in and they shouldn't destabilize the release in any way. If you'd prefer I can remove them but since we build LLDB with Xcode on Darwin they shouldn't have any adverse impact.

jimingham (approved):
Doesn't seem worth pulling out the cmake changes as long as you verified that the Xcode build still works.

PR #234, opened 2017-07-12
When evaluation user expressions, ignore InstrumentationRuntime breakpoints

kubamracek:


kubamracek:
@swift-ci please test

packages/Python/lldbsuite/test/functionalities/ubsan/user-expression/TestUbsanUserExpression.py
```diff
@@ -0,0 +1,49 @@
...
+        self.expect(
+            "file " + exe,
+            patterns=["Current executable set to .*a.out"])
+
+        self.runCmd("breakpoint set -f main.c -l %d" % self.line_breakpoint)
+
+        self.runCmd("run")
+
+        process = self.dbg.GetSelectedTarget().process
+        thread = process.GetSelectedThread()
```
jimingham, line 40: It is better to set the breakpoint with SBTarget::BreakpointSetByLocation, which returns the breakpoint set, and then assert that you've hit the breakpoint you intended, and that the thread you got is actually the thread that hit the breakpoint, by using lldbutil.get_threads_stopped_at_breakpoint.  That way if something happens like your main thread hits a breakpoint but some secondary thread some system decided ought to always be created crashes, you will see that right away.

I added an lldbutil routine to llvm.org (lldbutil.run_to_source_breakpoint) that packages all this up more conveniently.  When we merge that over to github, that will be the preferred way to gather up all this boilerplate.

jimingham (approved):
This is fine as a short-term solution if hitting these breakpoints is causing problems to clients of lldb (e.g. this would be fine for submission to the swift-4.0 branch).  

But this behavior should really be controlled by the --ignore-breakpoints flag.  Actually, I'm a little curious why the InstrumentationRuntime breakpoints don't get gathered up in this logic.  So for stable we should figure that out and make the --ignore-breakpoints flag control this.  Or maybe, since people in general don't know that the IR actions are implemented by breakpoints, it might be better to add a --ignore-instrumentation-runtime-events flag to make this clearer.

PR #239, opened 2017-07-18
The go OperatingSystem plugin can crash when the application

jasonmolenda:
has some third-party go code linked in to it, possibly created
with a different version of Go than lldb is currently able to
handle.

Disable the automatic creation of threads for goroutines
(plugin.os.goroutines.enable defaults to false) and also harden
the routine that was crashing when the goroutine data structures
weren't as expected.

Go developers can re-enable the goroutines thread creation by
changing the setting.

<rdar://problem/31451900>


jasonmolenda:
@swift-ci please test

scallanan (approved):
Looks good to me.

PR #241, opened 2017-07-20
[Swift] Fix 'frame var -O' for inout parameters

slavapestov:
Needs a test...


slavapestov:
@jimingham Can you suggest a good place to add a test for this? I just need to 'frame variable -O someInOutParam' and make sure we don't crash.

jimingham:
There's already a test for inout variables:

packages/Python/lldbsuite/test/lang/swift/variables/inout

It has a bunch of different functions with inout parameters.  Should be easy to add a test to that.

Jim

PR #242, opened 2017-07-20
[Swift] Fix 'frame var -O' for inout parameters [4.0]

slavapestov:
* Description: Fixes a crash when attempting to pass an LValue type to SIL type lowering, which is not allowed. Just unwrap the lvalue type first.

* Scope of the issue: Reported externally.

* Origination: Probably been there for a while, but at one point lldb had a "legal SIL type" check. That's totally wrong, though.

* Risk: Very low.

* Radar: Fixes <rdar://problem/33296509>

* Reviewed by: @jimingham


slavapestov:
@swift-ci Please test

PR #244, opened 2017-07-25
[Swift] Make classes and members of classes 'open' instead of 'public' [4.0]

slavapestov:
Fixes <rdar://problem/29039602>.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
@@ -835,10 +835,16 @@ void SwiftASTManipulator::MakeDeclarationsPublic() {
     virtual bool walkToDeclPre(swift::Decl *decl) {
       if (swift::ValueDecl *value_decl =
               llvm::dyn_cast<swift::ValueDecl>(decl)) {
-        value_decl->overwriteAccessibility(swift::Accessibility::Public);
+        auto access = swift::Accessibility::Public;
+        if (swift::isa<swift::ClassDecl>(value_decl) ||
+            swift::isa<swift::ClassDecl>(value_decl->getDeclContext())) {
+          access = swift::Accessibility::Open;
+        }
+
+        value_decl->overwriteAccessibility(access);
         if (swift::AbstractStorageDecl *var_decl =
                 llvm::dyn_cast<swift::AbstractStorageDecl>(decl))
-          var_decl->overwriteSetterAccessibility(swift::Accessibility::Public);
+          var_decl->overwriteSetterAccessibility(access);
       }
 
       return true;
--- a/packages/Python/lldbsuite/test/repl/breakpoints/TestREPLBreakpoints.py
+++ b/packages/Python/lldbsuite/test/repl/breakpoints/TestREPLBreakpoints.py
@@ -16,12 +16,12 @@ import time
 import unittest2
 import lldb
 import lldbsuite.test.decorators as decorators
-from lldbsuite.test.lldbrepl import REPLTest, load_tests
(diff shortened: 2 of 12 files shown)


slavapestov:
@jrose-apple Is this what you had in mind?

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
```diff
@@ -835,10 +835,16 @@ void SwiftASTManipulator::MakeDeclarationsPublic() {
     virtual bool walkToDeclPre(swift::Decl *decl) {
       if (swift::ValueDecl *value_decl =
               llvm::dyn_cast<swift::ValueDecl>(decl)) {
-        value_decl->overwriteAccessibility(swift::Accessibility::Public);
+        auto access = swift::Accessibility::Public;
+        if (swift::isa<swift::ClassDecl>(value_decl) ||
+            swift::isa<swift::ClassDecl>(value_decl->getDeclContext())) {
```
jrose-apple, line 840: Nitpick: `llvm::isa`. (We just borrow it into our namespace.)

jrose-apple (commented):
Yep, strictly an improvement over what we have today. Still would want to figure out what to do with extension methods (`@objc` ones can be overridden), but this is good enough and won't put complicated logic into LLDB.

jrose-apple:
Needs a test case, of course.

jimingham:
This looks fine. The test case repl/subclassing should test exactly this case.  The commands typed in by hand clearly fail.

jimingham (commented):
This looks fine except we have to figure out why the repl/subclassing test didn't catch the failure.  When you type its commands in by hand it clearly fails.

jrose-apple:
Looks like that test is XFAILed for unrelated reasons: https://github.com/apple/swift-lldb/blob/stable/packages/Python/lldbsuite/test/repl/subclassing/TestREPLSubclassing.py#L28

jimingham:
Grr...  Yeah, these tests were originally written using pexpect, but pexpect turned out to be weirdly flakey.  We need to convert all these tests to use the expression parser in repl mode.

jimingham:
I take that back, looks like Enrico did make these REPLTests use the expression parser in REPL mode rather than pexpect.  The x-fail was just for flakiness of the pexpect mechanism, so it looks like the missing step was to un-xfail this test.

jimingham:
To get the record straight, it was Sean that did the conversion...

slavapestov:
@jimingham If I re-enable the test, it does not fail. It looks like dotest.py doesn't run REPL tests at all?

slavapestov:
It looks like dotest runs a test if its defined like so:

```class REPLSubclassingTestCase (lldbrepl.REPLTest)```

But not like so

```class REPLSubclassingTestCase (REPLTest)```

slavapestov:
@swift-ci Please test

slavapestov:
Ok, here is a new version with tests. @jimingham and @jrose-apple does this look OK for 4.0?

slavapestov:
Also I ran the tests locally but I don't know how to trigger CI here...

slavapestov:
@swift-ci Please test

slavapestov:
@swift-ci Please test

jimingham (approved):
I don't have any objections, but I defer to Jordan for whether "open" is the best way to mark classes in the current REPL, his knowledge of the swift visibility model far exceeds mine.

jrose-apple:
Slava came up with a few other cases that still won't work, but they're not regressions. This seems like a strict improvement to me, so I'm 👍 as well.

jimingham:
Slava, can you file bugs on the cases that don't work?  Maybe we need to rethink how we accumulate code in the REPL to get this all to work smoothly.

PR #246, opened 2017-08-01
[testsuite] Update TestREPLSubclassing to be an inline REPL test

scallanan:
This fixes pexpect-related test failures in CI and makes the test more predictable.


jimingham (approved):
This seems fine.

PR #248, opened 2017-08-22
[4.0] Correctly escape newlines and backslashes in the JSON serializer

kubamracek:
JSON serializer fails to escape newlines and backslashes. Let's fix that.


kubamracek:
@swift-ci please test

jasonmolenda:
Previously reviewed this in https://reviews.llvm.org/D34322 - this still looks OK to me.  Normally we'd break  'if (ch == '\n') ch = 'n';'  across two source lines in lldb but I didn't care enough back in June to mention it and I still am not that worried about it.

jasonmolenda (approved):
Previously reviewed this in https://reviews.llvm.org/D34322 - this still looks OK to me. Normally we'd break 'if (ch == '\n') ch = 'n';' across two source lines in lldb but I didn't care enough back in June to mention it and I still am not that worried about it.

PR #251, opened 2017-09-08
Fix updated constructor calls of swift::SourceFile. rdar://34339199

nkcsgexi:


nkcsgexi:
@swift-ci please test

PR #252, opened 2017-09-27
Apply local llvm/clang/swift patches when using cmake build system

fjricci:
This is already supported in the Xcode and bash build scripts, add
to the cmake build scripts for parity.

### DIFF
--- a/scripts/build-swift-cmake.py
+++ b/scripts/build-swift-cmake.py
@@ -13,6 +13,7 @@
 # ------------------------------------------------------------------------------
 
 import argparse
+import fnmatch
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


jimingham (approved):
The Xcode build does this job, so it seems appropriate for the cmake build to do so as well.  

However, we want to avoid if at all possible having to maintain patchsets against other tools.  So while the mechanism for local development purposes is fine, particularly as you are working patches into the other tools, there will be a pretty high bar to actually adding .diff files to the github repository.

fjricci:
Yeah, we use the patchset feature for some internal changes and build system peculiarities, but I agree that it doesn't make sense in general for upstream work.

jasonmolenda:
I've needed to maintain patches to llvm/clang/swift that I couldn't get upstreamed for a short period in the past, I've had similar code on branches to do this too. The main question I have about the patch is the 'patch -p0' - normally git diffs need -p1 don't they?  Are the patches from svn, or is there some way to make git diff not do that?

fjricci:
Yeah, I've debated switching things to -p1 myself, but generally assumed we chose -p0 for some reason. I'd be happy to put up a diff for -p1 (it would make my life a bit easier for sure)

fjricci:
Also, would one of you mind merging this commit? I don't yet have swift commit access

jasonmolenda:
ach, I also committed your matching phabracator change on llvm.org, sorry looking at too many different branches today and I forgot that only the github change needed to be committed.

PR #253, opened 2017-09-28
Update swift-cmake build script patcher to match llvm r314366

fjricci:
Assumes git patch files instead of svn.


fjricci:
If I understand the branching scheme correctly, I think this should be merged into upstream-with-swift (since that has r314366), but not stable, since that doesn't yet have the llvm revision in question.

fjricci:
Not sure how to request a review for this, but should be trivial

PR #259, opened 2017-11-22
Fix Swift dictionary padding with enum keys

keith:
When you have a dictionary which has an enum as the key such as:

```
enum Foo: String {
  case a
}

let bar = [Foo.a: 1]
```

The stride of the key is 1 byte. This differs from if you have a tuple
with the same key and value as the dictionary such as:

```
let baz = (Foo.a: 1)
```

In order to read the values from a dictionary we create a tuple which
expects 16 bytes of data (in the case the key is an enum and the value
is an int). Previously because the key was only 1 byte, we only filled
it with 9 bytes, and therefore the offsets were wrong leading to us
printing garbage data. Now instead of just using the stride of the key,
we use the stride based on the tuple type we generate for printing these
values. This works since the buffer is initialized to all 0 and
therefore acts as padding for the key.

This fixes https://bugs.swift.org/browse/SR-6290

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/hashed_containers_enums/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/hashed_containers_enums/Makefile
@@ -1,0 +1,5 @@
+LEVEL = ../../../make
+
+SWIFT_SOURCES := main.swift
+
+include $(LEVEL)/Makefile.rules
--- a/packages/Python/lldbsuite/test/lang/swift/hashed_containers_enums/TestSwiftHashedContainerEnum.py
+++ b/packages/Python/lldbsuite/test/lang/swift/hashed_containers_enums/TestSwiftHashedContainerEnum.py
@@ -1,0 +1,96 @@
+# TestSwiftHashedContainerEnum.py
+
+"""
+Test combinations of hashed swift containers with enums as keys/values
+"""
+
+import lldb
+from lldbsuite.test.lldbtest import *
+import lldbsuite.test.decorators as decorators
+import lldbsuite.test.lldbutil as lldbutil
+import os
+import unittest2
+
+class TestSwiftHashedContainerEnum(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
+    @decorators.skipUnlessDarwin
+    @decorators.swiftTest
+    def test_any_object_type(self):
+        """Test combinations of hashed swift containers with enums"""
+        self.build()
+        self.do_test()
+
+    def setUp(self):
+        TestBase.setUp(self)
+        self.main_source = "main.swift"
+        self.main_source_spec = lldb.SBFileSpec(self.main_source)
+
+    def do_test(self):
+        """Test combinations of hashed swift containers with enums"""
+        exe_name = "a.out"
+        exe = os.path.join(os.getcwd(), exe_name)
+
+        # Create the target
+        target = self.dbg.CreateTarget(exe)
+        self.assertTrue(target, VALID_TARGET)
+
+        # Set the breakpoints
+        breakpoint = target.BreakpointCreateBySourceRegex(
+            '// break here', self.main_source_spec)
+        self.assertTrue(breakpoint.GetNumLocations() > 0, VALID_BREAKPOINT)
+
+        # Launch the process, and do not stop at the entry point.
+        process = target.LaunchSimple(None, None, os.getcwd())
+
+        self.assertTrue(process, PROCESS_IS_VALID)
+
+        # Frame #0 should be at our breakpoint.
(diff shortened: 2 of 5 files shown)


keith:
To test this I've been using this file:

```swift
enum Enum {
    case a, b, c
}

let testA = [Enum.c: 1, Enum.b: 2]
let testB = ["a": 1, "b": 2]
let testC = (key: Enum.b, value: 2)
let testD = Set([Enum.c])
print(testA)
```

And this `lldbinit` for convenience:

```
br set -f foo.swift -l 9
br command add 1 -o "fr v testA testB testC testD"
r
```

lattner:
Great fix @keith!!

DougGregor:
@swift-ci please test

keith:
Looks like the linux failures here might be unrelated (although I'm not sure if this is the actual failure cause):

```
fatal error: error in backend: IO failure on output stream.
```

The macOS failure looks like it could be related so I'm trying to repro these locally.

keith:
The macOS tests seem to pass for me locally on my branch. It looks like based on the log this one might have failed because of a timeout:

```
[TestSwiftDictionaryNSObjectAnyObject.py FAILED] (TIMEOUT)
```

Can we run these again if that makes sense?

CodaFi:
I'll kick off the tests, but that failure looks concerning.  It seems the printer is straying pretty far outside the bounds of the object it should be reading. I would guess your local setup would have trouble reproducing this because it would appear nondeterministically.

@swift-ci please test

keith:
Anyone have thoughts on why this would fail on CI both runs, but not fail locally in my testing?

keith:
I've submitted a related PR adding tests for this case. Since they should be valuable regardless of this fix (minus the one that is disabled until this bug is fixed) I thought that made sense to be in a separate PR https://github.com/apple/swift-lldb/pull/262

keith:
Ah it looks like some of the tests that failed here were disabled in this commit https://github.com/apple/swift-lldb/commit/909ec3e70551d7c5eb40632bd5020ebb87e32f77

keith:
I've rebased on the stable branch which has these changes, so if someone could run the tests again here they should be green this time (assuming the linux OOM failures have been fixed as well)

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

keith:
Is this linux failure a known issue?

```
clang: error: unable to execute command: Killed
clang: error: linker command failed due to signal (use -v to see invocation)
```

This is similar to how it has failed on Linux all 3 test runs so far

dduan:
@swift-ci please test

keith:
Looks like linux failed again for the same reason 🤔

CodaFi:
Linux failure doesn’t look infrastructure related.  @jimingham any ideas?

dcci (changes requested):
Hi @Keith, is there any change of adding a test for this?
Also, I looked at the failure and I'm pretty confident it's unrelated. 

Can you please consider rebasing the following patch on ToT? At that point I'll spawn another build. Sorry for hanging you for so long, but it seems there are more people looking at Swift debug support in LLDB than there used to be until very recently :)

dcci:
Nvm, I just noticed you submitted another pull request for the test. My comment on rebasing still stands.

keith:
@dcci I've rebased on `stable` (let me know if that's not what you meant) and added the test from https://github.com/apple/swift-lldb/pull/262 here instead.

Let me know if there's anything else I can improve here!

dcci:
Let's test this again and cross finger. If this still fails I'll take a look myself.

dcci:
@swift-ci please test

dcci:
@swift-ci please test

keith:
Looks like the Linux build failed with the same error as before

dcci:
@Keith yes, definitely not related. On a bus now but I'll check in 30 where this problem is.

shahmishal:
@swift-ci test Linux

shahmishal:
Not sure, I have not seen this before. 

```
clang: error: unable to execute command: Killed
clang: error: linker command failed due to signal (use -v to see invocation)
```

dduan:
It certainly _looks_ like OOM. But why would this change cause that....

shahmishal:
@swift-ci test Linux

keith:
It passed 🎉  Thanks @shahmishal !

dcci:
@shahmishal thanks. Can we get the bot back online so I can merge this?

Thanks!

--
Davide

shahmishal:
You should be able to merge now.

dcci:
It was an hw issue, merging this now.

keith:
Awesome! Thanks everyone for your help!

PR #262, opened 2017-12-06
Add tests for SwiftHashedContainer with enum keys

keith:
This tests how hashed containers behave with enums as keys.

This will ideally test the fix in https://github.com/apple/swift-lldb/pull/259, but currently that portion of the test is disabled until that fix is merged.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/hashed_containers_enums/TestSwiftHashedContainerEnum.py
+++ b/packages/Python/lldbsuite/test/lang/swift/hashed_containers_enums/TestSwiftHashedContainerEnum.py
@@ -1,0 +1,97 @@
+# TestSwiftHashedContainerEnum.py
+
+"""
+Test combinations of hashed swift containers with enums as keys/values
+"""
+
+import lldb
+from lldbsuite.test.lldbtest import *
+import lldbsuite.test.decorators as decorators
+import lldbsuite.test.lldbutil as lldbutil
+import os
+import unittest2
+
+class TestSwiftHashedContainerEnum(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
+    @decorators.skipUnlessDarwin
+    @decorators.swiftTest
+    def test_any_object_type(self):
+        """Test combinations of hashed swift containers with enums"""
+        self.build()
+        self.do_test()
+
+    def setUp(self):
+        TestBase.setUp(self)
+        self.main_source = "main.swift"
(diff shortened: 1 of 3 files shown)


keith:
It looks like https://github.com/apple/swift-lldb/pull/259 might pass on the next test run, if that's the case I'll update this PR to run the one test I have commented out

packages/Python/lldbsuite/test/lang/swift/hashed_containers_enums/TestSwiftHashedContainerEnum.py
```diff
@@ -0,0 +1,97 @@
...
+        # Frame #0 should be at our breakpoint.
+        threads = lldbutil.get_threads_stopped_at_breakpoint(
+            process, breakpoint)
+
+        self.assertTrue(len(threads) == 1)
+        self.thread = threads[0]
+        self.frame = self.thread.frames[0]
+        self.assertTrue(self.frame, "Frame 0 is valid.")
+
+        # TODO: This should work after https://github.com/apple/swift-lldb/pull/259
```
kastiglione, line 58: In theory, you could base this branch on that branch, but does swift-ci handle that structure?

keith: Yea, until I saw the green this morning on #259 I wasn't sure when that would be merged. I will rebase once I get an answer on the linux tests over there

modocache: > does swift-ci handle that structure?

Based on [this document](https://github.com/apple/swift/blob/9a5df02e5e0bae3ffff9b1f0590e92d46366ee9f/docs/ContinuousIntegration.md#cross-repository-testing), I think it might, using a comment like this:

```
https://github.com/apple/swift-lldb/pull/259

@swift-ci Please test
```

dcci (approved):
LGTM once the other change goes in.

dcci:
These can be probably merged separately. If I were you, the next time, I'd include the test in the original pull request as it makes archeology easier (and also saves a jenkins run for testing :)

keith:
I'll move these to the other PR, I only did it this way originally because I didn't want having these tests to be blocked if my fix didn't work out.

keith:
I pushed this test, without the TODO, to https://github.com/apple/swift-lldb/pull/259

PR #263, opened 2017-12-08
[swift-support] Catch up with recent changes in upstream.

dcci:
-> GetTemplateArgument() was replaced by GetTemplateArgumentKind() +
       GetTemplateArgumentKind(). In swift we use this for bound/unbound
       generics so let's introduce two variants to replace.
-> TargetSubtargetInfo.h was moved from Target/ to Codegen/ (in LLVM)
-> Archspec.h was moved from Core/ to Utility/ (in LLDB)

rdar://problem/35752437

I don't know who should actually review this, so I attached the suggested reviewers from GitHub.

### DIFF
--- a/include/lldb/Symbol/CompilerType.h
+++ b/include/lldb/Symbol/CompilerType.h
@@ -384,11 +384,10 @@ public:
   size_t GetNumTemplateArguments() const;
 
   lldb::TemplateArgumentKind GetTemplateArgumentKind(size_t idx) const;
-
   CompilerType GetTypeTemplateArgument(size_t idx) const;
 
-  CompilerType GetBoundGenericType(size_t idx) const;
-  CompilerType GetUnboundGenericType(size_t idx) const;
+  lldb::GenericKind GetGenericKind(size_t idx) const;
+  CompilerType GetGenericType(size_t idx) const;
 
   // Returns the value of the template argument and its type.
   llvm::Optional<IntegralTemplateArgument>
--- a/include/lldb/Symbol/CompilerType.h
+++ b/include/lldb/Symbol/CompilerType.h
@@ -384,7 +384,11 @@ public:
   size_t GetNumTemplateArguments() const;
 
   lldb::TemplateArgumentKind GetTemplateArgumentKind(size_t idx) const;
+
   CompilerType GetTypeTemplateArgument(size_t idx) const;
+
+  CompilerType GetBoundGenericType(size_t idx) const;
+  CompilerType GetUnboundGenericType(size_t idx) const;
 
   // Returns the value of the template argument and its type.
   llvm::Optional<IntegralTemplateArgument>
--- a/include/lldb/Symbol/CompilerType.h
+++ b/include/lldb/Symbol/CompilerType.h
@@ -386,8 +386,8 @@ public:
   lldb::TemplateArgumentKind GetTemplateArgumentKind(size_t idx) const;
   CompilerType GetTypeTemplateArgument(size_t idx) const;
 
-  lldb::GenericKind GetGenericKind(size_t idx) const;
-  CompilerType GetGenericType(size_t idx) const;
+  lldb::GenericKind GetGenericArgumentKind(size_t idx) const;
+  CompilerType GetGenericArgumentType(size_t idx) const;
 
   // Returns the value of the template argument and its type.
   llvm::Optional<IntegralTemplateArgument>
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -632,11 +632,10 @@ public:
 
   size_t GetNumTemplateArguments(void *type) override;
 
-  lldb::TemplateArgumentKind GetTemplateArgumentKind(void *type,
-                                                     size_t idx) override;
-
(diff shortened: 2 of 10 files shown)


source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
```diff
@@ -74,6 +74,22 @@ void SwiftUserExpression::DidFinishExecuting() {
   }
 }
 
+static CompilerType getBoundOrUnboundType(CompilerType &valobj_type,
+                                          uint32_t idx) {
+  CompilerType key_type;
+  switch (valobj_type.GetTemplateArgumentKind(idx)) {
+  case lldb::eBoundGenericKindType:
+    key_type = valobj_type.GetBoundGenericType(idx);
```
adrian-prantl: perhaps just `return valobj_type.GetBoundGenericType(idx);`?

adrian-prantl (commented):
I don't understand enough of this to provide meaningful feedback, but from a high level it would be nice to have a few more comments explaining what is going on here in the code.

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
dcci: I definitely agree. I can put some more comments, and in general the whole support needs to be more commented. I'll refine this once somebody who understands the swift support in lldb better than I do signs off the general approach. Thanks!

adrian-prantl:
@jimingham can you take a look at this?

dcci:
FWIW, I really don't yet understand this piece of code so any kind of feedback will be greatly appreciated :)

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
```diff
@@ -74,6 +74,22 @@ void SwiftUserExpression::DidFinishExecuting() {
   }
 }
 
+static CompilerType getBoundOrUnboundType(CompilerType &valobj_type,
```
jimingham: LLDB API's begin with an initial capitol.  So this should be GetBoundOrUnboundType.  But since those are the only types of generic, why not GetGenericType?

jimingham (commented):
I didn't go all the way through because it seems to me if we are going to split off Generics from Templates - which is also a fine idea - we should do it all the way.  So make another enum for Generics types, and don't go through GetTemplateArgumentKind, which seems awkward if we aren't trying to combine Template & Generics behaviors, but just use GetGenericArgumentType and have it's kind return the enum kind (Bound or Unbound.)

Then it would make some sense to move the GetGenericArgumentType to the SwiftASTContext, and implement the template equivalent to always return an invalid kind.

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
dcci: Sounds good.

dcci:
I'd like to get some feedback from somebody from the swift frontend about this as well, if possible.

dcci:
Rebased, now addressing Jim's comments.

dcci:
Addressed Jim's feedback.

include/lldb/Symbol/CompilerType.h
```diff
@@ -386,6 +386,9 @@ class CompilerType {
   lldb::TemplateArgumentKind GetTemplateArgumentKind(size_t idx) const;
   CompilerType GetTypeTemplateArgument(size_t idx) const;
 
+  lldb::GenericKind GetGenericKind(size_t idx) const;
```
jimingham: These need to be called GetGenericArgumentKind & GetGenericArgumentType.  I don't know what it would mean to get the i-th GenericType from a type.  But getting the Argument Type for the i-th argument to a type does make sense.

include/lldb/Symbol/SwiftASTContext.h
```diff
@@ -632,8 +632,10 @@ class SwiftASTContext : public TypeSystem {
 
   size_t GetNumTemplateArguments(void *type) override;
 
-  CompilerType GetTemplateArgument(void *type, size_t idx,
-                                   lldb::TemplateArgumentKind &kind) override;
+  lldb::GenericKind GetGenericKind(void *type, size_t idx);
```
jimingham: Same comment here.  GetGenericKind is an unclear name.

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -289,50 +289,53 @@ SwiftHashedContainerBufferHandler::CreateBufferHandlerForNativeStorageOwner(
...
+    }
     }
 
     CompilerType child_type(native_sp->GetCompilerType());
-    CompilerType element_type(child_type.GetTemplateArgument(1, kind));
-    if (element_type.IsValid() == false || kind != lldb::eTemplateArgumentKindType)
-        return nullptr;
+    CompilerType element_type(child_type.GetGenericType(1));
+    if (element_type.IsValid() == false ||
+        child_type.GetTemplateArgumentKind(1) != lldb::eBoundGenericKindType)
```
jimingham: THis needs to be Generic not Template

jimingham (approved):
Make the three changes and then feel free to push.

PR #264, opened 2017-12-21
Handle Python differences on Fedora

tachoknight:
Fedora's installation of Python gets the result of `distutils.sysconfig.get_python_lib(True,False)` as `lib64`, which causes a failure later on in the installation process as the install script is looking for the `python2.7/site-packages` directory in `lib`. This is not an issue with Ubuntu so the change tests for whether we're running on Linux, and if so and we're on Fedora, run `distutils.sysconfig.get_python_lib(False, False)` to return the same path as the rest of the Python files being built.

### DIFF
--- a/scripts/Python/finishSwigPythonLLDB.py
+++ b/scripts/Python/finishSwigPythonLLDB.py
@@ -38,6 +38,7 @@
 # Python modules:
 import os           # Provide directory and file handling, determine OS information
 import sys          # System specific parameters and functions
+import platform     # Provides information about the system we're working with
 import shutil       # High-level operations on files and collections of files
 import ctypes       # Invoke Windows API for creating symlinks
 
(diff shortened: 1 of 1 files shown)


scripts/Python/finishSwigPythonLLDB.py
```diff
@@ -649,7 +650,13 @@ def get_framework_python_dir_windows(vDictArgs):
             vDictArgs["--cmakeBuildConfiguration"])
 
     if strPythonInstallDir.__len__() != 0:
-        strWkDir = get_python_lib(True, False, strPythonInstallDir)
+        if platform.system() == 'Linux':
```
dcci: Please move this to a separate function.

dcci (changes requested):

dcci:
Please rebase and squash this in a single commit.
I think this hack is alright, but I'll run the tests to make sure this doesn't break other Linux(es).
I'd wait if anybody else has opinions on this, otherwise merge
cc: @fredriss / @adrian-prantl / @jasonmolenda / @jimingham / @JDevlieghere / @vedantk

dcci:
@swift-ci please test linux

tachoknight:
Per dcci's request, I moved the code into a function and put it near the other platform-specific functions.

dcci (approved):

dcci:
Please rebase this as it contains a bunch of other unrelated changes now.
Otherwise, LGTM, and I'll merge it.

dcci:
This looks still really weird (there are 40 commits in this pull request).

tachoknight:
Yes, sorry about that, trying to fix it.

tachoknight:
Fixed with the help of the friendly folks on the #git irc channel (thank you preaction and tkazi)

dcci:
@swift-ci please test

dcci:
@tachoknight I'll merge this if it passes CI. Thanks for your contribution.

dcci:
@swift-ci please test

PR #269, opened 2018-01-05
[ArchSpec] Don't consider Unknown MachO64 as invalid.

dcci:
Even without a proper arch we can access line tables, etc..
Fixes rdar://problem/35778442
This time on the correct branch.


dcci:
@swift-ci please test

PR #274, opened 2018-01-11
[Swift] Update for type reconstruction behavior change with inout parameters

slavapestov:
InOutType comes up here in two cases:

- When a function parameter has InOutType. In this case we want to treat it
  identically to a mutable 'var' local variable, and so it should have
  LValueType.

- Inside the parameter list of a value of function type. Here we want to
  preserve the InOutType.

Previously, type reconstruction only supported the first case by always
returning LValueType. I fixed the second case but broke the first by
always returning InOutType. The correct fix is to check inside LLDB and
special-case an InOutType that appears at the top level only, turning it
into an LValueType.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -4087,6 +4087,13 @@ SwiftASTContext::GetTypeFromMangledTypename(const char *mangled_typename,
                      .getPointer();
 
     if (found_type) {
+      // If we have an inout type at the top level, turn it into an lvalue type.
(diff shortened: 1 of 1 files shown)


adrian-prantl (approved):
See my comments in https://github.com/apple/swift-lldb/pull/275

benlangmuir (approved):

slavapestov:
This got automerged from stable to swift-4.1-branch. It still needs to go into upstream-with-swift and swift-5.0-branch though.

PR #275, opened 2018-01-11
[Swift] Update for type reconstruction behavior change with inout parameters [4.1]

slavapestov:
InOutType comes up here in two cases:

- When a function parameter has InOutType. In this case we want to treat it
  identically to a mutable 'var' local variable, and so it should have
  LValueType.

- Inside the parameter list of a value of function type. Here we want to
  preserve the InOutType.

Previously, type reconstruction only supported the first case by always
returning LValueType. I fixed the second case but broke the first by
always returning InOutType. The correct fix is to check inside LLDB and
special-case an InOutType that appears at the top level only, turning it
into an LValueType.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -4087,6 +4087,13 @@ SwiftASTContext::GetTypeFromMangledTypename(const char *mangled_typename,
                      .getPointer();
 
     if (found_type) {
+      // If we have an inout type at the top level, turn it into an lvalue type.
+      // Function parameters that are inout are treated the same as mutable vars
+      // here.
+      if (auto *inout_type = found_type->getAs<swift::InOutType>()) {
+        found_type = swift::LValueType::get(inout_type->getObjectType());
+      }
+
       CacheDemangledType(mangled_name.GetCString(), found_type);
       CompilerType result_type(ast_ctx, found_type);
       if (log)


slavapestov:
@jimingham I've created two PRs, one for stable branch and one for swift-4.1-branch. Do I need to merge both or is it sufficient to merge just this one?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -4087,6 +4087,13 @@ SwiftASTContext::GetTypeFromMangledTypename(const char *mangled_typename,
                      .getPointer();
 
     if (found_type) {
+      // If we have an inout type at the top level, turn it into an lvalue type.
+      // Function parameters that are inout are treated the same as mutable vars
+      // here.
+      if (auto *inout_type = found_type->getAs<swift::InOutType>()) {
+        found_type = swift::LValueType::get(inout_type->getObjectType());
+      }
```
adrian-prantl, line 4095: Thanks for tackling this! Do you understand better why turning it into an LValue type is the right thing to do here? I'm asking because down in ...

adrian-prantl: line SwiftASTContext::GetNumChildren() we have a really odd workaround for Classes that are LValue types and I wonder if what you want is that workaround.

adrian-prantl: Hmm.. I don't know. I tried cloning that workaround for InOutTypes but that alone does not fix the test (but there many switch(TypeKind)s so I might have missed one. Doing it this way seems harmless and it's documented so let's do it.

adrian-prantl (approved):

benlangmuir (approved):
Not an expert here, but the fix seems reasonable.

source/Symbol/SwiftASTContext.cpp
slavapestov: I don't claim to understand most of what goes on in SwiftASTContext, but it does have the general feeling that there are workarounds that only exist because of workarounds elsewhere. With LValueType, I think lldb should not be using it to represent anything at all. It's an implementation detail of the type checker and we would like to eliminate it from the type system altogether. What's going on here is that lldb is constructing LValueTypes from whole cloth and passing them around, checking for LValueTypes elsewhere within lldb. It should use some other mechanism for this purpose. At least now, TypeReconstruction and mangling never accept or produce LValueTypes, so the "badness" is entirely isolated inside lldb.

slavapestov:
I'm going to merge this because swift-4.1-branch is broken right now (the type reconstruction change was reverted on master but not 4.1).

PR #278, opened 2018-01-12
[Plugins] Catch up with recent changes in the compiler.

dcci:
Same as https://github.com/apple/swift-lldb/commit/e4fc1afa52b9561716ed98e67122ad9d6893d3fc but for upstream-with-swift

### DIFF
--- a/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
+++ b/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
@@ -128,8 +128,8 @@ static std::string TranslateObjCNameToSwiftName(std::string className,
           nominal->loadAllMembers();
         }
 
-        if (func->isAccessor()) {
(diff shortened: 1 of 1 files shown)


rjmccall:
LGTM, thanks.

dcci:
No problem, I went ahead and merge to unbreak the bot.
Currently the branch `upstream-with-swift` is not in a great state but I'm putting some energy to make it fully green.

We discussed ia bit about the future workflow and I think the right thing to do for compiler changes that impact lldb will be:

commit & merge to `upstream-with-swift`
-> backport to `stable`
  -> backport to $X (if needed)

Once things are stable I'll discuss this more carefully on the mailing list.

rjmccall:
Okay.  I just didn't think about the fact that there would be an upstream-with-swift branch, but of course there is.

dcci:
@rjmccall until very recently this didn't even build as far as I can tell, but now that the group has more manpower there's some desire to keep it green as it happens with the `upstream-with-swift` branches in `swift-llvm `and `swift-clang`. I don't even think we have CI for this, yet, I just happened to discover the issue building locally.

PR #282, opened 2018-01-17
[TestExec] Unbreak a test which was failing due to a merge mistake.

dcci:
Another step towards a green `upstream-with-swift`


vedantk:
Oops, we raced on this one. Weirdly, the git history shows two identical commits happening consecutively :/.

PR #285, opened 2018-01-18
[Swift/tests] Enable more tests that now pass.

dcci:
<rdar://problem/35650693>

PR #286, opened 2018-01-18
Revert "[Swift/tests] Enable more tests that now pass."

vedantk:
Reverts apple/swift-lldb#285

https://ci.swift.org/view/swift-master-next/job/oss-lldb-incremental-osx-master-next/1850


dcci:
Thanks. I'm fairly surprised this passes on my machine but not on yours.
I guess the difference is that you're using the `cmake` build or some different setting? 
Anyway, let's have a proper chat later today :)

PR #287, opened 2018-01-22
Pass defines from compile flags to swift clang importer

fjricci:
Since the clang importer will preprocess the module headers,
it will need access to any defines that the user supplied
when building those headers.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1475,6 +1475,10 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
                 swift_ast_sp->AddClangArgumentPair("-iquote",
                                                    cc_options[i + 1].c_str());
               }
+
+              if (!cc_options[i].compare(0, 2, "-D")) {
+                swift_ast_sp->AddClangArgument(cc_options[i].c_str());
+              }
             }
           }
         }


fjricci:
@swift-ci please smoke test

dcci (changes requested):
This needs a test.

jimingham:
I have no objections to this patch in particular.

But this is in the !got_serialized_options branch of handling setting up a swift AST.  In the normal course of things, we shouldn't actually use that path any more.  Having lldb try to understand a complex compile line to pull out the required options was pretty fragile.  The "modern" way to do this is that the compiler serialized the ClangImporter options in the swift module, and then lldb just passes that blob to the compiler.

How were you getting into this branch of SwiftASTContext::CreateInstance?

fjricci:
We only run into this while remote debugging, and it looks like we don't manage to load the AST file in that case. I'll look into why that is, perhaps this isn't needed after all.

jimingham:
That would be good.  The old grub-the-compile-line method needs to get pruned from CreateInstance; you should always have the compile blob around somewhere nowadays.  This whole bit of business needs to get cleaned up, it has history from a bunch of different tries at making Swift debug reconstruction work, and is unnecessarily hard to read.  It will require judicious excision, however.

fjricci:
Sounds good. Will close this for now.

fjricci:
Actually, it appears that all of the AST file logic is macho-specific, and I'm targeting ELF. I think you can still generate swiftmodule files for ELF, but I don't know how lldb would find them, without any sort of debug info support. Maybe this is the best approach after all?

fjricci:
I guess if you're gonna parse the compile flags anyway, you could look for "-emit-module-path" and get it that way...

jimingham:
lldb has to find the swiftmodule somehow, we need that to debug.  

It looks like on ELF they are kept in a section of the binary called .swift_ast.  They end up getting marked with a section type of eSectionTypeSwiftModules when the object files get read in from DWARF.  We either have them in the debug info (we append them to the .dSYM files).  We also look for them in the Source Files in the debug linetable with an extension of .swiftmodule...  

Anyway, you have to have been able to get your hands on the swift module in one form or other or you will only see the public interfaces of all the classes, so you have to have found them somehow.

fjricci:
Looks like lld doesn't support ast insertion currently, but I may be able to pipe the swiftmodule into the swift_ast section using objcopy.

fjricci:
Closing in favor of https://github.com/apple/swift-lldb/pull/288

PR #288, opened 2018-01-23
Allow .swift_ast ELF section to be stored in external debug info

fjricci:
Like the other DWARF sections, the swift ast should be loaded if present only in separate debug-info files.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/split_ast/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/split_ast/Makefile
@@ -1,0 +1,9 @@
+LEVEL = ../../../make
+
+SWIFT_SOURCES := main.swift
+
+SPLIT_DEBUG_SYMBOLS=YES
+
+SPLIT_SWIFT_AST=YES
+
+include $(LEVEL)/Makefile.rules
--- a/packages/Python/lldbsuite/test/lang/swift/split_ast/TestSwiftSplitAST.py
+++ b/packages/Python/lldbsuite/test/lang/swift/split_ast/TestSwiftSplitAST.py
@@ -1,0 +1,61 @@
+# TestSplitAST.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""
+Test that split ast works properly
+"""
+import lldb
+import lldbsuite.test.decorators as decorators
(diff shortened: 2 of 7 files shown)


fjricci:
@swift-ci Please test

dcci (changes requested):
@fjricci can you add a test case?

fredriss:
I'll let Jim and others comment on the patch itself, but any change needs to go to master too. The commit message doesn't explain anything about what this patch brings and there's no test case to explain what it's supposed to fix.

jimingham:
IIUC, when you were using separate debug info files on ELF, we weren't loading the swift_ast section, (which is the way we store the swiftmodule in debug information).  How was debugging swift working in that case?  I would have expected the swift separate debug info variant of the testsuite to have failed most of the swift tests, and then this would fix that variant.  If that's not the case, then I'm still not sure how this was working.

fjricci:
Let me fix my PR for the split-debug-info Swift test (should be a fairly simple fix) and reupload that. I'll look into how things are working there, and see what's different in the case where I need this patch.

fjricci:
Looks like `objcopy --only-keep-debug/--strip-debug` does not prune the `.swift_ast` section, which is why the split-debug-info test (https://github.com/apple/swift-lldb/pull/289) passes. However, when we split out debug info, we also split out the `.swift_ast` section, since (in our opinion, at least), the AST is debugging information and not something we need to keep in the main binary (install on a remote device, etc).

I can modify the split-debug test to also copy over the `.swift_ast` section, and it does fail in that case, but I suppose this comes down to the behavior that lldb wants to support (since the strict --strip-debug in objcopy doesn't require this patch).

fjricci:
For more context, we do something along the lines of:
`cp a.out a.out.debug &&
 objcopy --strip-debug --remove-section .swift_ast a.out a.out`

fjricci:
ping - @jimingham - this is fairly easy to write a test for (pending #307) but I want to make sure this is something we want to support before I go ahead and do that.

dcci:
@fjricci I think the direction is reasonable, but please add a test.

fjricci:
Added a commit which allows a section to be present in either the debug info or the main binary (the existing behavior clobbered binary sections with debug info sections, even if they were empty). Also added testing.

fjricci:
@swift-ci please test

fjricci:
ping

dcci:
@fjricci I'm OK with this going in, but I'd love to hear @jimingham opinion on this as he knows this code better than I do.

fjricci:
ping again

jimingham (commented):
Sorry for the delay.  The approach looks fine to me as well.

I don't have enough experience with ELF to say that adding that size != 0 check where you did won't cause any issues.  Looks correct to me, but the experts on ELF are really on the llvm.org list where most of the Linux support goes on.  So you should really run that change 
by the llvm.org equivalent of PR's.  That process is described here:

https://llvm.org/docs/Phabricator.html

I also don't want to end up maintaining any unnecessary divergences from the llvm.org sources.

PR #290, opened 2018-01-29
[Batch Mode] Pass PrimarySpecificPaths to createEmptyModule and IRGenModule()

davidungar:
Must be merged along with https://github.com/apple/swift/pull/14230 after both PRs have been reviewed and polished.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -35,6 +35,7 @@
 #include "swift/Basic/Dwarf.h"
 #include "swift/Basic/LangOptions.h"
 #include "swift/Basic/Platform.h"
+#include "swift/Basic/PrimarySpecificPaths.h"
 #include "swift/Basic/SourceManager.h"
 #include "swift/ClangImporter/ClangImporter.h"
 #include "swift/ClangImporter/ClangImporterOptions.h"
@@ -4578,10 +4579,14 @@ swift::irgen::IRGenModule &SwiftASTContext::GetIRGenModule() {
       if (sil_module != nullptr) {
         swift::irgen::IRGenerator &ir_generator =
             GetIRGenerator(ir_gen_opts, *sil_module);
+        swift::PrimarySpecificPaths PSPs =
+            GetCompilerInvocation()
+                .getFrontendOptions()
+                .InputsAndOutputs.getPrimarySpecificPathsForAtMostOnePrimary();
         m_ir_gen_module_ap.reset(new swift::irgen::IRGenModule(
             ir_generator, ir_generator.createTargetMachine(), nullptr,
-            GetGlobalLLVMContext(), ir_gen_opts.ModuleName,
-            ir_gen_opts.getSingleOutputFilename()));
+            GetGlobalLLVMContext(), ir_gen_opts.ModuleName, PSPs.OutputFilename,
+            PSPs.MainInputFilenameForDebugInfo));
         llvm::Module *llvm_module = m_ir_gen_module_ap->getModule();
         llvm_module->setDataLayout(data_layout.getStringRepresentation());
         llvm_module->setTargetTriple(triple);
--- a/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
@@ -40,6 +40,7 @@
 #include "lldb/Utility/Stream.h"
 
 #include "llvm-c/Analysis.h"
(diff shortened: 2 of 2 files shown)


source/Symbol/SwiftASTContext.cpp
```diff
@@ -19,25 +19,26 @@
 #include <sstream>
 
 #include "swift/AST/ASTContext.h"
+#include "swift/AST/ASTMangler.h"
 #include "swift/AST/DebuggerClient.h"
 #include "swift/AST/Decl.h"
 #include "swift/AST/DiagnosticEngine.h"
 #include "swift/AST/ExistentialLayout.h"
-#include "swift/AST/IRGenOptions.h"
-#include "swift/AST/ASTMangler.h"
```
dcci: Please submit the header sorting separately.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -4556,8 +4557,8 @@ swift::SILModule *SwiftASTContext::GetSILModule() {
   VALID_OR_RETURN(nullptr);
 
   if (m_sil_module_ap.get() == NULL)
-    m_sil_module_ap = swift::SILModule::createEmptyModule(GetScratchModule(),
-                                                          GetSILOptions());
+    m_sil_module_ap = swift::SILModule::createEmptyModule(
+                                                          GetScratchModule(), GetSILOptions(), swift::PrimarySpecificPaths("", "<lldb>"));
```
dcci: Can you please `clang-format` your changes?

dcci (commented):
Sorry, I must have missed this one. Anyway, I see you didn't merge the compiler bits yet, so here's a quick review

source/Symbol/SwiftASTContext.cpp
davidungar: Thanks for looking at this. I have no need to reorder the headers, it's just what git-clang-format did. So, I'll revise the PR to only add the new header.

source/Symbol/SwiftASTContext.cpp
davidungar: Yup, sorry for the oversight.

davidungar:
Have addressed the concerns and pushed.

dcci:
Thanks David. The patch makes sense to me. 
I followed batch mode (but not particularly closely, I have to admit). I wonder if this can have impacts on debuggability? (I honestly don't expect so, as if this is the only change, it's just an API change), but I would like to confirm with you, so we can plan in advance changes, if needed.

davidungar:
My pleasure, Davide. I appreciate the support in checking my work. 
Impacts on debuggability? I hope not, but would be happy to discuss it. The intention of these changes is to have no impact at all, but there's often a slip twixt the cup and the lip.

dcci:
I really need to double check this (if we use for anything real) but you're swapping the assignment of a member variable to an assignment to a temporary (for `mainInputFileName`).
So, I'll take a closer look tomorrow, but, if we need this to be set, for, e.g., IRGen, then this change has a small bug (if all the tests still passes, well, we should add a test for it).

Otherwise, we might nuke that member variable entirely. You really don't need to do that, I can take care of it, but your change might help us discovering some unused code.

davidungar:
The issue is that I'm trying to move some state out of IRGenOpts: MainInputFile and OutputFIlenames. CompilerInstance owns a CompilerInvocation which owns an IRGenOpts. So the status quo is that MainInputFIle and OutputFilenames are owned by CompilerInstance. But this does not work when there are multiple primaries, since each primary determines the MainInputFile, and logically, the OutputFilename. So, these entities migrate to parameters to the SILModule and IRGenModule. Since there are other such filenames as well, I'm introducing PrimarySpecificFilenames to hold all such.
The debugger changes are intended to preserve its behavior while letting me remove those instance variables from IRGenOpts.

dcci:
I took a closer look at this after your comment and I guess this should be safe. The changes look good.

davidungar:
Please test with the following PR:
https://github.com/apple/swift/pull/14230

@swift-ci Please test

davidungar:
@dcci Davide, the compiler bits are ready to merge as soon as the tests pass. (I'm having trouble with the compiler side not getting the latest changes I've made to this PR.)
There have been some very small changes to this code since you looked at it, just passing two strings instead of a PrimarySpecificPaths structure. Do you want to look it over again? And/or approve the changes? Or do you want me to just merge them as soon as the tests pass, since you did write that the changes look good. Thanks.

davidungar:
I've looked over the changes again, and they are quite simple. I'm going to land them so I can land the big changes in the frontend.

PR #293, opened 2018-01-30
[Playground] Replaced "$builtin" with "__builtin" in the playground preamble.

cwakamo:
This aligns with Swift, which is changing the instrumenter to use "__builtin" instead. (See apple/swift#14257.)

This is part of rdar://problem/36031860.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/playgrounds/PlaygroundsRuntime.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/playgrounds/PlaygroundsRuntime.swift
@@ -53,19 +53,19 @@ class LogRecord {
 }
 
 @_silgen_name ("playground_log_hidden") public func builtin_log_with_id<T>(_ object : T, _ name : String, _ id : Int, _ sl : Int, _ el : Int, _ sc : Int, _ ec: Int) -> AnyObject? {
-  return LogRecord(api:"$builtin_log", object:object, name:name, id:id, range : SourceRange(sl:sl, el:el, sc:sc, ec:ec))
+  return LogRecord(api:"__builtin_log", object:object, name:name, id:id, range : SourceRange(sl:sl, el:el, sc:sc, ec:ec))
 }
 
 @_silgen_name ("playground_log_scope_entry") public func builtin_log_scope_entry(_ sl : Int, _ el : Int, _ sc : Int, _ ec: Int) -> AnyObject? {
-  return LogRecord(api:"$builtin_log_scope_entry", range : SourceRange(sl:sl, el:el, sc:sc, ec:ec))
+  return LogRecord(api:"__builtin_log_scope_entry", range : SourceRange(sl:sl, el:el, sc:sc, ec:ec))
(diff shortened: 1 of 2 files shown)


cwakamo:
Please test with this PR:

apple/swift#14257

@swift-ci Please test

cwakamo:
Please test with this PR:

apple/swift#14257

cwakamo:
Please test with this PR:

apple/swift#14257

@swift-ci Please test

cwakamo:
Please test with this PR:

apple/swift#14257

@swift-ci Please test

dcci:
@cwakamo now that the suite is stable, we should be able to try this again.

dcci:
Please test with this PR:

apple/swift#14257

@swift-ci Please test

cwakamo:
@dcci Thanks for retriggering testing! I was a bit confused how my changes could break the build that way…

@jimingham or @adrian-prantl, would one of you mind taking a look at these changes? Also, I'm targeting this PR at "stable" because it's tied directly to apple/swift#14257, which is targeting master; are there other branches I need to manually cherry-pick this to (other than potentially swift-4.1-branch, as I'm still strongly considering nominating it for Swift 4.1)?

dcci:
@cwakamo the changes lgtm, we currently run only a subset of the tests, so I'd love to make sure the one you modified is enabled. What you did is correct. Once this is merged, you might consider opening another PR against `upstream-with-swift` which will match `master-next` (for swift).

cwakamo:
Resurrecting this PR to try to land it soon; just rebased my changes on top of the latest `stable` branch.

Please test with this PR:

apple/swift#14257

@swift-ci Please test

cwakamo:
@dcci @jimingham @adrian-prantl If one of y'all could take a peek at this PR and review it, I would greatly appreciate it! I rebased it on top of `stable` and would like to land it on Monday so I can potentially nominate it for 4.2 shortly thereafter.

dcci (approved):
This is fine assuming it passes the lldb testsuite.

jimingham (approved):
This seems fine to me.  lldb doesn't really care what these names are.  The $ was nice because then they would never conflict with any real swift names.  But if the swift folks are willing to protect __builtin_* from common usage, then this is okay.

cwakamo:
Thanks for the quick reviews!

@jimingham: `__builtin_*` is used by the [PC macro](https://github.com/apple/swift/blob/master/lib/Sema/PCMacro.cpp), so this change aligns the playground transform with the PC macro in a reasonably-obscure namespace. (It also *only* affects code built with `-playground` or, in the case of the PC macro, `-pc-macro`, which means if there's any impact it should be fairly limited, and it's something we can change in the future if we discover issues or conflicts with this naming scheme.)

@dcci: This passed PR testing for lldb, and it's currently running through PR testing for Swift. Assuming Swift's PR testing passes, I'll plan to land this first thing Monday so there's plenty of time to deal with any fallout in case this testing turned out not to be sufficient.

PR #296, opened 2018-02-01
Implement support for most of the new string ABI

vedantk:
This teaches the string data formatter about some additional bits of the new ABI related to bridged NSStrings. Support for the 32-bit ABI is still WIP.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/variables/bridged_string/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/variables/bridged_string/main.swift
@@ -13,12 +13,12 @@ import Foundation
 
 func main()
 {
-	var s1 = "Hello world"
-	var s2 = "ΞΕΛΛΘ"
-	var s3 = s1 as NSString
-	var s4 = s2 as NSString
-	var s5 = "abc" as NSString
-	var s6 = String(s5)
+	let s1 = "Hello world"
+	let s2 = "ΞΕΛΛΘ"
+	let s3 = s1 as NSString
+	let s4 = s2 as NSString
+	let s5 = "abc" as NSString
+	let s6 = String(s5)
 	print(s1) // Set breakpoint here
 }
 
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -242,8 +242,7 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
     bool is_utf_16 = objectAddr & (1ULL << 60);
     uint64_t payloadAddr = objectAddr & ((1ULL << 56) - 1);
 
-    if (!is_opaque && ((is_a_value && !is_Cocoa_or_small) ||
-                       (!is_a_value && !is_Cocoa_or_small))) {
+    if (!is_opaque && !is_Cocoa_or_small) {
       uint64_t count = otherBits_sp->GetValueAsUnsigned(0) & ((1ULL << 48) - 1);
       if (count == 0) {
         stream.Printf("\"\"");
@@ -269,8 +268,19 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
       else
         return StringPrinter::ReadStringAndDumpToStream<
             StringPrinter::StringElementType::UTF16>(read_options);
-    } else if (is_Cocoa_or_small) {
-      // TODO:
+    } else if (is_Cocoa_or_small && !is_a_value) {
+      CompilerType id_type =
+          process_sp->GetTarget().GetScratchClangASTContext()->GetBasicType(
+              lldb::eBasicTypeObjCID);
+      ValueObjectSP nsstringhere_sp = ValueObject::CreateValueObjectFromAddress(
+          "nsstringhere", payloadAddr, valobj.GetExecutionContextRef(),
+          id_type);
+      if (nsstringhere_sp)
+        return NSStringSummaryProvider(*nsstringhere_sp.get(), stream,
+                                       summary_options);
+    } else if (is_Cocoa_or_small && is_a_value) {
+      return NSStringSummaryProvider(*otherBits_sp.get(), stream,
+                                     summary_options);
     }
   } else {
     // TODO:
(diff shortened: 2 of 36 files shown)


source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -166,89 +166,6 @@ static bool invokeNSStringFormatter(Process &process, ValueObject &valobj,
...
 }
 
-bool lldb_private::formatters::swift::StringCore_SummaryProvider(
-    ValueObject &valobj, Stream &stream, const TypeSummaryOptions &options) {
-  return StringCore_SummaryProvider(
-      valobj, stream, options,
-      StringPrinter::ReadStringAndDumpToStreamOptions());
-}
-
-bool lldb_private::formatters::swift::StringCore_SummaryProvider(
```
dcci, line 176: Whoah, thanks for removing all this cruft.

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -325,15 +242,7 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
     uint64_t payloadAddr = objectAddr & ((1ULL << 56) - 1);
 
-    // FIXME: Remove printf debugging.
-    // stream.Printf("payload address = %llu\n", payloadAddr);
-    // stream.Printf("t (is-a-value) = %d\n", is_a_value);
-    // stream.Printf("v (sub-variant) = %d\n", is_Cocoa_or_small);
-    // stream.Printf("o (opaque) = %d\n", is_opaque);
-    // stream.Printf("w (width) = %d\n", is_utf_16);
-
-    if (!is_opaque && ((is_a_value && !is_Cocoa_or_small) ||
```
dcci, line 335: I didn't understand the whole check to begin with.
I trust you that you can simplify it, but, maybe, can we add a comment?

vedantk:
This fixes several test failures, but we can't land this at the moment because the bots are red.

dcci (commented):
Some comments inline. Not only this fixes several test failures, but also removes some code. Great work.

source/Plugins/Language/Swift/SwiftFormatters.cpp
vedantk: This checks for strings which aren't opaque or bridged. There is a "monster" comment above which documents why this class of strings need specific handling.

vedantk: But it could use a comment like "Handle plain-old native Swift strings"

vedantk:
Yes, the TestSwiftBridgedStringVariables.py I un-xfailed contains examples of each of these kinds of strings, and tests both the `fr v` and `fr v -O` code paths (i.e we test the path through the data formatter and the path through `stringForPrintObject`).

packages/Python/lldbsuite/test/lang/swift/variables/bridged_string/main.swift
```diff
@@ -13,12 +13,12 @@ import Foundation
...
 func main()
 {
-	var s1 = "Hello world"
-	var s2 = "ΞΕΛΛΘ"
-	var s3 = s1 as NSString
-	var s4 = s2 as NSString
-	var s5 = "abc" as NSString
-	var s6 = String(s5)
+	let s1 = "Hello world"
+	let s2 = "ΞΕΛΛΘ"
```
jimingham, line 17: var's and let's are handled differently by the swift compiler.  I presume you changed this to remove the compiler warnings, but in tests like this we should really have a mix of var's and lets and we don't really care about the "you didn't use the var" warnings.  If the warnings bother you, just print all the variables at the end of the test...

Also, now that you know the String format, are each of the variants you had to cover below represented here?

packages/Python/lldbsuite/test/lang/swift/variables/bridged_string/main.swift
```diff
@@ -13,12 +13,12 @@ import Foundation
...
+	let s3 = s1 as NSString
```
vedantk, line 18: @jimingham `s3` is the `is_Cocoa_or_small && !is_a_value` case

packages/Python/lldbsuite/test/lang/swift/variables/bridged_string/main.swift
```diff
@@ -13,12 +13,12 @@ import Foundation
...
+	let s4 = s2 as NSString
+	let s5 = "abc" as NSString
+	let s6 = String(s5)
```
vedantk, line 21: `s6` is the `is_Cocoa_or_small && is_a_value` case

jimingham (approved):
I didn't study the new layout of string, so I can't comment on how accurately this represents the new layout.  But provided the test feeds one of each of the cases you had to cover at the formatters, I don't doubt you got that right.

The code itself looks fine.

packages/Python/lldbsuite/test/lang/swift/variables/bridged_string/main.swift
vedantk: Yes, I changed this to appease the compiler, but I'll try something like '-w' if it bothers me in the future.

All the variants we need to cover are taken care of, but we will need separate handling for the 32-bit ABI.

packages/Python/lldbsuite/test/lang/swift/variables/bridged_string/main.swift
jimingham: It's a little weird that the test case has 6 variables in it but it only tests s3-s6.  s1 & s6 are probably the same, but we only test the NSString version of the non-ascii String (s2)  Does that one work?

vedantk: Thanks for the catch! I didn't bother adding these checks because they are covered in another test, but it makes sense to have them for completeness. I've just pushed the change.

jimingham: Oh, yes, while you are messing with this, lines 45-67 of that test case can be replaced with:
 
    	        (target, process, thread, bkpt) = lldbutil.run_to_source_breakpoint(self,
    	                                   "Set breakpoint here", self.main_source_spec)
 
Be nice to do this if possible whenever we touch a test case.  This is much easier to read and would have meant hundreds fewer test case changes in Adrian's work to move the build out of tree.

vedantk: Done. And thanks, it's good to have this reminder!

vedantk:
Hi @shahmishal, this is the pull-request we discussed. It contains a data formatter fix for the new string ABI and a slew of test-infra fixes. Specifically, it creates the `swiftpr` test category we need for lldb pull-request testing. Could you force-merge this?

vedantk:
Thank you @shahmishal!

PR #300, opened 2018-02-02
Switch hard-coded CountableRange in test to be type Range

airspeedswift:
To match the elimination of the type (except for a compatibility alias) done in apple/swift#13342


airspeedswift:
Not sure if the tests would be better refactored to eliminate some duplication that is only there because there used to be two types.

airspeedswift:
And then at some point, we can remove the other references to `CountableRange` completely... but wanted to fix this test quickly first.

vedantk:
Thanks for the patch! After applying it locally, I am seeing a failure in the test:
'frame variable c' returns expected result, got '(CountableClosedRange<Int>) c = 1...100'

Is this something you can reproduce locally? Either way, I'll rebase and test again.

jimingham:
This looks fine to me.  Since "..." and "..<" are two different user gestures to make ranges it seems worthwhile testing both, even if we know under the covers they are the same.

You mentioned having a "compatibility alias" for CountableRange.  If we ran across a type that wasspecified with that alias, that will trigger the data formatter for CountableRange.  Presumably now that is going to print an incorrect value so we really do need to remove the data formatter as well.  Then the formatter matching will fall back to the type that the alias points to (presumably Range) which will do the right thing.

Depending on how long the CountableRange type alias is going to hang around we might want to add a test for that as well.

But I'm fine with doing that as a follow-up task.

jimingham (approved):
This is fine as a fix to the testsuite.

We do need to remove the stale formatters (both CountableRange & CountableClosedRange, right?) since they might trigger and return the wrong result.

vedantk (changes requested):
I'd like some more time to test this.

airspeedswift:
@vedantk I only landed the switch for `someInt..<someInt` to return a regular not countable range last night, so you may need to pull and rebuild from latest master.

vedantk:
Thanks @airspeedswift, after rebasing the test LGTM! I do have one more change to request. Could you add "@decorators.add_test_categories(["swiftpr"])" below "@decorators.swiftTest"? That way, this test can run as a part of pull-request testing for Swift.

dcci (approved):
LGTM if Vedant is happy/done with his testing.
Thank you @airspeedswift !

vedantk:
The test looks good, but it'd be nice to have this in the `swiftpr` testing category.

airspeedswift:
@vedantk done!

vedantk (approved):

vedantk:
Thank you @airspeedswift! @shahmishal could you force-merge this please?

vedantk:
I’m not sure why my merge dropped Ben’s last change, I’ll just have to add it back in later.

airspeedswift:
I think it's still there, just that GitHub isn't showing it in the Files Changed view?  ¯\\\_(ツ)_/¯

vedantk:
You're right, I must have added in the decorator yesterday.

PR #305, opened 2018-02-05
[SwiftASTContext] Start pruning unused and presumably stale functions.

dcci:
If somebody needs this functionality in the future, it should go
through remote mirrors.

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -748,9 +748,6 @@ public:
   bool IsReferenceType(void *type, CompilerType *pointee_type,
                        bool *is_rvalue) override;
 
-  static bool IsInoutType(const CompilerType &compiler_type,
-                          CompilerType *original_type);
-
   bool
   ShouldTreatScalarValueAsAddress(lldb::opaque_compiler_type_t type) override;
 
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -5221,31 +5221,6 @@ bool SwiftASTContext::IsReferenceType(void *type, CompilerType *pointee_type,
   return false;
 }
 
-bool SwiftASTContext::IsInoutType(const CompilerType &compiler_type,
-                                  CompilerType *original_type) {
-  if (compiler_type.IsValid()) {
-    if (auto ast = llvm::dyn_cast_or_null<SwiftASTContext>(
-            compiler_type.GetTypeSystem())) {
-      swift::CanType swift_can_type(GetCanonicalSwiftType(compiler_type));
-      swift::LValueType *lvalue = swift_can_type->getAs<swift::LValueType>();
-      if (lvalue) {
-        if (original_type)
-          *original_type =
-          CompilerType(ast, lvalue->getObjectType().getPointer());
-        return true;
-      }
-      swift::InOutType *inout = swift_can_type->getAs<swift::InOutType>();
-      if (inout) {
-        if (original_type)
-          *original_type =
-          CompilerType(ast, inout->getObjectType().getPointer());
-        return true;
-      }
-    }
-  }
-  return false;
-}
-
 bool SwiftASTContext::IsFloatingPointType(void *type, uint32_t &count,
                                           bool &is_complex) {
   if (type) {


dcci:

vedantk (approved):

jimingham (approved):
If this isn't being used, it's fine to remove it.  But asking an CompilerType for a function whether one of its arguments is an inout argument seems a perfectly fine thing to do.  I don't see why that should go through remote mirrors.

dcci:
Because it's not the business of the debugger storing this information, instead we should ask the compiler as the only true source of information. History shows it's fairly fragile and can break.

jimingham:
We are asking the compiler.  We say "Compiler, what AST Type does this mangled name correspond to", and it produces a swift::Type that represents it, and we ask that questions.  For static properties of types & functions (like is your third argument inout, or what are the names & static types of a struct's fields) this has not been fragile.  

For dynamic properties (what is the size of X, where are its fields, what is the real type of a generic) for sure we should use remote mirrors.  But anything that can be statically determined it should be fine to get out of the type.  

Another thing that's important is that remote mirrors requires a process.  But if you have read in the debug information and are looking at a function type, you reasonably expect an answer to the question you're trying to ask w/o a process.  In this case you can, it's part of the type signature of the function so we should be able to get to it w/o going through remote mirrors.  We should provide as much information as we can before running the process.

dcci:
If remote mirrors need a process to run, then there should be some other interface to the compiler we can query to get informations *transparently*.

i.e. we might consider entirely skipping the "please give me a type starting from a mangled name so that I can ask questions".
I'd rather think about this in terms of "please given a mangled name and presumably and AST context whether the type corresponding to this mangled name is an inout".

You don't need to squint too hard to realize that the way the `isInOutType()` is written assumes knowledge of too many informations.

i.e.
```
func isInOut() {
getCanonicalType()
if (cast<LValueType>())
  do something
  return
if (cast<InOutType>())
  do somethingElse
  return
}
```

should be:
```
fun isInOut() {
  return CompilerInterface::isThatInOut()
}
```

The second representation has several advantages:
1) It can be entirely tested in the compiler through a dedicated tool
2) It doesn't break the debugger, if, e.g. somebody changes `LValueType`
3) The notion of `inOutType` and `LValueType` don't really leak to the debugger, which, IMHO, is a fairly good thing considering how often swift changes these days.

Unless we really need to do this type inference in the compiler, but in which case, I'd really love to understand why.

fredriss:
One downside of the implement-in-the-compiler option is that this creates another rev-lock between LLDB and the compiler. In the future we'd like to be able to debug code that wasn't generated with the exact same compiler we embed. Maybe with ABI stability implementing those kind of things in the compiler will become a better option not tying us too much to a specific compiler version.

I agree that in general, requiring a process when we can avoid it is bad, but we really need to ask what ourselves what we can do when there's no process. If we won't be able to print a variable anyway, then there's little reason to avoid relying on the debugged process.

jimingham:
I disagree.  We aren't asking "entity, are you inout".  We are asking "entity, are you a thing that has arguments."  Then "entity, if you have arguments, give me something I can ask questions about the first argument."  Then "okay, entity that represents the first argument, are you inout".  So we still need a structured thing that represents composite objects, and the AST types seem perfect for that since it was designed by the Swift compiler engineers for this purpose.

The CompilerType is a thin wrapper on top of swift::Type, so certainly they can be tested in the compiler.  

I don't know why this API was put in (and isn't needed now).  Presumably the question it was asking was whether the value was passed as an Lvalue, and both LValueType and InOutType have the same semantics so from the debugger's point of view they have the same characteristics.  The name maybe was borrowed too directly from the swift version.

jimingham:
We should make a distinction between using the AST from swift to represent types and using the swift module format as the data source for types.  LLDB quite happily uses the Clang AST types for it's type, and that does not introduce any rev lock.  Similarly it isn't so much that we are using the data structures invented by the compiler folks to represent Swift types - which seems to me a good idea - as we are using an unstable storage form as the data source for our types, which I agree wasn't a great idea.

jimingham:
And surprisingly people use the debugger to introspect the types of their apps all the time.  There are some reasons to do this that you won't be able to get from swift (you don't know alignment and sizes without a particular version of the runtime.)  But it is still even before running a fairly convenient programmable interface to types.  I'd like not to have that require a running process.

dcci:
@fredriss I was afraid of the rev locking too, but, as long as your interface is stable'ish (and that should come with ABI stability, as you pointed out), it shouldn't be particularly bad having a shared object you can load which allows you to debug multiple versions of the debugger.

fredriss:
One big difference between Clang and Swift in this regard is that the languages Clang supports are stable. Swift on the other side can introduce new language constructs.

jimingham:
Maybe I'm missing something.  We're going to have to use some structure on our end to represent types.  We need to ask "how many children do you have", "what are their names and types".  We can't keep recalculating this information, that would be slow.   So we'll have be some data structures that really represent at least the information from the type that we care about.  

We could cook up an lldb-specific representation that we populate from DWARF, from RemoteMirrors, from a stable SwiftModule, whatever.   But we won't be able to forward proof that either, right?  If a new construct comes in we'll have to add it to whatever internal representation we have for types.  We aren't really expecting to make an "any possible future swift" data structure for storing swift types are we?

Note also that to support the Expression Parser, we have to have the ability to provide types to the compiler in the way that it uses to reason about them.  At present that's the swift::Type, which is what we use to back our CompilerType for Swift.  So I don't see how we get away from supporting that as A type representation.  If that's true, then having two parallel schemes "lldb swift types" and "swift::Types" seems fragile to me.

So using the representation that the compiler devised to represent types seems to me like it promises the most fidelity, and also reduces the testing surface since it is something heavily tested in the compiler.  If we get to the point where the debugger supports two versions of swift that are so incompatible that they can't be transcoded into the same AST's then we'll have to figure out how to support both versions.  But it seems more likely we will be able to get away with the same "swift 3.1 parsed by the swift 4.0" as the compiler did recently.  And if that's not possible we aren't going to be able to support the expression parser for both variants at all either, and will have to come up with a different solution.

dcci:
The unclear bit to me is "that would be slow". Was this tried in the past and you have evidence that it doesn't scale? If so, can you please point out the history so that I can review it? Thanks!

jimingham:
I haven't actually tried writing a debugger that calculated types on the fly each time it needed them.  It seemed sufficiently obvious that not caching types once you've figured them out would make the debugger too slow that I don't think it ever entered our minds to not cache the types we've found somewhere.  The fact that even when lldb does cache them it's still too slow at handling types is I guess partial evidence to back up this guess.

dcci:
@jimingham the caching doesn't necessarily need to be done in the debugger.
If you have a library that the debugger can ask questions to, well, then you can cache the informations there. This way, the caching mechanism is completely hidden, and whoever wants to ask the same questions just goes through the same cache instead of rolling their own.

Unfortunately (and this is the reason I asked :)) I wasn't really able to find out really bad cases where a profiler (e.g. instruments) shows up we spend all our time doing type resolutions in the debugger [or where we spend a significant amount]. Without that, I'm reluctant to say that lldb is too slow at resolving types even with caching. Do you have an example you can provide so I can play with it?

Thanks!

jimingham:
I'm still confused.  At present lldb DOES use a library to swift cache types behind the covers, namely we reconstruct them into Swift's AST's and ask questions of the AST's.  If we want a nicer API out of the Swift Type interface, we can ask for that, but I don't think that has so much been the problem.  Again, static properties of Swift types hasn't been a big source of problems.

Swift won't show big performance problems with type reconstruction because the swift compiler dumps a mappable form of all the swift types in the swiftmodule, and then can read them directly out of there.  That is one of the benefits of using a predigested form of the entire module as the data store.  

But if you look on the Clang side, where we have to do work to restore types, you can very easily see the debugger churning away working to reconstruct types.  And if we go to anything else but a compiler-state-dump for a data format for types, we will start to see the same problem with swift.

dcci:
OK, I'm glad we agree the performance problem isn't a problem on the swift side (which is what this PR is about). And, FWIW, it's not only about a nicer API, but instead an API where the debugger isn't concerned about the details of how things are implemented more than needed.

I think that if we decide to go to " anything else but a compiler-state-dump for a data format for types", then, this decision can be re-evaluated, but I wouldn't do that before, as the status quo is different.

jimingham:
We do do some caching of mangled-name to AST type (see SwiftASTContext::m_mangled_name_to_type_map).  If you are curious you could try taking that out and see what kind of performance hit you take and what extra work we end up doing.

Most of the code in SwiftASTContext.cpp that deals with translating between lldb's generic type system representation is just to support the generic lldb CompilerType layer.  We do need to have this generic layer, since there's a lot of code in lldb that can deal with swift & clang types interchangeably.  We can't guarantee that there will be a trivial mapping between the generic layer & swift, so we might have to combine two things the debugger doesn't think are different but swift does (e.g. the example you just deleted).  But most of the code in SwiftASTContext that just deals with statically determinable properties of types is a pretty thin wrapper on top of the swift API's. I don't see anything wrong with that.

Of course, where we can push computations onto the Swift side of the API - when it makes sense for the swift API's - we do.  For instance, we did add a few API's to pick out properties of swift functions to support stepping through Swift trampolines in the 4.0 release.  And if you see any of the lldb glue layer in SwiftASTContext doing too much work, feel free to get propose a simpler (from our point of view) API to the swift compiler.

But I don't think this is where we get into trouble with swift.  One immediately relevant place where we got into trouble was trying to infer runtime properties of types from the AST's.  For instance trying to fake out SILGen into computing offset information, or poking at metadata by hand to determine dynamic types.  We don't do much of that anymore, and where we do we certainly should move that code to asking questions of the runtime.  

We also get into huge trouble trying to reconstruct this compiler-state-dump.  That is the weak link for definite sure.  But anything we do to remediate that problem will mean there's more work to be done constructing types.

dcci:

dcci:

dcci:
@swift-ci please test

vedantk:
@dcci shall I go ahead and merge this? I filed https://github.com/apple/swift-lldb/pull/308 for the test failure.

dcci:
@swift-ci please test

vedantk:
@dcci Looks like @shahmishal's change worked 👍

dcci:
yay!

PR #308, opened 2018-02-06
Fix linux fallout from c++ gmodules enable

vedantk:
TestLibcxxListLoop - fails because the evil "define private public"
  trick does not work with gmodules. The purpose of the test is not to
  test debug info parsing so I just mark it as no_debug_info_testcase.
  In the long term it may be interesting to write a mock std::list which
  will allow us to test bad inputs to data formatters more easily.
TestGModules - seems to be a genuine bug. Filed pr36107 and xfailed.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@323520 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 4c11a54e3e22614f7991321995b3d435d7042e48)


vedantk:
@swift-ci test Linux platform

PR #310, opened 2018-02-07
lldb running on an ios device is using the _dyld_get_all_image_infos()

jasonmolenda:
SPI call to to find its own shared cache's UUID.  On newer sytems we
need to use the a new SPI which will return the UUID directly.

<rdar://problem/36625871>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@324437 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 63f24576c2deffa02857dd7d2efdcf21df958a66)

### DIFF
--- a/source/Plugins/ObjectFile/Mach-O/ObjectFileMachO.cpp
+++ b/source/Plugins/ObjectFile/Mach-O/ObjectFileMachO.cpp
@@ -2179,6 +2179,11 @@ UUID ObjectFileMachO::GetSharedCacheUUID(FileSpec dyld_shared_cache,
            sizeof(uuid_t));
     dsc_uuid.SetBytes(uuid_bytes);
   }
+  Log *log(lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_SYMBOLS));
+  if (log && dsc_uuid.IsValid()) {
+    log->Printf("Shared cache %s has UUID %s", dyld_shared_cache.GetPath().c_str(), 
+                dsc_uuid.GetAsString().c_str());
+  }
   return dsc_uuid;
 }
 
@@ -5942,6 +5947,9 @@ UUID ObjectFileMachO::GetProcessSharedCacheUUID(Process *process) {
     dl->GetSharedCacheInformation(load_address, uuid, using_shared_cache,
                                   private_shared_cache);
   }
+  Log *log(lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_SYMBOLS | LIBLLDB_LOG_PROCESS));
+  if (log)
+    log->Printf("inferior process shared cache has a UUID of %s", uuid.GetAsString().c_str());
   return uuid;
 }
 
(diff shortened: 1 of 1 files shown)


jasonmolenda:
@swift-ci please test

dcci:
Hmm, this has no test. I don't feel comfortable merging this as it's changing a fairly subtle detail and chances are this will break in the future.

dcci:
Reading the commit, this API should exist on 10.12 and beyond, which is, among others, the version we currently use. It should be possible to test this change on High Sierra, e.g.

jasonmolenda:
This function is only built when lldb is built on iOS - it's guarded by #ifdef's.  The only way to test it would be to intercept a call to dlsym() to interpose our own copy of their SPI function, when built on iOS.

dcci:
Then we should consider adding a unittest for `GetSharedCacheUUID` and run that on configuration where this is supported. We seriously can't keep leaking changes untested.

jasonmolenda:
Our testing for iOS is concentrated on our customer use scenario -- lldb running on the mac, debugserver and the inferior process running on the device.  Looking into testing lldb on the device is something we should do in the future, but it is a much lower priority workflow.

dcci:
We really should have regression testing for anything that changes functionality. Tomorrow somebody can show up, remove this code, and we're going to ship with a broken functionality in the wild. I apologize if this sounds pedantic, but not testing this feature right now is just pushing the problem away a little bit, as any other similar feature will hit the same "we can't test it". In other words, I'd rather try to solve this problem now than postpone it.

jasonmolenda:
It's a good discussion to have.  Testing new areas of lldb, new corners and environments that we have not been able to test before, is for the good of the project.  But that's not where we are today.  And while this does impose the requirement that people work with some care, and there is the chance of regressions, lldb has worked pretty well for the past decade.  We can do better, we will do better, but lldb was designed with one approach to testing in mind, and it will take a lot of time and design consideration to build on that.

vedantk:
I think we agree on our shared goals here. But I think testing should be one of our highest, if not the highest priority right now. As I understand it we have a few weeks to land this change. I think we should take the time to figure out how to test this now, to put ourselves in a better position going forward.

vedantk (changes requested):
Let's set up some way to test this.

jasonmolenda:
I think having a discussion about how to test the on-device lldb is a good discussion to have.  I don't think changing the requirements for lldb's testing to say that a patch needs to have infrastructure in place that does not exist is appropriate.

vedantk:
I agree that it's a good time to hash out on-device testing for lldb. So long as landing this isn't an emergency, I'd ask that we flesh that out and come up with a test plan before checking in more uncovered code. I realize this is a new requirement, but still think that it's appropriate, as most of the team has been working on lldb test infrastructure and the work is paying off.

fredriss:
Let's have the discussion about what it would take to test this then. We can then make an informed decision. As I said earlier, I'd really like every functional change to be covered by a test case.

jasonmolenda:
On-device lldb testing is far, far down the list of testing priorities that we already have.  There's no way I can reasonably allocate any time working on testing on-device lldb ahead of the tasks we already have planned.

vedantk:
Per our offline conversation, we'll work on getting build/test infrastructure for this going in parallel. Thanks Jason!

vedantk (approved):

dcci:
Closing this. Feel free to reopen if you want to merge it.

PR #313, opened 2018-02-07
[SwiftASTContext] Remove now always failing caching of common demangl…

dcci:
…ing.

This never kicks in because the mangling prefix changed from `_T` to
`$S`. Also, it's not the debugger job to decide which names are more common,
and it's unsure whether this decision should be static.

I plan to remove this cache altogether, my first set of measurements
show this is not a time sink, and this mechanism exhibits some fragility.

I think a better way of handling this situation is either:
a) Speed up the demangler if it turns out to be slow
b) Have a mechanism in the demangler library that transparently
caches results (and implements some suitable cache evition policy).
This way, the debugger can dynamically communicate what are the set
of symbols more commonly demangled, and the library can decide to
cache them accordingly. This avoids the hardcoding problem and
makes the whole thing more robust.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -3062,21 +3062,6 @@ swift::ASTContext *SwiftASTContext::GetASTContext() {
     }
 
     GetASTMap().Insert(m_ast_context_ap.get(), this);
-
-    // Store common useful manglings for quick lookup - this also ensures that
-    // types that didn't come out of the visitor (e.g. fallback ObjCPointers)
-    // still exist in our tables for later mangled name retrieval
-    // (the expression parser needs to do this).
-    CacheDemangledType(ConstString(SwiftLanguageRuntime::GetCurrentMangledName("_TtBO").c_str()).GetCString(),
-                       m_ast_context_ap->TheUnknownObjectType.getPointer());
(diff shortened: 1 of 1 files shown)


dcci:
@swift-ci please test

vedantk (approved):
We'll have to measure, but there's reason to be optimistic about the perf impact of removing this cache. @eeckstein has done a lot of great work on the demangler over the past year.

dcci:
Yes, I think it's much easier working with Erik if there's something slow than having this cache in.

jimingham:
I'm a little worried about the comment 

"this also ensures that types that didn't come out of the visitor (e.g. fallback ObjCPointers)"

which makes it sound like this was more than just a performance issue.

It isn't (didn't used to be?) 100% true that we never see the _T mangled names, since they exist in some system frameworks even when the main swift compiler doesn't have them, so lldb gets presented with them even though the current swift compiler doesn't emit them.

We should check with Erik to make sure this is no longer a problem.

dcci:
The tests pass, which either means
1) we need better coverage
2) this is not required anymore.

Either way, we'll discover, and this time we'll add a test. In general, I think it's a fundamental brokenness having the expression parser relying on things being cached, so, I'd rather have this one broken for a little bit & find a way to fix it, than have this big hack in place for longer.

PR #316, opened 2018-02-08
[SwiftLanguageRuntime] Fail when we can't retrieve the type.

dcci:
This would've made diagnosing other bugs much easier because
we fail immediately instead of going on and producing an
invalid expression. It may seem very brusque to just fail here,
but I think this is good to shake out issues instead of hiding
them.

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1437,21 +1437,12 @@ SwiftLanguageRuntime::MetadataPromise::FulfillTypePromise(Status *error) {
       m_remote_ast->getTypeForRemoteTypeMetadata(
           swift::remote::RemoteAddress(m_metadata_location));
 
-  if (result) {
-    m_compiler_type = CompilerType(m_swift_ast, result.getValue().getPointer());
-    if (log)
-      log->Printf("[MetadataPromise] result is type %s",
-                  m_compiler_type->GetTypeName().AsCString());
-    return m_compiler_type.getValue();
-  } else {
-    const auto &failure = result.getFailure();
-    if (error)
-      error->SetErrorStringWithFormat("error in resolving type: %s",
-                                      failure.render().c_str());
-    if (log)
-      log->Printf("[MetadataPromise] failure: %s", failure.render().c_str());
-    return (m_compiler_type = CompilerType()).getValue();
-  }
+  assert(result && "RemoteAST answer is invalid!");
+  m_compiler_type = CompilerType(m_swift_ast, result.getValue().getPointer());
+  if (log)
+    log->Printf("[MetadataPromise] result is type %s",
+                m_compiler_type->GetTypeName().AsCString());
+  return m_compiler_type.getValue();
 }
 
 llvm::Optional<swift::MetadataKind>


dcci:
@swift-ci please test

JDevlieghere (approved):
Seems in line with our goals. No objection from my side.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1437,21 +1437,12 @@ SwiftLanguageRuntime::MetadataPromise::FulfillTypePromise(Status *error) {
...
-  } else {
-    const auto &failure = result.getFailure();
-    if (error)
-      error->SetErrorStringWithFormat("error in resolving type: %s",
-                                      failure.render().c_str());
-    if (log)
-      log->Printf("[MetadataPromise] failure: %s", failure.render().c_str());
-    return (m_compiler_type = CompilerType()).getValue();
-  }
+  assert(result && "RemoteAST answer is invalid!");
```
vedantk, line 1440: Do you want user stack traces when this fails in release mode? If so, we should use lldbassert() here.

Otherwise this lgtm, as I don't see much benefit from logging the failure cause, and the code removal is good.

dcci:
From what I see `lldbassert` isn't really what we want here.
My understanding is that 
1) lldbassert logs (but doesn't assert) in release
2) asserts in debug.

I'd really love to be stricter. For the bug of last week lldbassert wouldn't have helped as enabling the log causes the bug to disappear. In other words, assert is a much more reliable mechanism here.

vedantk:
Ah, right, we want something that asserts in release mode also. Yep, this lgtm.

source/Target/SwiftLanguageRuntime.cpp
jimingham: So this is definitely not a fatal error, and is one that we need to handle so an assert is not appropriate.  For instance if swift ever were to lazily initialize some generic variable, it's quite likely that when we try to print it it's type is currently unknown.  At that point returning an empty CompilerType is entirely appropriate.

jimingham (changes requested):
How do you know that there are never cases we might run into (like not-yet initialized lazily initialized variables) where RemoteAST will fail to figure out the type?

If you don't know that then turning a handleable error - which is signaled by returning an empty compiler type - is going to assert unnecessarily.

jimingham:
We try to make a distinction between "errors on user data" and "errors on validated computed data".  It is totally appropriate to assert on the latter.  We try never to assert on the former, but instead find a way to back out and not crash.  This is DEFINITELY in the "errors on user data" category, and we have a well-established way to handle ValueObjects whose types we can't determine, we signal that by returning an invalid type, and since we always know that reconstructing the types for values is an iffy thing, we alway check that CompilerTypes are valid before using them.

This is not a change we should make.

dcci:
The purpose of this is finding cases where it fails, and fix them accordingly (or decide whether we want to make them errors). 

In the context of an expression being evaluated, if you have
`Patatino<T>` where `T = String` and you can't resolve it, well, you'd better stop really soon because chances are you're going to construct an invalid expression and fail with a really cryptic and spectacular error. If there are cases where it's actually legit that the type can't be resolved, then the API between compiler and debugger should be enhanced to discriminate between the two.

source/Target/SwiftLanguageRuntime.cpp
vedantk: Presently it appears we don't have a test for this case, and the point of this patch is to make it easier to find these cases. Once we do, we can return an invalid CompilerType here with more confidence, or perhaps change the API to return an `Optional<CompilerType>` to eliminate issues which arise when using a type without checking IsValid().

vedantk:
The general idea here is to let us operate on CompilerTypes reliably (it should never be an 'iffy' thing), and to let us kill things like CompilerType::IsValid() (which we can do by asserting on bad data -- user or otherwise -- and passing around `Optional<CompilerType>`).

So, fwiw, this is a change I'd be happy to have :).

vedantk:
@jimingham To elaborate on why I think operating on CompilerType should never be iffy, consider how types are handled in llvm/clang/swift etc.

llvm::Type can never be invalid, clang::QualType can never be invalid, swift::Type can never be invalid, etc. This is great because you can do things with types with confidence, without lots of menial double-checking of "Ty.isValid()", which you'll invariably forget to do.

Asserting early on invalid inputs and passing around only valid CompilerTypes or `Optional<CompilerType>` will get us to non-iffy type handling.

jimingham:
It is never safe to expect that you can reconstruct the type for something.  The DWARF could be screwed up.  The variable could not be initialized yet and we are relying on runtime properties to figure it out.  We also can sometimes figure out the name of a variable (from the ObjC runtime) even though we can't figure out its type.  That's still a somewhat useful object.  

So the iffiness is in the behavior of the world, not in the behavior of lldb.  We just can't 100% guarantee that we can figure out the type of variables.  In lldb the clear rule is it is NOT a fatal error not to be able to reconstruct the type, and if you can't figure it out then you return an empty compiler type and if you are going to use the type of something you need to check if it is valid.

jimingham:
lldb types are totally different things from clang types.  A clang type is figured out from source and if you couldn't figure it out you would have had an earlier error.  OTOH, for instance, the default clang behavior is to only emit class definitions in the debug information for the module that has the vtable for the class.  So if I give you headers and a dylib that contain a class you are supposed to subclass, but DON'T give you debug information for the dylib, then lldb won't be able to construct the type for the subclass.  That's expected behavior and we have to deal with it.  Types are computed data not user data for clang.  Types are user data (from DWARF, etc) for lldb.  So it is appropriate for clang to treat types as things that always have to be valid.  It is NOT appropriate for lldb to treat them that way.

vedantk:
It's fine to not be able to reconstruct a type, but a CompilerType should either be valid or not exist.

If there's other useful information about a Value, like it's name, it should be moved out of the CompilerType if it isn't already.

For this particular patch: there's no test coverage for invalid CompilerType generation in this method. Until we have that test coverage (i.e, until we're really sure it happens), I don't see why we'd bother handling invalid types here. Having an assert will help us determine whether or not that happens.

vedantk:
(By "really sure", I mean, we have a solid understanding of it, as well as a reliable test :)

jimingham:
I disagree.  I don't think you will ever be able to write test cases for all the ways user data in the compiler could fail.  For instance, a newly introduced bug in clang that causes horked up data in a way you could never have thought of.  Such a thing might even get out into the field and then we would be crashing the debugger even though the users might not even care about that type.  Would you feel okay telling users "It was okay that I crash on you every time you load your binary built with a new clang: I couldn't have known to write a test case for that so I it wasn't my responsibility to handle it gracefully."

Seems to me the way you build software that actually gets up to the point where it is useful enough that you can start using it is you have to first of all have clear policies for handling errors and implement the policy everywhere.  In this case the policy is clear: "The inability to construct a CompilerType is NEVER a fatal error, and you signal that by returning an empty CompilerType."  It's not a clear policy if it is only followed when you can construct a positive test case that traverses that error.  You apply the policy everywhere, then go back and fill in with as many tests as you can manage to construct.

jimingham:
We use lldbasserts because then if the compiler (e.g.) gets a new bug, and we're lucky and our test suite triggers the bug, then the test will crash.  So we catch it early.  But if the bug in our user data (DWARF, RemoteAST, whatever) gets out into the field, then we DON'T want to take down a user's debug session because of a bug in other parts of the system.   And especially if you can easily see how to recover - have an invalid compiler type, and say "couldn't figure out type"...

But given that RemoteAST doesn't claim to always be able to figure out the type (that's why it returns a checkable entity) we should not claim a stronger promise by asserting here than RemoteAST was willing to assert.

vedantk:
We can write the tests we need if we try, and that means we need to make failure cases easy to discover, i.e using assertions.

This is what clang/swift do: e.g yes, the parser may have a bug and accept invalid code, but the response isn't to make all type handling iffy, because that would pose a high maintenance burden. Instead, the response is to test the parser thoroughly. lldb isn't different.

And, of course I don't want the debugger to crash on users :). I just think the way to achieve that is by being a bit stricter about what we consider valid internal data structures, and to ramp up testing.

I take your point about needing clear policies to build useful software, but think the policy needs a change :). To restate it: CompilerTypes should either be valid or not exist, and we should enforce that strictly (e.g with test coverage). That's where this patch is going.

vedantk:
@jimingham I'm worried we're talking a bit past each other at too high a level.

Perhaps we can look at removing the IsValid() method from CompilerType and propagating `Optional<CompilerType>` instead -- I'm hoping we can agree that it should be impossible to accidentally use an invalid CompilerType.

If you're not convinced by this patch, we can revisit the issue of test coverage / assertions later, perhaps?

jimingham:
Again, you give an example of clang which is not instructive.  If clang can't figure out what a type is, something has gone wrong, the compile is going to fail and it's fine to get out of there.  lldb OTOH has to cope with the fact that it can't always construct types (see the example above, but there are many others).  You can't treat a CompilerType as a non-failable thing in lldb, it isn't.

I don't want to argue that point if you don't mind, it's been true in so many ways both in gdb & lldb, you just have to take my word for it.

vedantk:
To be clear, I agree with you that lldb has to deal with not being able to reconstruct types, and realize that it's different from clang in that it needs to be resilient to more types of errors. I'm not disagreeing on that point at all.

I do think adding asserts and test coverage is the way to discover failure modes & avoid operating on invalid state, but am fine with discussing that later.

I was hoping to at least reach a decision on CompilerType::IsValid() (or is it ::IsInvalid()?). Are you OK with the following plan:

1) Make the CompilerType() constructors private
2) Add `Optional<CompilerType> CompilerType::create(...)` as a static method
3) Kill CompilerType::IsValid()

?

That would make it impossible to accidentally use an invalid CompilerType, which is what we really want here.

jimingham:
I am actively opposed to any patch that asserts that we can always get a compiler type from out in the world - and going through RemoteAST is definitely going out into the world.

As for whether you use a compiler enforced policy or a social policy, that I have no strong opinion on.  The dangers of not enforcing these policies have been mostly theoretical so we haven't set aside the time to go plumb through some forced-check access policy.

But if there's some compile-time way to force using code to check the validity of the object before touching it, and it doesn't make the contents of the code get obscured by the policy enforcement, and you have the time to go do this, then I have no objection to that.

vedantk:
Ok then, that at least gives us a path forward. Having done similar refactoring in llvm before, I'm sure we'll catch and mitigate accidental misuse of CompilerType by wrapping it in Optional (which also  tends to be more legible, in my experience).

And I am fine with discussing having asserts+coverage later :).

jimingham:
Sure, go for it.  If you're interested in the history of the "asserts in lldb" discussion it's come up a bunch of times previously on the lldb list.

JDevlieghere:
FWIW, I'm very much in favor of the Optional approach. We are pushing for the same thing in LLVM's `libDebugInfo` where we also encounter all kinds of potentially invalid user input that we need to handle gracefully. 

> Asserting early on invalid inputs and passing around only valid CompilerTypes or Optional<CompilerType> will get us to non-iffy type handling.

What exactly would we still need to assert once we have CompilerType wrapped in an Optional?

vedantk:
> What exactly would we still need to assert once we have CompilerType wrapped in an Optional?

Possibly asserting invariants on the CompilerType itself. It's hard to say, it's a bit too high level at the moment.

jimingham:
One last comment about testing  vrs. production of entities from "user data'.  The view we took with lldb was that unless you can prove that an operation can not fail, you should always assume it can fail and plan to handle that.  Even if you can't actually figure out how to generate the failure, you must assume it can fail and handle it.  After all, debuggers deal in imperfect data and SPI's and they lie and fail in unexpected ways all the time.

Then as time permits, you should go through and try to produce failures in tests, to make sure you handle them correctly.

You must not think, "that might fail, but I can't figure out how to trigger the failure, so I don't need to handle a failure here."  Then you will crash on some user somewhere and they won't have any recourse.

vedantk:
It's the planning to handle failure cases bit that's biting us (instances of which @dcci has heroically dug into, and which this patch is a response to). Sometimes the failure handling causes problems worse than the initial issue (UB heisenbugs, printing wrong values, etc :). As a user I'd say a crash is preferable to this -- but certainly not the only alternative.

Techniques like wrapping possibly-invalid objects in Optional will help.

My experience has shown that assertions are a valid tool in uncovering and mitigating bugs due to bad external inputs. But we're not making progress on that discussion now, we can revisit it later.

jimingham:
This particular case, however is one we should expect might fail sometimes.  After all, the API returns a fallible object...

It's perfectly reasonable to assert in the test suite, saying "all the operations we know of in the test suite should not fail here".  That's what lldbassert is for.  

But it is not appropriate to assert on users - or to just do something undefined like accessing an object that the provider said was not good and then asserting if that didn't manage to crash before you got to the assert.

In this case, the error is going to be very clear.  There will be some variable somewhere that we won't be able to print because we can't recover it's type.  So you really aren't saving yourself any work figuring out the error here.  At best you might force users to report the fact that we didn't get a value for their variable because instead of just not getting the value for that one variable and leaving the debug session going along fine, we will crash.  That is NOT a win.

vedantk:
The assert is fine until there's test coverage for the failure case. This goes back to our general disagreement on what the error-handling policy should be, so we're rehashing things now. Let's stop while we're ahead, as we've found identified work on CompilerType that we're in agreement on :).

dcci:
> In this case, the error is going to be very clear. There will be some variable somewhere that we won't be able to print because we can't recover it's type. So you really aren't saving yourself any work figuring out the error here.

I'm afraid this is not the case. The bug in this case manifested in two different instances:
`frame var Patatino`
prints 
`MyClass<T> Patatino = ...`
instead of
`MyClass<String> Patatino = ...`

so, no error, just a wrong type silently emitted.

If you instead try
`print Patatino`
you get an error while parsing the expression cryptic enough.

On top of that:
1) This only reproduces in release builds
2) If you enable the log, the bug goes away.
3) The assertion catches the bug right away.

So, I wouldn't necessarily claim "the error is very clear" and the assertion doesn't buy you anything.

jimingham:
I think in the frame var case the contents of your "..." was significant.  

When we thought the variable was of a generic type because we couldn't reconstruct it's real type we also didn't print a value for it.  Unable to print any value for a variable is a pretty clear error.  If you were debugging and cared about that variable you would immediately notice (e.g. in the Xcode locals window) that the variable had no value and report it to us.  If you didn't care about the value of that variable you would be very happy we didn't crash on you.

IIRC, we also had a test case whose failure led you down this path.  So if you had an lldbassert in the place you wanted to bail altogether, it WOULD have asserted in the testsuite and done exactly what you wanted.

If the error message in the expression parser when we can't reconstruct the type of a variable that's part of the expression is unclear, we should address that directly, since that's a pretty straight-forward error.

dcci:
We had an offline conversation about a different way of handling this.
I'll probably revisit the approach later in the week. Closing this for now, and thanks to everybody involved!

PR #317, opened 2018-02-09
[build-swift-cmake] Genericizes build script for use on Windows

AndrewSB:
This change depends on swift/utils/build-script also being updated to
run on Windows.

### DIFF
--- a/scripts/build-swift-cmake.py
+++ b/scripts/build-swift-cmake.py
@@ -70,7 +70,7 @@ def use_gold_linker():
     """@return True if the gold linker should be used; False otherwise."""
     return os.path.isfile("/usr/bin/ld.gold")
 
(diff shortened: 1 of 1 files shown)


dcci:
I think this is a good one _regardless_ of Windows support as it improves compatibility.
I'm going to spawn a bot run to make sure this doesn't break our bots.
Also, if you can try your changes on Linux and MacOS, that would be great.

dcci:
@swift-ci please test

AndrewSB:
🙃 I thought the same. 

This worked for me on linux and it looks like Jenkins is happy too. Should be good to merge

dcci (approved):
LGTM

keith:
This appears to have broken the build, fix here: https://github.com/apple/swift-lldb/pull/341

PR #318, opened 2018-02-09
[test] Disable TestRepl{Dictionary,NSString} in PR testing

vedantk:
They appear to be flaky.

rdar://37407231


vedantk:
@swift-ci test

jimingham:
pexpect seems to be flakey and beyond our control.  All the REPL tests that were written using lldbsuite.test.lldbrepl should be converted to use lldbsuite.test.lldbinrepl.  That uses the expression parser directly, and is much more stable.  Those tests are actually much easier to read as well.

I'm fine with not using these tests for swiftpr, but we should remember to get back to them and convert them to the lldbinrepl form instead.

vedantk:
Gotcha, I've got a radar to do just that.

jimingham:
Excellent!

dcci:
After a fair amount of yak shaving, #340 and #343 fix this.

PR #321, opened 2018-02-12
Cleanup unexpected success for C & C++

JDevlieghere:
I'm going through the unexpected successes in the test suite. I'm re-enalbing the ones for which the radar suggests they might/are fixed on the condition that they works on all of my machines. 

Actually, all the tests in this PR were already re-enabled upstream, so I'm not expecting any issues at all.


JDevlieghere:
@swift-ci please test

dcci:
Yes, this is all good. The only thing I'm asking you Is to sync up with Adrian as you might want to consider back porting his changes to the testsuite to build each test in its own subdirectory (or, well, at least make sure your changes still work after [they should,but double checking won't hurt])

JDevlieghere:
Sounds good!

dcci (approved):

PR #322, opened 2018-02-12
[test] Remove inline tests from the `swiftpr` category

vedantk:
The logic to apply decorators to these tests does not quite work
correctly. More tests than intended end up in the `swiftpr` category
when a decorator is added to a single inline test.

Related to r://37173570.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/foundation_value_types/date/TestSwiftFoundationTypeDate.py
+++ b/packages/Python/lldbsuite/test/lang/swift/foundation_value_types/date/TestSwiftFoundationTypeDate.py
@@ -13,6 +13,4 @@ import lldbsuite.test.lldbinline as lldbinline
 import lldbsuite.test.decorators as decorators
 
 lldbinline.MakeInlineTest(
-    __file__, globals(), decorators=[
-        decorators.add_test_categories(["swiftpr"]),
-        decorators.skipUnlessDarwin])
+    __file__, globals(), decorators=[decorators.skipUnlessDarwin])
--- a/packages/Python/lldbsuite/test/lang/swift/foundation_value_types/indexpath/TestSwiftFoundationTypeIndexPath.py
+++ b/packages/Python/lldbsuite/test/lang/swift/foundation_value_types/indexpath/TestSwiftFoundationTypeIndexPath.py
@@ -13,6 +13,4 @@ import lldbsuite.test.lldbinline as lldbinline
 import lldbsuite.test.decorators as decorators
 
 lldbinline.MakeInlineTest(
-    __file__, globals(), decorators=[
-        decorators.add_test_categories(["swiftpr"]),
-        decorators.skipUnlessDarwin])
+    __file__, globals(), decorators=[decorators.skipUnlessDarwin])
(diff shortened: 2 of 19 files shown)


vedantk:
@swift-ci test

dcci:
I'm torn about this one. 
We have already relatively little coverage and this would just decrease it. What's the immediate problem we have with the current solution [there are some tests which are failing] ?

vedantk:
This is a temporary measure as I have no clean solution in mind.

Apart from TestSwiftReturns.py, which is failing on my machine (I believe you ran into this too?), there is no immediate problem. But it's a problem waiting to happen as the bots are running tests not marked `swiftpr`. IMO it's worth having reliable control over which tests run during PR testing -- the "full" lldb testing jobs can be responsible for running everything.

Some alternatives;
- Add and support a categories kwarg to MakeInlineTest
- Have the decorator maintain a list of tests which should run

dcci:
I think having the decorator maintaining a list is probably a better approach than this one, but, I'd love to hear your opinions as well (or Adrian/Jonas as they worked on the testsuite relatively extensively recently).

vedantk:
IMO it'd be a bit of a leaky interface if the decorator maintained a list of tests (seems like something the test runner should be in exclusive control of?). I think @JDevlieghere had a clearer idea of how this might work.

dcci:
I don't disagree with you, but I'm really afraid about losing even more coverage. Any chance we can postpone a decision until Jonas comments? Thanks!

vedantk:
Sounds good

JDevlieghere:
I'm having a quick look at some alternatives. I'll report back here as soon as I know more!

JDevlieghere:
Alright, so here's what I found out. The way we assign categories is by setting an attribute on the function or method. Before calling the test, we check what categories are set. The problem is that for `inlineTest`s we just have different instances of a class with the same methods (e.g. we always have `_InlineTest__test_with_dwarf`, `_InlineTest__test_with_gmodules`, etc). Once the category is applied, it applies to every instance of the `InlineTest`, which are effectively different test. This is obviously not what we want... 

So I wondered why the other decorators did work with the `InlineTest`s. From looking at the source code for `skipIf` in `unittest2` I found out that they wrap the test in another function. The checks are performed in the wrapper, when the function/method is called. This is different from how categories work, which are checked *before* the function is called. 

I think the proper solution is to follow the same approach for `add_test_categories`. Maybe it's just a matter of turning this into an implementation of `skipIf` that skips the test if the current categories are not defined. Alternatively we'd have to roll our own implementation that performs the necessary checks in the wrapper.

vedantk:
Nice! It seems that there's some logic in decorators.py to handle custom skip reasons. I'll try it out.

dcci:
Good, thanks for looking at this, Jonas!

dcci:
@JDevlieghere did you get a chance to look at this? Thanks!

vedantk:
Yes, Jonas and I have been iterating on this offline. There's a prototype for using a skipIf-style decorator to mark tests for swift-pr testing, but we're not sure whether it's running the correct set of tests yet.

dcci:
@vedantk / @JDevlieghere Can this one be closed now that we are iterating on #342 ?

PR #325, opened 2018-02-13
Fall back to compile flags when serialized debugging options are missing

fjricci:
lldb already has access to the compilation flags from the DWARF,
we should make a best effort to fall back to those flags if the module
was built without serialized debugging options enabled.


fjricci:
@swift-ci Please test

fjricci:
Hmmm, I'm trying to base this test on TestResilience, but that one is disabled on swiftpr, and fails when I run it locally.

dcci:
Adrian is working to re-enable that test.

fjricci:
Awesome, I'll watch out for that.

fjricci:
WIP while I migrate the test to lit.

fjricci:
@swift-ci please test

fjricci:
@swift-ci please test

fjricci:
@swift-ci please test

fjricci:
Verified that this passes `ninja check-lldb-lit` locally on macOS, since the CI bots don't run the lit suite afaik.

fjricci:
@swift-ci please test

fjricci:
@swift-ci please test

jrose-apple:
Please don't do this. We've been trying to kill the flags out of DWARF for years now and there keeps being one more client using them. We've already identified that they cause more problems than they solve.

fjricci:
I'm not sure that there's a good alternative. Swift debugging relies heavily on compilation info, and -serialize-debugging-options can cause build failures in subtle cases (which means that in the cases we care about, it's unusable). This doesn't even really add more uses of the compile flags, it just slightly widens the case where existing code to use the flags is used.

fjricci:
As an aside - added the test as lit test instead of python due to lower overhead (aka smaller diff) and higher reliability (the python version was failing for infrastructure-related reasons).

jrose-apple:
> I'm not sure that there's a good alternative. Swift debugging relies heavily on compilation info, and -serialize-debugging-options can cause build failures in subtle cases (which means that in the cases we care about, it's unusable). This doesn't even really add more uses of the compile flags, it just slightly widens the case where existing code to use the flags is used.

The kinds of things that are in the raw options are things like header maps, which are also the things that cause the most issues.

Swift debugging relies heavily on *search paths.* If -serialize-debugging-options is causing problems, that means the debugger will have problems regardless. The DWARF copy of the flags won't be better.

adrian-prantl:
Thanks for you effort, but I agree with Jordan that this is not the right approach. In my experience the primary reason for failing Clang imports is when we have too many (conflicting) header search options, piling more options from various sources on top won't make this better. What's your motivating example for this patch?

fjricci:
The issue we have is that we build some libraries using a certain path to a set of module files. We then export those libraries for use in applications. When the applications are built, they use the same set of module files but at a different path (for reasons that don't relate to Swift, but which we can't change). For developers who work on both the initial set of exported libraries and the applications, both paths exist on the system, and debugging (or building with serialized options) will fail due to duplicate definitions.

I know that our situation with different module paths for different libraries isn't necessarily expected, so we may be stuck carrying a local patch for this (although would be great to have an upstream solution).

jrose-apple:
This particular patch would run into the exact same problem, then.

fjricci:
Yeah, this patch is only intended to resolve the lack of serialized debugging options, it isn't intended to (and doesn't) solve the entirety of our particular use case. I'm okay with closing this and carrying locally if needed, but I'm still not sure that's the right decision. The existing code already has logic for missing serialized debugging options (`if (!got_serialized_options)`, etc), and this patch just fixes/improves that logic. If we don't want to support that, I understand, but presumably we should just delete that code then, rather than keeping it in a broken state? Keeping it around and also not fixing it doesn't seem like the right call to me.

PR #326, opened 2018-02-13
[CompilerType] Expose GetMangledTypeName() to ValueObject.

dcci:
The whole logic was already there, just not exposed yet. This
will be used as we want to pass the mangled type name to remote
mirrors.


dcci:
@swift-ci please test

dcci:
Note, once some other languages use this, we can move the whole logic upstream. But currently Swift is the only language requiring this support.

PR #329, opened 2018-02-17
Fix the cmake build and disable a flaky test

vedantk:


vedantk:
@swift-ci test Linux platform

vedantk:
@test OS X platform

vedantk:
@swift-ci test OS X platform

dcci (changes requested):
I think this is really good, but we shouldn't disable the test.
The reason why it's not passing on your machine is because it's picking the wrong debug server.
In fact, you can see that this is passing on the bot (at least I think). Am I wrong?

Thanks for doing this, BTW!

dcci:
Also, the reason why I don't really want to disable this is because it's testing a fairly fundamental feature in the swift lldb support. 
I think we might consider putting that bit on hold, and discuss a little offline about whether we can rewrite this test to not rely on the debug server behavior.

vedantk:
Hm, but if it's not passing locally with the same invocation the bots use (plus --lldb-build-with-cmake), I worry that other engineers will see the failure.

E.g until we know _why_ it's picking the wrong debug server and address that, it shouldn't be running.

vedantk:
@dcci and I discussed this offline, and agreed to disable the test for now.

dcci:
Go ahead Vedant!

PR #333, opened 2018-02-20
Disable ThreadIdsInJstopinfoTest as the stable version is flakey

jimingham:
This test has been failing about 50% of the time since we turned the unit tests back on.

The test and its related infrastructure have changed significantly on TOT, so it's not worth working on this version of the test.

### DIFF
--- a/tools/debugserver/source/CMakeLists.txt
+++ b/tools/debugserver/source/CMakeLists.txt
@@ -95,14 +95,14 @@ set(LLDB_CODESIGN_IDENTITY "lldb_codesign"
   CACHE STRING "Identity used for code signing. Set to empty string to skip the signing step.")
 
 if(NOT LLDB_CODESIGN_IDENTITY STREQUAL "")
-  set(DEBUGSERVER_PATH $<TARGET_FILE:debugserver>)
+  set(DEBUGSERVER_PATH $<TARGET_FILE:debugserver> CACHE PATH "Path to debugserver.")
 else()
   execute_process(
     COMMAND xcode-select -p
     OUTPUT_VARIABLE XCODE_DEV_DIR)
   string(STRIP ${XCODE_DEV_DIR} XCODE_DEV_DIR)
   set(DEBUGSERVER_PATH
-    "${XCODE_DEV_DIR}/../SharedFrameworks/LLDB.framework/Resources/debugserver")
+    "${XCODE_DEV_DIR}/../SharedFrameworks/LLDB.framework/Resources/debugserver" CACHE PATH "Path to debugserver.")
   set(SKIP_DEBUGSERVER True)
 endif()
 message(STATUS "Path to the lldb debugserver: ${DEBUGSERVER_PATH}")


jimingham:
@swift-ci please test

dcci:
Fine, but please open a `rdar://` as this needs to be reenabled once we rebase.

jimingham:
@dcci Not sure what you mean by that.  I'm only disabling this on stable.  If we rebase from stable we still won't want to work on this test because it is a very out-of-date version.  If we rebase from upstream-with-swift, we'll get the new version which isn't disabled.

dcci:
@jimingham I mean when we merge from `upstream-from-swift` we need to make sure we resolve the conflict correctly (i.e. we re-introduce this test).

BTW, I'm afraid this is not the correct way of disabling this test.
In CMake, if you have a file in a directory but that's not tracked, the build system will complain (presumably this will happen here as well).

jimingham:
@swift-ci please test

dcci:
ehm, not sure about commenting the whole test either. This is going to cause a lot of pain when we want to resolve merge conflicts. If you want to go for this, maybe `#if 0` the whole block?

Also, can you please squash your commits? Otherwise the history will become terribly messy. Thanks.

dcci:
The tests didn't start so I guess you're still good to make these changes I suggested.

dcci:
I think you just changed text indentation. Please undo that, and while you're there, squash. It's not great that there are 5 commits in the history for disabling a test.

dcci:
Sorry, s/you changed text indentation/you still have commented code/

jimingham:
You are right about that, how annoying.  I don't want to delete the file.  But #if 0 is nicer, I'll try that.  

I still don't understand your point about upstream-with-swift.  I'm only disabling this on stable.  I don't remember if we still have an automerger from stable->upstream-with-swift, but if we do it's going to horribly mangle this test anyway, and we'll just have to put the  UWS version back in place.  Otherwise the change will just sit on stable.  There's no intention of disabling this on UWS, at least not until we determine that the revised version of the test is still flakey.

jimingham:
@swift-ci please merge

dcci:
OK, fine re `upstream-with-swift`. Thanks for making these changes. Please squash your commits and then this is good to go for me.

dcci:
If you're unfamiliar with the practice, `GitHub` has an excellent article on the topic.
https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

This keeps the whole repo in a much better state (and also makes archeology much easier :)
Thanks for taking the time for doing this.

jimingham:
@swift-ci please test

vedantk:
@jimingham, here's how to squash these commits and update the PR:

$ git fetch origin
$ git rebase -i origin/stable
$ git push -f jimingham stable

shahmishal:
@swift-ci please test

dcci:
This seems to pass. Can you please unclutter the history rebasing/squashing (look at the previous comment) & commit this?

jimingham:
Well, that didn't work.  I'll try again with another test branch.

PR #334, opened 2018-02-20
[ExpressionParser] Clean up SwiftREPL::CreateInstance

vedantk:
This is a mostly-NFC change which makes SwiftREPL::CreateInstance easier
to read with early returns and references.

The only functional change is a bugfix for the call to
GetScratchTypeSystemForLanguage. If that call fails, we will now
correctly tear down the REPL process and the event handler thread.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
@@ -44,201 +44,236 @@ lldb::REPLSP SwiftREPL::CreateInstance(Status &err, lldb::LanguageType language,
                                        Debugger *debugger, Target *target,
                                        const char *repl_options) {
   if (language != eLanguageTypeSwift) {
-    return lldb::REPLSP();
+    // TODO: EnumerateSupportedLanguages should make checking for this
+    // unnecessary.
+    return nullptr;
   }
 
-  if (target) {
-    // Sanity check the target to make sure a REPL would work here.
-    if (!target->GetProcessSP() || !target->GetProcessSP()->IsAlive()) {
-      err.SetErrorString("can't launch a Swift REPL without a running process");
-      return lldb::REPLSP();
-    }
-
-    SymbolContextList sc_list;
-    target->GetImages().FindSymbolsWithNameAndType(
-        ConstString("_swift_release"), eSymbolTypeAny, sc_list);
-
-    if (!sc_list.GetSize()) {
-      err.SetErrorString("can't launch a Swift REPL in a process that doesn't "
-                         "have the Swift standard library");
-      return lldb::REPLSP();
-    }
-
-    // Check that we can get a type system, or we aren't going anywhere:
-    TypeSystem *type_system =
-      target->GetScratchTypeSystemForLanguage(nullptr, eLanguageTypeSwift, 
-                                              true, repl_options);
-    if (!type_system) {
-      err.SetErrorString("Could not construct an expression "
-                         "context for the REPL.\n");
-      return lldb::REPLSP();
-    }
-    // Sanity checks succeeded.  Go ahead.
-    SwiftREPL *repl = new SwiftREPL(*target);
-    REPLSP repl_sp(repl);
-    repl->SetCompilerOptions(repl_options);
-    return repl_sp;
-  } else if (debugger) {
-    const char *bp_name = "repl_main";
-
-    FileSpec repl_executable;
-
-    if (HostInfo::GetLLDBPath(ePathTypeSupportExecutableDir, repl_executable)) {
-      repl_executable.GetFilename().SetCString("repl_swift");
-      std::string repl_exe_path(repl_executable.GetPath());
-      if (repl_executable.Exists()) {
-        const char *target_triple = NULL;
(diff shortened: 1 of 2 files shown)


vedantk:
@swift-ci test

source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
```diff
@@ -44,193 +44,229 @@ lldb::REPLSP SwiftREPL::CreateInstance(Status &err, lldb::LanguageType language,
                                        Debugger *debugger, Target *target,
                                        const char *repl_options) {
   if (language != eLanguageTypeSwift) {
+    err.SetErrorString("can only create a REPL for Swift");
```
dcci: This seems like a functional'ish change for swift. Is there any chance we can trigger this and add a test?

dcci (commented):
Thank you very much! This code is ow much much better and easier to understand. I have a comment inline.

source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
vedantk: Good question, I'll take a look! This might end up depending on the cmake change for repl_swift you're looking into.

vedantk: Ah, this is dead code. EnumerateSupportedLanguages already filters this out. I'll update the patch.

vedantk:
@swift-ci test

source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
```diff
@@ -43,202 +43,232 @@ using namespace lldb_private;
 lldb::REPLSP SwiftREPL::CreateInstance(Status &err, lldb::LanguageType language,
                                        Debugger *debugger, Target *target,
                                        const char *repl_options) {
-  if (language != eLanguageTypeSwift) {
+  assert(language == eLanguageTypeSwift && "can only create a REPL for Swift");
```
jimingham: This isn't right.  If you asked for a eLanguageTypeC++ REPL, then the REPL Factory would go through all the things that could create REPL's asking if it could create one for C++.  The proper response is to return an empty REPLSP meaning, "no I can't create a REPL for C++".  Then the factory for REPL's would go on to the next one.

vedantk:
@swift-ci test

jimingham (changes requested):
The assert at the beginning of CreateInstance, is wrong, see inline comment.  Otherwise this is nicer, thanks for cleaning this up.

source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
jimingham: No, it was right the way it was.  In lldb CreateInstance methods always return an empty one of the thing they were supposed to create.  That's a general pattern.  The more general support for REPL's is nascent because we only support one of them.  But we shouldn't change how CreateInstance works.

source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
vedantk: I'll switch back to the original behavior.

Ideally EnumerateSupportedLanguages would make this unnecessary, and a test would have caught this :/.

vedantk:
@swift-ci test

vedantk:
@swift-ci test

jimingham:
> On Feb 20, 2018, at 3:40 PM, Vedant Kumar <notifications@github.com> wrote:
> 
> @vedantk commented on this pull request.
> 
> In source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp <https://github.com/apple/swift-lldb/pull/334#discussion_r169496184>:
> 
> > @@ -43,202 +43,232 @@ using namespace lldb_private;
>  lldb::REPLSP SwiftREPL::CreateInstance(Status &err, lldb::LanguageType language,
>                                         Debugger *debugger, Target *target,
>                                         const char *repl_options) {
> -  if (language != eLanguageTypeSwift) {
> +  assert(language == eLanguageTypeSwift && "can only create a REPL for Swift");
> I'll switch back to the original behavior.
> 
> Ideally EnumerateSupportedLanguages would make this unnecessary, and a test would have caught this :/.
> 
> 

EnumerateSupportedLanguages is only used as a convenience so that if the sum of languages that all the REPL plugins installed == 1, you can say —repl w/o having to say —repl-language.  We could also use it to do command completion for “expression —repl —language <TAB>.  But it isn’t used in the code path that actually creates the REPL.  That’s the way it should be as it keeps this in line with all the other Create factory methods.

Feel free to add a test!

Jim

dcci:
Vedant, feel free to merge this after https://github.com/apple/swift-lldb/pull/340 goes in.
I'll start CI testing in the meanwhile.

dcci:
@swift-ci please test

vedantk:
I'll make a small update to this PR and kick off testing again, now that we have coverage from https://github.com/apple/swift-lldb/pull/340.

vedantk:
@swift-ci test

vedantk:
@swift-ci test

PR #335, opened 2018-02-21
[SwiftRepl] Improve the CMake build & simplify.

dcci:
This is a grab bag of improvements for the repl. There are
several issues.
1) The repl needs a dummy target to run. There were two, one
written in C, the other in swift. The latter doesn't really
work and the only reason why nobody noticed is that only the
CMake build used it and there was no test whatsoever. The good
news is that's unneeded and can be removed.
2) The way the dummy executable was built with CMake was wrong
and unnecessarily complicated. This commit rewrites the logic
to be more readable (and commented).
3) This logic expects optimizations to be turned off, but
that was delegated to the build system. Now it's explicit
in the code via attributes, so that we don't have to duplicate
the logic across CMake/XCodebuild.

The logic is still fairly fragile as we dlopen() a library
looking inside @rpath. After speaking with Louis, he told me
that we shouldn't rely on this behaviour (and I tend to agree).
It's more complicated to fix because it involves touching also
the xcodebuild. This is left as a follow-up.


dcci:
@swift-ci please test

jimingham:
The comment explaining the dlopen & weak linking (sic) to libswiftCore.dylib is:

    * Weak link libswiftCore.dylib from repl_swift to sever a build dependency on OS X.
    * In repl_swift.c, open libswiftCore lazily, and only look in the rpath.

I don't remember what the "build dependency" issue was - this was a Sean comment, so we can't ask him what's up.  

But the other part of this is trying to make sure you only ever pull in the libswiftCore that comes with lldb - it explicitly sets the RPATH to:

@executable_path/../../../../../Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx @executable_path/../../../../../../usr/lib/swift/macosx @executable_path/../../../../../../../usr/lib/swift/macosx @executable_path/LLDB.framework/Resources/Swift/macosx @executable_path/Swift/macosx

That's missing from your cmake version.

dcci:
I looked and it seems my `$RPATH` is set to nothing before the invocation so I'm not sure whether is required. Speaking with Louis, he said we really shouldn't mess around with `@executable_path` either as it's very dangerous.

jimingham:
Why is @executable_path dangerous?

In this case, it is important to make sure that you don’t by any means load a version of libswiftCore.dylib that isn’t the one you were built with.  If you don’t do that the REPL will fail.  On a system with lots of SDK’s installed, and then when running this to do testing out of a built tree, etc, there are lots of wrong choices you could make.    The RPATH ensures that won’t happen.  The fact that your version worked may be by luck, and it would be better not to count on that.

Jim

dcci:
The version specified in the `rpath` directory will have precedence on the others. As long as the build fills the library in the correct path, we should be fine, I think.
In the short term we're going to get rid of this code anyway. If this fails, please assign me a bug and I'll deal with the fallout.

tools/repl/swift/CMakeLists.txt
```diff
@@ -1,50 +1,10 @@
-set(LLVM_NO_RTTI 1)
+# Set the correct rpath to locate libswiftCore
+set(CMAKE_EXE_LINKER_FLAGS "-Wl,-rpath \
+    -Wl,${CMAKE_BINARY_DIR}/${LLDB_FRAMEWORK_INSTALL_DIR}/${LLDB_FRAMEWORK_RESOURCE_DIR}/Swift/macosx")
```
vedantk: Should we move this, and the add_lldb_tool invocation, into the `if (LLDB_BUILD_FRAMEWORK)`?

dcci: Probably. I'll do this tomorrow.

dcci:
@swift-ci please test linux

dcci:
@swift-ci please test

dcci:
@swift-ci please test

dcci:
@swift-ci please test

dcci:
@swift-ci please test linux

dcci:
@swift-ci please test linux

dcci:
@swift-ci please test linux

PR #336, opened 2018-02-21
This test is flakey on Linux, disable it.

jimingham:
The code has varied significantly on TOT so fixing
it on stable is not worth the effort.

<rdar://problem/37725893>


jimingham:
@swift-ci please test

PR #337, opened 2018-02-21
Linux / cmake: Install the clang resource directory to the expected P…

adrian-prantl:
…OSIX path.

HostInfoPosix::ComputeClangDirectory() expects the clang resource
directory to be in lib/lldb/clang/$VERSION. This patch fixes the
Swift-specific CMake to install it there instead of in lib/lldb/clang/.

<rdar://problem/37354542>
https://bugs.swift.org/browse/SR-6954

(cherry picked from commit 0f090945431a91eb0a6a89a7c8dd85b0c0d5c1f8)

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/foundation/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/foundation/Makefile
@@ -1,0 +1,3 @@
+LEVEL = ../../../make
+SWIFT_SOURCES := main.swift
+include $(LEVEL)/Makefile.rules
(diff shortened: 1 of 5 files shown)


adrian-prantl:
@swift-ci test

adrian-prantl:
@swift-ci nominate
Explanation: Changes the install location of the clang resource directory inside of LLDB's install root from lldb/clang to lldb/clang/$clang-version, which is the location where LLDB expects it to be on POSIX systems. Without this change it is not possible to debug Swift modules that depend on Clang modules that import Clang resource headers, such as swift-corelibs-foundation.
Scope: This is a CMake change and thus affects Linux only.
Radar: rdar://problem/37354542
https://bugs.swift.org/browse/SR-6954
Risk: Zero for Darwin, low risk/high payoff for Linux.
Testing: New LLDB testcase that tests importing Foundation on both Darwin and Linux.

fredriss (approved):
This looks safe and correct

adrian-prantl:
@swift-ci test

adrian-prantl:
@swift-ci test

PR #340, opened 2018-02-22
[REPL] Rewrite dictionary testing & improve infra to be more reliable.

dcci:
This now uses lit. It's shorter and hopefully more reliable than
pexpect. As part of this, I also did all the plumbing for testing
the repl in lldb, so what I've done here can be used for other repl
tests. While I'm at it, remove the old test, we don't need it anymore.

<rdar://problem/37407231>


dcci:
@swift-ci please test

lit/SwiftExpr/Dict.test
```diff
@@ -0,0 +1,50 @@
+# Test that dictionaries work in the REPL.
+# REQUIRES: darwin
+
+# RUN: %lldb --repl < %s | FileCheck %s --check-prefix=DICT
+
+[1:2, 2:3, 3:9]
+# DICT: Welcome to Swift version {{.*}} (LLVM {{.*}}, Clang {{.*}}, Swift {{.*}})
+# DICT: $R0: [Int : Int] = 3 key/value pairs {
+# DICT:   [0] = {
```
vedantk: Now that we're using FileCheck, we can tighten these checks up with -NEXT. But it'll do as a follow-up.

vedantk (approved):
LGTM, thanks!

dcci:
I decided to be a little more conservative in this. If you think it's always safe also in the repl to use `CHECK-NEXT` I'll go for it.

vedantk:
If there's repl output between those check lines, I'd be concerned / would want a test failure to learn about the issue. I don't feel very strongly about this and wouldn't block the PR over it.

dcci:
Also, thanks to @jrose-apple for suggesting the right way of testing multiple commands via `<` :)

jrose-apple:
Uh, note that you should use `//` for your comments, then! `#` is not a comment character in Swift.

dcci:
@jrose-apple thanks! I'll update my pull request accordingly.

JDevlieghere (commented):
Great, thanks for taking the time to improve these tests! 
LGTM (modulo comment fix)

dcci:
@swift-ci please test

dcci:
@swift-ci please test

PR #341, opened 2018-02-22
Fix python build argument

keith:
Previously this concatenated "pythonswift/utils" this was broken in
80e032cb3fc68d1c56ed0c112f1fc81baf1a9454

### DIFF
--- a/scripts/build-swift-cmake.py
+++ b/scripts/build-swift-cmake.py
@@ -205,7 +205,7 @@ else:
     elif args.use_system_debugserver:
         build_script_impl_arguments += ['--lldb-use-system-debugserver']
 
-args = ["python" + os.path.join("swift", "utils", "build-script")] + \
+args = ["python", os.path.join("swift", "utils", "build-script")] + \
     build_script_arguments + ["--"] + build_script_impl_arguments
 
 print(" ".join(args))


dcci:
I'm at a loss here trying to understand how we didn't see this breakage anywhere.
@keith can you please tell me how you build so I can try to repro this locally?

dcci:
Also, @AndrewSB tried this on Windows and it didn't break for him so I want to try to understand this better :)

AndrewSB:
In hindsight I’m surprised that worked for me. There _should_ be a space between those strings.
I was able to successfully build on windows with my change though. Very curious. I’ll try building again later today with my old commit as the tip and report back

keith:
I just did an update checkout and then:

```
$ ./lldb/scripts/build-swift-cmake.py
pythonswift/utils/build-script --debug-swift --debug-lldb --skip-build-benchmarks --extra-cmake-options=-DCMAKE_CXX_FLAGS=-fno-limit-debug-info -- --build-swift-static-stdlib=1 --install-swift --install-lldb --install-destdir /Users/ksmiley/dev/oss-swift/install --swift-install-components=compiler;clang-builtin-headers;stdlib;stdlib-experimental;sdk-overlay;editor-integration;tools;testsuite-tools;dev --lldb-use-system-debugserver
Traceback (most recent call last):
  File "./lldb/scripts/build-swift-cmake.py", line 213, in <module>
    return_code = subprocess.call(args)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

dcci:
I see. FWIW, I'm not entirely sure why lldb has to provide yet another layer of indirection when you can just use build script. I wonder whether we can just get rid of the script and standardize on using `build-script` ? That way we have a single way of building, and if something fails, we realize as it fails on the bots.

(currently, nothing in the build jobs uses `build-swift-cmake.py`, that's why we didn't notice).

@Keith/ @AndrewSB what do you think?

keith:
I'm fine with whatever you want to do here. I will say that it has been very nice being able to run this script with no arguments and not really worry about any configuration, so I guess if it's deleted we might want to provide some docs on common build arguments you need to pass.

dcci:
@Keith if you use it, I'll leave it in place. I'm only a little worried that this is something not tested by a bot.

dcci:
This LGTM, anyway.

keith:
Makes sense. I'm using a mix of this too bootstrap the build, and the cmake setup to actually iterate on changes since that's so much faster, especially if you can test your changes without LLDB.framework.

dcci:
@swift-ci please test

dcci:
@swift-ci please test

dcci:
@Keith feel free to merge this yourself once the testing completes.

keith:
AFAIK I don't have access to do that 😄 🤔

AndrewSB:
Thanks for fixing my mistake @keith 🙃

PR #342, opened 2018-02-22
[PATCH] [test] Introduce the skipUnlessSwiftPR decorator

JDevlieghere:
This determines whether or not a test is skipped due to not being marked
`swiftpr` in the wrapper function, i.e after the test method is called,
instead of before.

This fixes strange issues with categories which resulted in the wrong
set of tests being marked `swiftpr` and running accidentally. It also
fixes discrepancies which arise between the single and multi threaded
test runners due to attempts to edit an attribute on a function instance
shared between multiple inline tests.

Thanks to Vedant Kumar for helping me come up with a solution!

### DIFF
--- a/packages/Python/lldbsuite/test/configuration.py
+++ b/packages/Python/lldbsuite/test/configuration.py
@@ -36,9 +36,6 @@ skipCategories = ["darwin-log", "frame-diagnose"]
 if platform.system() == 'Linux':
     skipCategories.append('watchpoints')
 
-# use this to track per-category failures
-failuresPerCategory = {}
-
 # The path to LLDB.framework is optional.
 lldbFrameworkPath = None
 
--- a/packages/Python/lldbsuite/test/decorators.py
+++ b/packages/Python/lldbsuite/test/decorators.py
@@ -160,7 +160,8 @@ def _decorateTest(mode,
                   debug_info=None,
                   swig_version=None, py_version=None,
                   macos_version=None,
-                  remote=None):
+                  remote=None,
+                  categories=None):
     def fn(self):
         skip_for_os = _match_decorator_property(
             lldbplatform.translate(oslist), self.getPlatform())
@@ -177,6 +178,16 @@ def _decorateTest(mode,
             triple, lldb.DBG.GetSelectedPlatform().GetTriple())
         skip_for_remote = _match_decorator_property(
             remote, lldb.remote_platform is not None)
+
+        skip_for_categories = False
+        if configuration.useCategories:
+            # If we're using categories, we should skip the test if (1) no
+            # categories are set or (2) none of its categories are enabled.
+            skip_for_categories = not categories or all(
+            [
+                cat not in configuration.categoriesList
+                for cat in categories
+            ])
 
         skip_for_swig_version = (
             swig_version is None) or (
@@ -208,7 +219,8 @@ def _decorateTest(mode,
                       (swig_version, skip_for_swig_version, "swig version"),
                       (py_version, skip_for_py_version, "python version"),
                       (macos_version, skip_for_macos_version, "macOS version"),
-                      (remote, skip_for_remote, "platform locality (remote/local)")]
+                      (remote, skip_for_remote, "platform locality (remote/local)"),
+                      (categories, skip_for_categories, "missing categories")]
         reasons = []
         final_skip_result = True
         for this_condition in conditions:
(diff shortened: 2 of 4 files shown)


JDevlieghere:
@vedantk Do we want to keep the SwiftPR decorator? The way things are currently implemented, it can remain a test category like before.

JDevlieghere:
@swift-ci please test

dcci:
This has exposed several failures. I don't think it's your fault, but we need to investigate them anyway.

```
Issue Details
=============
FAIL: test_extension_self (lang/swift/expression/class_constrained_protocol/TestClassConstrainedProtocol.py)
FAIL: test_extension_weak_self (lang/swift/expression/class_constrained_protocol/TestClassConstrainedProtocol.py)
FAIL: test_static_closure_type (lang/swift/closures/TestSwiftClosures.py)
ERROR: test_generic_closure_call (lang/swift/closures/TestSwiftClosures.py)
ERROR: test_generic_closure_type (lang/swift/closures/TestSwiftClosures.py)
ERROR: test_static_closure_call (lang/swift/closures/TestSwiftClosures.py)
ERROR: test_with_dwarf (lang/swift/foundation/TestSwiftFoundation.py)
ERROR: test_with_dwo (lang/swift/foundation/TestSwiftFoundation.py)
ERROR: test_with_gmodules (lang/swift/foundation/TestSwiftFoundation.py)
```

Are you able to reproduce locally?

JDevlieghere:
I see these failures as well , at least after rebasing. I'm pretty sure they weren't there when I created the patch, so they might be recent. Anyway, I'll have a look!

vedantk:
> Do we want to keep the SwiftPR decorator? The way things are currently implemented, it can remain a test category like before.

@JDevlieghere I think it's a convenient utility to have.

vedantk:
And the failing tests don't appear to be in the swiftpr category, hm..

jimingham:
All of the swift tests that try to build shared libraries currently fail because they don’t use lldbtest.build(), and so don’t work in the out-of-tree build world.  I’m in the middle of fixing that, so don’t worry about any of those failures yet.

A lot of the other swift tests fail because even though they build correctly, they don’t get the actual a.out from the build directory correctly.  This failure is easy to spot they will get an error about not getting a valid target.  I’ve fixed some of those already, but not all.  If other folks want to bang on them please do, but if possible don’t fix them by hand, but rather convert them to lldbutil.run_to_source_breakpoint or lldb.run_to_name_breakpoint.  That way we keep all this logic centralized and it makes the tests much easier to read. The last couple of test checkins I made to upstream-with-swift show examples of doing this.  In general, any time you touch a test it would be great to convert them to this startup method, as it reduces a ton of duplicated boilerplate.

Jim

vedantk:
> All of the swift tests that try to build shared libraries currently fail because they don’t use lldbtest.build(), and so don’t work in the out-of-tree build world.  I’m in the middle of fixing that, so don’t worry about any of those failures yet.

@jimingham does this affect the stable branch?

jimingham:
No, the stable branch doesn’t have Adrian’s build-out-of-tree changes.  That’s only fairly recently showed up on upstream-with-swift.  We knew that we were going to have to fix up all the swift tests there once these changes were merged into upstream-with-swift, just like Adrian had to do for all the other tests when he originally made this patch.

Jim

JDevlieghere:
Apparently my understanding of the logic was inverted for the case were no categories were set.

JDevlieghere:
@swift-ci please test

JDevlieghere:
@swift-ci please test

vedantk:
@jimingham if the failures you're seeing aren't on stable, then we don't need to worry about them in this PR.

JDevlieghere:
The failing test defines no decorators, which is why the categories are never checked.

JDevlieghere:
Also, not to self, I think we can do this much cleaner by using `skipUnless`. I'll have a look at both tomorrow!

vedantk:
It looks like we're still seeing some unexpected results with the skipUnlessSwiftPR decorator. I've sketched out an alternate / hackier solution here: https://github.com/apple/swift-lldb/pull/354 (treat all inline tests as swiftpr tests).

vedantk:
Shall we close this now that #354 has been merged?

JDevlieghere:
Jup. Thanks Vedant!

PR #343, opened 2018-02-22
[REPL] Replace a flakey NSString test with a robust version.

dcci:
This was disabled, sadly because it was flakey.
The new version is shorter and more reliable.

<rdar://problem/37407231>


dcci:
@swift-ci please test

PR #346, opened 2018-02-23
Revert "Linux / cmake: Install the clang resource directory to the ex…

adrian-prantl:
…pected POSIX path."

This reverts commit c29f16288bc2fd3959cd1eef13fc6c41b6051800
because it caused

https://bugs.swift.org/browse/SR-7065
REPL broken on Linux (cannot import anything)

It turns out that this breaks the REPL on Linux, which is an even worse regression, so I'm going to revert the change from swift-4.1-branch again. I'm still convinced that this is the right resolution for this issue, but we shouldn't land this before we sort out the REPL problem.


adrian-prantl:
@swift-ci test

adrian-prantl:
I will also make sure we have a REPL test on Linux going forward.

dcci:
The tests I wrote for the repl (this week) have been macOS only as they rely on foundation. I think we should just make sure they work on Linux as well (fun fact, they're failing, but I thought it was because the repl wasn't unsupported, instead it was broken but we never noticed).

PR #350, opened 2018-02-27
Convert all our non-xfailed REPL tests to lit tests

vedantk:
This converts all of our REPL tests which use pexpect-backed routines to lit tests, which are more reliable.

There is one exception: I've left TestREPLBreakpoints.py alone because it's still x-failed. We'll need to look into that one separately.


vedantk:
@swift-ci test

dcci (changes requested):
I think you might want to make these `REQUIRES: Darwin` as we haven't finished to add the CMake support to Linux for the repl. I think these are going to fail on the bot. Other than that, it looks fine.

vedantk:
@dcci gotcha, I’ll look into fleshing out the Linux cmake support tomorrow. I’d rather not land this without deleting the pexpect tests that need to be replaced.

lit/SwiftExpr/BreakpointREPL.test
```diff
@@ -0,0 +1,13 @@
+// Test that we can set breakpoints in the REPL.
+
+// RUN: %lldb --repl < %s | FileCheck %s
+
+func foo() -> Int {
+  return 0
+}
+
+:b foo
```
jimingham: You switched this from a file & line breakpoint in the original test to a function breakpoint.  It would be fine to add a function name test but you shouldn't remove the test for file & line breakpoints.

lit/SwiftExpr/RecursiveClass.test
```diff
@@ -0,0 +1,43 @@
...
+// CHECK-NEXT: }
+
+a.aFoo = a
+a
+// CHECK: $R0: Foo = {
+// CHECK-NEXT:   aFoo = 0x{{[0-9a-fA-F]+}} {...}
+// CHECK-NEXT:   x = "Hello World"
+// CHECK-NEXT: }
+
+// Also test the more basic non-recursive case.
```
jimingham: The Recursive class test was testing a particular tricky bit of the REPL - that you could use an optional of a class in it's definition.  Combining it with the basic "can we define classes" tests seems confusing to me.  Now if I want to add another test of class behavior where would I put it?  We should either keep this one narrowly focused and make a generic class behavior test (by just changing the name of the test...) or make another test for collecting more general class behavior.

lit/SwiftExpr/TypeLookup.test
```diff
@@ -0,0 +1,13 @@
+// Test that type lookup chooses the right language.
```
jimingham: Can we change the name of this test.  You didn't do this, but...  The point of the test is that, having imported Foundation, we can lookup a type from it.  So it's really a test of importing Foundation, type lookup is really just the means.  It would actually be better to have the test try to use Foundation...  And testing the functionality of "type lookup" which is a straight-up lldb command isn't best done in a REPL test...

lit/SwiftExpr/SimpleExpressions.test
```diff
@@ -0,0 +1,71 @@
+// Test that we can define and use basic functions/expressions in the REPL.
```
jimingham: This one does pretty much everything TestREPLBasic does.  Add a simple variable assignment, and then you should be able to delete that test as well.

jimingham (changes requested):
Why is the directory for these tests called SwiftExpr?  That makes it sound like it is testing the swift expression parser, but it's actually testing REPL behavior.  Some of that tests generic expression parser behavior, but the REPL really only does a small subset of the full expression parser's jobs.  Be better to call this SwiftREPL.

lit/SwiftExpr/BreakpointREPL.test
jimingham: That's the hard part, after all - making sure that subsequent REPL invocations keep a useful line number mapping...

vedantk: This is a new test, it's not meant to replace anything (I mentioned this in the PR description). It's just a basic sanity check.

lit/SwiftExpr/RecursiveClass.test
vedantk: Sure, I'll split it back out.

lit/SwiftExpr/TypeLookup.test
vedantk: I'll find a better name. Maybe ImportFoundation.test.

lit/SwiftExpr/SimpleExpressions.test
vedantk: Will do, thanks!

vedantk:
@jimingham Thanks for your comments, I believe I've addressed them now.

I still need to get things working on Linux before we can kick off testing.

vedantk:
@swift-ci test

vedantk:
@swift-ci test

vedantk:
Great, tests look good on macOS/Linux now. Is this fine to merge?

jimingham:
This is fine.  If you add a file & line test to the breakpoints test, then you can delete TestREPLBreakpoints.py which gets us further to no pexpect based tests, but you can do that as a separate commit.

jimingham (approved):

dcci:
The plumbing is correct, I only am uncertain about a macro, but if that works (and you tested) please go ahead

tools/repl/swift/main.c
```diff
@@ -12,9 +12,10 @@
 
 #ifdef __APPLE__
 #include <mach/mach_time.h>
+#include <dlfcn.h>
 #endif
 
-#ifdef __APPLE__
+#ifdef __gnu_linux__
```
dcci, line 18: Linux has kind aleph-zero different ways of telling to the preprocessor "am I running Linux", so we might change this in the future.

tools/repl/swift/main.c
```diff
@@ -30,6 +31,8 @@ int main() {
 #ifdef __APPLE__
   // Force loading of libswiftCore.dylib, which is not linked at build time.
   dlopen("@rpath/libswiftCore.dylib", RTLD_LAZY);
+#else
```
dcci, line 34: `#elif Linux` ?

dcci (approved):

tools/repl/swift/main.c
vedantk: This was the first macro I found in clang that can be used to identify linux targets.

But, yes, we'll absolutely need to rethink this later.

vedantk:
Testing on rad-linux looks good. The macro guards in the dummy repl's main.c aren't great, but we can iterate on it as needed.

PR #354, opened 2018-02-27
[test] Treat inline tests as part of the swiftpr category

vedantk:
1) Treat all inline tests as if they are a part of the swiftpr category.
This means "dotest.py -G swiftpr" will include all inline tests, unless
they are disabled in the MakeInlineTest constructor.

2) In the MakeInlineTest constructor, only enable a test if it has an
appropriate debug info kind. E.g, don't enable the dsym variant of a
test if "--skip-categories dsym" has been specified.

3) In the MakeInlineTest constructor, treat all tests as a part of the
swiftpr category *unless* they have decorators. An inline test with
decorators is usually either skipped on some platform or x-failed: for
now, we will not consider these tests for PR testing.

4) Remove the 'swiftpr' decorators from inline tests as they are not
reliable. They cause the wrong set of tests to be included in the PR
testing category.


packages/Python/lldbsuite/test/lldbinline.py
```diff
@@ -212,6 +212,24 @@ def ApplyDecoratorsToFunction(func, decorators):
...
+    # If the platform is unsupported, skip the test.
+    if not test_categories.is_supported_on_platform(debug_flavor,
+            target_platform, configuration.compiler):
+        return False
+
+    # If the debug flavor is skipped, skip the test.
+    if debug_flavor in configuration.skipCategories:
+        return False
+
+    # If there are any decorators, and we're in swift PR testing mode, skip the
```
JDevlieghere, line 225: Can you explain why this is necessary? I understand that we do not care, but I don't like this localized exception which makes looking at an inline test ever more opaque than it already is.

packages/Python/lldbsuite/test/test_result.py
```diff
@@ -109,6 +108,14 @@ def getCategoriesForTest(self, test):
         """
         Gets all the categories for the currently running test method in test case
         """
+
+        # It isn't possible to reliably attach categories to inline tests
+        # because they all share the same instance method: punching in the
+        # category there causes the wrong set of tests to be included in it.
```
JDevlieghere: I'd rephrase this, maybe something like "punching in the category for one inline test causes all inline tests to belong to that category".

packages/Python/lldbsuite/test/lldbinline.py
vedantk: It's necessary to keep expectedFail tests out of swift PR testing (I'll update the comment to reflect this). Happy to do this some other way if possible!

JDevlieghere: Fair enough. This is also a problem for non-inline tests, but there we would just remove the `swiftpr` category if a test is XFAIL'ed/skipped while here we can't do that, hence why we need this. Am I understanding that correctly?

vedantk: Exactly.

JDevlieghere: Alright, makes sense, maybe worth mention that it's more a matter of "can't" rather than "don't want".

vedantk: Does the new comment look any better?

JDevlieghere: Yes, thanks!

JDevlieghere (approved):
For the record, I can't LGTM this without saying that I very much dislike this solution, but since we don't have anything better and because this fixes swift-PR testing, we don't really have choice.

vedantk:
@swift-ci test

vedantk:
@swift-ci test Linux platform

PR #356, opened 2018-03-02
[Repl] Don't crash when passing additional arguments.

dcci:
The assertion was overly aggressive, we can actually
have a non-empty set of options in this codepath. While
here, add a test to make sure this doesn't regress.

<rdar://38042866>

### DIFF
--- a/lit/SwiftREPL/CrashArgs.test
+++ b/lit/SwiftREPL/CrashArgs.test
@@ -1,0 +1,4 @@
+// Make sure we don't crash if we pass args to the repl.
+
+// RUN: %lldb --repl="-some-argument" | FileCheck %s --check-prefix=SWIFT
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci please test

jimingham:
The swift driver always invokes the repl by:

/Volumes/Gonzales/Xcodes/Xcode.app/Contents/Developer/usr/bin/lldb --repl="-enable-objc-interop -sdk /Volumes/Gonzales/Xcodes/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk -color-diagnostics"

Why didn't that cause a crash when your options did?

dcci:
@jimingham that's a good point, I'll check it.

jimingham:
Ah, I see.  This all came from Adrian's refactor of CreateInstance.  The released versions from stable don't have this assert.

We should probably run the REPL in our testing the same way the swift driver runs it.

PR #357, opened 2018-03-02
Fix swift AST section name for COFF objects

fjricci:
According to: `swift/include/swift/Basic/Dwarf.h`,
the correct AST section name for COFF is "swiftast".


fjricci:
@swift-ci please test

fjricci:
https://github.com/apple/swift/blob/master/include/swift/Basic/Dwarf.h for reference

fjricci:
Also verified that this is the section name created when I run `swift -modulewrap` on a windows target locally, not sure whether we really do windows testing here currently.

fjricci:
ping

dcci:
LGTM but maybe @jimingham or @adrian-prantl have opinions on this. 
I'm not sure what's the best way to test this either.

dcci:
@fjricci isn't working on this anymore, we can resurrect if we want to.

PR #360, opened 2018-03-05
[REPL] Build the dummy executable always with `-g0`.

dcci:
This is just an implementation detail, and nobody touches it
anyway. Given that's a C program, lldb frame language recognition
gets confused, so the repl malfunctions. It only manifests in
debug builds, that's the reason why nobody noticed (until now).

An alternative would be that of having the dummy executable
being written in swift. At some point this was the case, but
then it was converted to a C program. Eventually we might want
to re-evaluate the choice.

<rdar://problem/38121735>


dcci:
@swift-ci please test

PR #361, opened 2018-03-05
[SwiftLanguage] Enforce a valid ASTContext when resolving archetypes.

dcci:
The callers now have to provide an AST all the time. In case they
don't we still bail out instead of crashing, but at least in assert
build we notice the violation sooner rather than later.

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -319,9 +319,8 @@ public:
 
   bool IsRuntimeSupportValue(ValueObject &valobj) override;
 
-  virtual CompilerType
-  DoArchetypeBindingForType(StackFrame &stack_frame, CompilerType base_type,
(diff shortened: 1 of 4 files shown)


dcci:
@swift-ci please test

jimingham:
I would go the other way with this change.  CompilerTypes always know their AST Context's (the SwiftASTContext is the TypeSystem for Swift types.)  In some cases you might be importing a type from one TypeSystem to another.  But in this case you want the CompilerType to use its own TypeSystem, so don't pass two things, just pass the CompilerType and fetch the TypeSystem from it.

dcci:
I think this is a really good suggestion. I'll make the change soon & resubmit the tests.

dcci:
Addressed Jim's comments. Thanks!

dcci:
@swift-ci please test

dcci:
@swift-ci please test

PR #362, opened 2018-03-06
[test] Update the swift stepping behavior test

vedantk:
apple/swift#15025 fixes the stepping behavior for switches, so we need
to update our test accordingly.


vedantk:
https://github.com/apple/swift/pull/15025

@swift-ci Please test

PR #363, opened 2018-03-07
Make the clang module cache setting available without a target

adrian-prantl:
It turns out that setting the clang module cache after LLDB has a
Target can be too late. In particular, the Swift language plugin needs
to know the setting without having access to a Target. This patch
moves the setting into the *LLDB* module cache, where it is a global
setting that is available before any Target is created and more
importantly, is shared between all Targets.

rdar://problem/37944432

Differential Revision: https://reviews.llvm.org/D43984

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@326628 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit f577c4eb9b63878171dfda7468d1f86c24b4b946)


adrian-prantl:
@swift-ci test

adrian-prantl:
Abandoning. This is too difficult to cherry-pick onto stable., I'll put it on the llvm-swift5-transition branch instead.

PR #365, opened 2018-03-08
 Swift: Disable headermap include paths in Swift's ClangImporter by de

adrian-prantl:
…fault.

This a big hammer. Based on the assumption that a well-formed
framework module will not need any headermaps to be imported, this
patch strips out all -I.*\.hmap ClangImporterOptions by default.

<rdar://problem/35732279>


adrian-prantl:
@swift-ci test

jimingham:
I feel uncomfortable supporting this much generic code change just on github.  Maybe you can add the "don't use spotlight" property as the excuse for it to exist on llvm.org?

adrian-prantl:
The generic code also exists on llvm.org for the clang modules cache setting, but that change depends on clang-1000 and is therefore only on the llvm-swift5-transition branch on github.

adrian-prantl:
I'm going to wait 24 hours until stable has been rebased on the newer clang and then update this with the much smaller patch instead.

PR #366, opened 2018-03-08
Remove intended divergence between the github lldb and llvm.org lldb

jasonmolenda:
Remove intended divergence between the github lldb and llvm.org lldb
sources.  Todd mis-identified the nature of the failure and xfailed
this, but it has been correctly fixed on llvm.org since then.

(cherry picked from commit 3323f58a7231f530513406b8832cd1d075d1921f)


jasonmolenda:
@swift-ci please test

PR #368, opened 2018-03-09
[SwiftLanguageRuntime] Delete unused demangler functionality.

dcci:
I originally thought this should live in the compiler, to then
realize, this is actually unused. Jim tells me this was needed
at some point, but it's not anymore, so let's nuke this.

It was the last bit of demangler functionality in the debugger,
now everything has been moved to the compiler or removed.

<rdar://problem/37710513>

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -215,10 +215,6 @@ public:
   
   static bool IsSwiftClassName(const char *name);
   
-  static bool IsMetadataSymbol(const char *symbol);
-  
-  static bool IsIvarOffsetSymbol(const char *symbol);
-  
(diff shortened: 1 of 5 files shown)


dcci:
@swift-ci please test

jimingham:
Yes, this was all code we needed when we were taking apart metadata and figuring out ivar offsets by hand.  But we don't do that any more, we use RemoteAST for that purpose, so we don't need this code anymore.

jimingham (approved):
Yup, we don't need to track these symbols specially anymore.

jimingham:
You should probably also remove the test:

variables/class/TestSwiftMetadataSymbols.py

Since we don't need to handle metadata symbols specially we don't need to fix this test for identifying them in the new mangling scheme.

PR #372, opened 2018-03-10
Fix standalone lldb cmake build.

gottesmm:
The code here works in the in-tree version of the lldb cmake build. But when we
do out of tree using llvm-config, we do not have access to the clang target. If
we included llvm's targets directly via the exported llvm cmake configuration,
this would work, but for whatever reason lldb isn't doing that (I thought it
was... but maybe my memory is wrong).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,10 @@ add_subdirectory(tools)
 option(LLDB_INCLUDE_TESTS "Generate build targets for the LLDB unit tests."
   ${LLVM_INCLUDE_TESTS})
 if(LLDB_INCLUDE_TESTS)
-  if (TARGET clang)
+  # TARGET clang works for a non-standalone build.
+  #
+  # FIXME: This can be avoided by importing the cmake configuration from swift.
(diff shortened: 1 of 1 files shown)


gottesmm:
@swift-ci test

gottesmm:
@fredriss @shahmishal @adrian-prantl @vedantk @dcci

gottesmm:
Just for good measure. I am going to do another linux build just to show that without cleaning things work.

gottesmm:
@swift-ci test linux platform

gottesmm:
... didn't go to the same builder ... I am pretty sure this is safe and should repair the PR build.

fredriss:
@vedantk can you double check this commit and see if something like this is needed upstream. This didn't merge cleanly to upstream-with-swift and I'll resolve the conflict by taking the upstream version which doesn't have the LLDB_BUILD_STANDALONE check.

vedantk:
@gottesmm I don't see any other uses of LLDB_BUILD_STANDALONE in swift/lldb/llvm. How are you using this?

PR #373, opened 2018-03-12
Fix assertion in ObjectFileELF

vedantk:
In D40616 I (mistakenly) assumed that logging an llvm::Error would clear
it. This of course is only true if logging is actually enabled.

This fixes the assertion by manually clearing the error, but it raises
the point of whether we need a special error-clearing logging primitive.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@322664 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 9436395663f0baa4a0e52c2af0c0dbc7aefe4c37)


vedantk:
@swift-ci test

vedantk:
Heads-up: @dcci, @fredriss, @adrian-prantl

PR #374, opened 2018-03-12
 Run a test that requires an in-tree debugserver only when we have one

vedantk:


vedantk:
@swift-ci test

jasonmolenda:
good change.  Shouldn't this be on llvm.org though?

jasonmolenda:
Ah, it is.  I didn't realize this was a cherry pick PR.

PR #378, opened 2018-03-13
Swift: Disable headermap include paths in Swift's ClangImporter by de…

adrian-prantl:
…fault.

This a big hammer. Based on the assumption that a well-formed
framework module will not need any headermaps to be imported, this
patch strips out all -I.*\.hmap ClangImporterOptions by default.

<rdar://problem/35732279>


adrian-prantl:
@swift-ci test

jrose-apple:
Please do not do this in LLDB. The whole point of putting search options in the swiftmodule instead of in regular debug info was so that the compiler would import things the same way the debugger does.

(Also, what about *non*-framework modules?)

adrian-prantl:
I need a way to enable/disable this feature from within LLDB. Not serializing headermaps in the first place is more elegant, but I want an escape hatch while we are actively reworking how header search works in LLDB.

> so that the compiler would import things the same way the debugger does

I think we already established that the compiler at the time where it generates a swiftmodule is not necessarily in the same state as the compiler when it loads a swiftmodule. The best example was the mixed-ObjC-Swift framework where the compiler building the swiftmodule is importing the Clang part of the framework module from the source directory, but any compiler importing the finished module will the copy of the headers relative to the framework directory. And in LLDB we are closer to that second, consumer use-case.

> (Also, what about *non-*framework modules?)

This is effectively phasing out support for modules that require a headermap to build. A non-framework module implies that the modulemap is maintained manually, and then it is reasonable to ask that all headers that appear in the module map are specified relative to the module.modulemap file.

adrian-prantl:
@swift-ci test

adrian-prantl:
@swift-ci test

jrose-apple:
> I think we already established that the compiler at the time where it generates a swiftmodule is not necessarily in the same state as the compiler when it loads a swiftmodule. The best example was the mixed-ObjC-Swift framework where the compiler building the swiftmodule is importing the Clang part of the framework module from the source directory, but any compiler importing the finished module will the copy of the headers relative to the framework directory. And in LLDB we are closer to that second, consumer use-case.

This wasn't about generating vs. loading; this was about how a framework is imported when building an app vs. debugging the app. And things got a *lot* better—or at least more consistent—when we started processing search paths the same way LLDB does.

> This is effectively phasing out support for modules that require a headermap to build. A non-framework module implies that the modulemap is maintained manually, and then it is reasonable to ask that all headers that appear in the module map are specified relative to the module.modulemap file.

I'm sorry, I still don't get it. How do you know there's a module map? I didn't see that in the diff.

adrian-prantl:
Arrg. This was not the window I meant to click on "merge" in. Sorry.

adrian-prantl:
revert in https://github.com/apple/swift-lldb/pull/392

PR #382, opened 2018-03-13
[ExpressionParser] Fix crash when evaluating invalid expresssions.

dcci:
Typical example, illformed comparisons (operator== where LHS and
RHS are not compatible). If a symbol matched `operator==` in any
of the object files lldb inserted a generic function declaration
in the ASTContext on which Sema operates. Maintaining the AST
context invariants is fairly tricky and sometimes resulted in
crashes inside clang (or assertions hit).

The real reason why this feature exists in the first place is
that of allowing users to do something like:
(lldb) call printf("patatino")

even if the debug informations for printf() is not available.
Eventually, we might reconsider this feature in its
entirety, but for now we can't remove it as it would break
a bunch of users. Instead, try to limit it to non-C++ symbols,
where getting the invariants right is hopefully easier.

Now you can't do in lldb anymore
(lldb) call _Zsomethingsomething(1,2,3)

but that doesn't seem to be such a big loss.

<rdar://problem/35645893>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@327356 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/lit/Expr/Inputs/basic.cpp
+++ b/lit/Expr/Inputs/basic.cpp
@@ -1,0 +1,12 @@
+class Patatino {
+private:
+  long tinky;
+
+public:
+  Patatino(long tinky) { this->tinky = tinky; }
+};
+
+int main(void) {
+  Patatino *a = new Patatino(26);
+  return 0;
+}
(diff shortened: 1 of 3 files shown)


dcci:
@swift-ci please test

dcci:
@vedantk `lang/swift/stepping/TestSwiftStepping.py` is failing, is this the test that you touched recently or am I mistaken here?

vedantk:
@dcci I made a swift change which didn't go into 4.2. Due to some racy-ness, the corresponding lldb change *did* go into 4.2. This should be resolved with https://github.com/apple/swift-lldb/pull/383.

vedantk:
@swift-ci test

dcci:
@swift-ci please test

PR #384, opened 2018-03-13
Temporarily disable TestCallStopAndContinue.test on Linux

vedantk:
This is a workaround to unblock the Swift Linux builders. It's unclear
why, but this test is sporadically failing.

rdar://38378404


vedantk:
@swift-ci test

PR #385, opened 2018-03-13
Fix CMake build

adrian-prantl:
rdar://38387074


adrian-prantl:
@swift-ci test

vedantk:
Thanks Adrian! My only nitpick would be that [0-2] should suffice to match dictionary 'indices', but this lgtm.

PR #388, opened 2018-03-14
Linux / cmake: Install the clang resource directory to the expected P…

adrian-prantl:
…OSIX path.

HostInfoPosix::ComputeClangDirectory() expects the clang resource
directory to be in lib/lldb/clang/$VERSION. This patch fixes the
Swift-specific CMake to install it there instead of in lib/lldb/clang/.

<rdar://problem/37354542>
https://bugs.swift.org/browse/SR-6954

(cherry picked from commit 0f090945431a91eb0a6a89a7c8dd85b0c0d5c1f8)


adrian-prantl:
@swift-ci test

PR #389, opened 2018-03-14
[test] Tighten up a dictionary test

vedantk:
This test isn't stringent enough.

It looks like DICT: key = 4 can match, then DICT: key = 3 can match, all before
"x[3] = 5" happens.

rdar://38459944

### DIFF
--- a/lit/SwiftREPL/Dict.test
+++ b/lit/SwiftREPL/Dict.test
@@ -38,10 +38,12 @@ var x = $R0
 
 x.updateValue(8, forKey : 4)
 x
+// DICT-LABEL: $R5: [Int : Int] = 4 key/value pairs
 // DICT: key = 4
 // DICT-NEXT: value = 8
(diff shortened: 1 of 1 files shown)


vedantk:
@swift-ci test

dcci:
I originally thought this was a bug in dictionary, but I guess you're right on this one, we just need another check-label

dcci:
Looks like @lorentey found a similar issue independently.

lorentey (approved):
This is exactly right! The stdlib now uses randomly seeded hashing, which flushes out these issues.

There is a way to disable random seeding through an environment variable; however, unless you have tests that would be inconvenient to rewrite, I suggest to leave it enabled. I'm happy to help either way!

dcci (approved):

vedantk:
Thanks all, this seems like the simplest option

PR #390, opened 2018-03-14
Skip pexpect-backed MI tests, and delete stale xfailed MI tests

vedantk:
Cherry-pick two commits from upstream which 1) skip pexpect-backed lldb-mi tests and 2) delete some stale lldb-mi tests.


vedantk:
The test suite health improves locally with this change. CI won't actually test this, so I'll just merge it.

PR #402, opened 2018-03-15
Fix a couple of more tests to not create files in the source tree

jimingham:
Summary:
These were not being flaky, but they're still making the tree dirty.

These tests were using lldbutil.append_to_process_working_directory to
derive the file path so I fix them by modifying the function to return
the build directory for local tests.

Technically, now the path returned by this function does not point to
the process working directory for local tests, but I think it makes
sense to keep the function name, as I think we should move towards
launching the process in the build directory (and I intend to change
this for the handful of inferiors that actually care about their PWD,
for example because they need to create files there).

Reviewers: davide, aprantl

Subscribers: lldb-commits

Differential Revision: https://reviews.llvm.org/D43506

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@325690 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 6d2b66f2740b2600b5afef8ba704245ce47a4343)


jimingham:
@swift-ci please test

PR #406, opened 2018-03-16
Fix a copy of a fixed length, possibly non-nul terminated, string

jasonmolenda:
into a std::string so we don't run off the end of the array when
there is no nul byte in ProcessElfCore::parseLinuxNotes.
Found with ASAN testing.

<rdar://problem/37134319>

Differential revision: https://reviews.llvm.org/D42828

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@324156 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 01cf28c211cd32c850bbe8f8110bf28d221fc54b)


jasonmolenda:
@swift-ci test

PR #411, opened 2018-03-16
[lldb-dotest] Set the --swift-compiler

JDevlieghere:
This sets the swift compiler in the lldb-dotest wrapper. We have the
path to swiftc avaialable in CMake, so it is just a matter of updating
the python wrapper.


JDevlieghere:
@swift-ci please test

dcci (approved):
This was actually easier than expected :)

PR #414, opened 2018-03-16
Cherry pick modules test fixes

vedantk:
This eliminates differences between upstream llvm.org and stable for this particular test.

rdar://38550145


adrian-prantl (approved):

vedantk:
Thanks! CI isn't set up to test this patch, so I'll just merge it. Testing looks good locally.

PR #419, opened 2018-03-19
Cherry pick rdar38011072 rdar38010986

fredriss:
<rdar://problem/38010986> If we have two classes with the same base name we get the wrong dynamic type
<rdar://problem/38011072> Breakpoints conditioned by thread name no longer function


fredriss:
@swift-ci please test

fredriss:
@swift-ci please test

PR #434, opened 2018-03-21
[lldb-dotest] Check that swif compiler and libs path exists.

JDevlieghere:
The @LLDB_PATH_TO_SWIFT_BUILD variable is not known at configuration
time so we need to read LLDB_SWIFTC and LLDB_SWIFT_LIBS which are (will
be) set by the build script.


JDevlieghere:
@swift-ci please test

JDevlieghere:
`LLDB_SWIFT_LIBS` is not yet set by the build script, but the os.path.exists call should prevent it from being set until the build script is updated.

PR #401, opened 2018-03-15
Run two additional REPL tests on Linux

vedantk:

PR #435, opened 2018-03-21
[Commands] Add coverage for language swift refcount.

dcci:
To the best of my knowledge this feature was not really tested.
As I'm going to do changes in this area (e.g. adding unowned
printing) I figured I could just add some coverage.

This turned out to be a nice way of experimenting writing
lldb tests in a single file (using lit). Maybe the syntax can
be cleaned up a bit, but I have to say I'm quite happy with the
ending results. Please let me know what you think.

This commit also adds the plumbing to invoke `swiftc` in lit tests,
which I didn't add before because all the swift lit tests were
using the repl.

In preparation for <rdar://30538363>.

### DIFF
--- a/lit/Swift/Refcount.swift
+++ b/lit/Swift/Refcount.swift
@@ -1,0 +1,31 @@
+// REQUIRES: darwin
+
+// RUN: rm -f %t.cmds %t
+// RUN: echo 'breakpoint set -p "break here"' > %t.cmds
+// RUN: echo 'run' >> %t.cmds
+
+// RUN: echo 'language swift refcount Arg0' >> %t.cmds
+// CHECK: refcount data: (strong = 0, weak = 1)
+
+// RUN: %swiftc %s -g -Onone -o %t && %lldb -b -s %t.cmds -- %t | FileCheck %s
+
+class Patatino {
+  var Foo : String
+
+  init(_ Foo : String) {
+    self.Foo = Foo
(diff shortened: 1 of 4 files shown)


dcci:
@swift-ci please test

lit/Swift/Refcount.swift
```diff
@@ -0,0 +1,31 @@
+// REQUIRES: darwin
```
vedantk: I found that some older lit tests did not really require Darwin (see: https://github.com/apple/swift-lldb/pull/401). Are we sure we need that here?

mikeash (approved):

lit/Swift/Refcount.swift
dcci: Not quite needed actually, just muscle memory and fat fingering. Let me fix it.

dcci:
@swift-ci please test

dcci:
@swift-ci please test

dcci:
After discussing this at length with Adrian I'm going to rewrite this as an inline test once I find the time.

dcci:
Will create a new one with the inline test.

vedantk:
It would be nice to have the run commands somewhat closer to the bits of code they're related to. Alternatively, maybe we could use 'cat <<EOF ... > %t.cmds', so the 'echo' commands don't clutter things up.

PR #443, opened 2018-03-23
Add a sanity check to the swift Native Hashed Container data formatter.

jimingham:
This checks that we can read the used items bitmask, and if we can't we
don't try to construct children for the container.  The variable can't
be valid if this is not initialized.

<rdar://problem/37475403>


jimingham:
@swift-ci please test

jimingham:
@swift-ci please test

PR #448, opened 2018-03-24
Add a testcase that uses a pure-ObjC main executable and two Swift

adrian-prantl:
dylibs with conflicting modules. This one neatly demonstrates the
difference between the Module- and the Target-SwiftASTContext.

<rdar://problem/38686915>


adrian-prantl:
@swift-ci test and merge

PR #451, opened 2018-03-24
Add a testcase where the expression evaluator fails with a module

adrian-prantl:
build error, but frame variable still works.

rdar://problem/38686915


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

PR #452, opened 2018-03-26
Make TestSwiftTupleTypes.py robust against type-sugar changes in the compiler

DougGregor:
Whether we print the `Builtin` part of the name of something from the builtin module isn't necessarily fixed, since these types aren't really supposed to be exposed to the user anyway.


DougGregor:
@swift-ci test and merge

PR #454, opened 2018-03-26
[Swift] Make this test more robust relaxing output checks.

dcci:
It was just too strict and matches what Doug did for another test.
While I'm here add it to `@swiftpr` so that we notice if this
regresses again during PR testing.


dcci:
@swift-ci test and merge

dcci:
@DougGregor can you please take a look and confirm this makes sense to you?

DougGregor (approved):
LGTM. There's no reason for us to qualify `Swift.AnyObject` in our output.

PR #457, opened 2018-03-26
Add a testcase where ClangImporter fails because of macro conflicts.

adrian-prantl:
rdar://problem/38686915


adrian-prantl:
@swift-ci test and merge

PR #459, opened 2018-03-26
Set the "debugger typealias" bit on synthesized typealiases.

DougGregor:
LLDB creates typealiases that involve archetypes that weren't introduced
in the enclosing context, which violates some of Swift's AST invariants.
Mark such typealiases as "debugger typealiases" so the Swift frontend
can tip-toe around them when needed.

We'll eventually want to rework the LLDB <-> Swift frontend
interaction here, but this allows us to tighten up AST verification in
the Swift frontend without breaking LLDB.


DougGregor:
This adopts the bit introduced in https://github.com/apple/swift/pull/15527.

DougGregor:

PR #461, opened 2018-03-26
[swift] Small string formatting support

milseman:
To only be merged in conjunction with https://github.com/apple/swift/pull/14755

### DIFF
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -242,6 +242,21 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
     if (objectAddr == LLDB_INVALID_ADDRESS)
       return false;
 
+    uint64_t topNibbleMask = 0xF000000000000000;
+    uint64_t smallUTF8TopNibble = 0xE000000000000000;
+
+    // Small UTF-8 string
+    if ((objectAddr & topNibbleMask) == smallUTF8TopNibble) {
+      uint64_t countMask = 0x0F00000000000000;
+      uint64_t count = (objectAddr & countMask) >> 56U;
+      assert(count <= 15 && "malformed small string");
+      uint64_t buffer[2] = {otherBits->GetValueAsUnsigned(0), objectAddr};
+      stream.Printf("\"");
+      stream.Write(&buffer, count);
+      stream.Printf("\"");
+      return true;
+    }
+
     bool isValue = objectAddr & (1ULL << 63);
     bool isCocoaOrSmall = objectAddr & (1ULL << 62);
     bool isOpaque = objectAddr & (1ULL << 61);
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -247,6 +247,17 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
 
     // Small UTF-8 string
     if ((objectAddr & topNibbleMask) == smallUTF8TopNibble) {
+      // Small UTF-8 strings store their contents directly in register. The
+      // high nibble of the most significant byte of the first word denotes
+      // that the string is small, the low nibble holds the count. The least-
+      // significant byte of the second word is the first UTF-8 code unit.
+      //
+      // {object, otherBits} =
+      //   { isSmallUTF8Nibble << 60 | count << 56 | c_14 << 48 | ... | c_8,
+      //     c_7 << 56 | c_6 << 48 | ... | c_0
+      //   }
+      //
+      //
       uint64_t countMask = 0x0F00000000000000;
       uint64_t count = (objectAddr & countMask) >> 56U;
       assert(count <= 15 && "malformed small string");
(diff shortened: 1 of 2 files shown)


vedantk (approved):
Thanks! Lgtm with a comment about the mask, as @dcci suggests.

milseman:
I added a comment with the small string layout.

milseman:
(Note to future self/merger: squash and merge)

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -242,6 +242,32 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
+      //     c_7 << 56 | c_6 << 48 | ... | c_0
+      //   }
+      //
+      //
+      uint64_t countMask = 0x0F00000000000000;
+      uint64_t count = (objectAddr & countMask) >> 56U;
+      assert(count <= 15 && "malformed small string");
+      uint64_t buffer[2] = {otherBits->GetValueAsUnsigned(0), objectAddr};
+      stream.Printf("\"");
+      stream.Write(&buffer, count);
```
vedantk: Because 0 isn't a terminator, you'll need to explicitly escape it here in order for it to render properly. I suggest using ReadBufferAndDumpToStreamOptions + DumpUTFBufferToStream to do this for you.

vedantk (changes requested):
I've left an inline comment which should help address the test failure here: https://ci.swift.org/job/swift-PR-osx/3925/consoleFull#376610391ba62d58e-7248-467b-91e0-c7508d5cf947

source/Plugins/Language/Swift/SwiftFormatters.cpp
milseman: So I spent some time trying to understand StringPrinter, figured out that was a waste of time and tried to copy the approach that's done for small Character. Here's what I wrote, however it gives me the same results as before with '\0' in the middle of a string:

```c++
      // Small UTF-8 strings store their contents directly in register. The
      // high nibble of the most significant byte of the first word denotes
      // that the string is small, the low nibble holds the count. The least-
      // significant byte of the second word is the first UTF-8 code unit.
      //
      // {object, otherBits} =
      //   { isSmallUTF8Nibble << 60 | count << 56 | c_14 << 48 | ... | c_8,
      //     c_7 << 56 | c_6 << 48 | ... | c_0
      //   }
      //
      //
      uint64_t countMask = 0x0F00000000000000;
      uint64_t count = (objectAddr & countMask) >> 56U;
      assert(count <= 15 && "malformed small string");
      uint64_t buffer[2] = {otherBits->GetValueAsUnsigned(0), objectAddr};

      DataExtractor data(buffer, count, process->GetByteOrder(),
                         process->GetAddressByteSize());

      StringPrinter::ReadBufferAndDumpToStreamOptions options;
      options.SetData(data)
          .SetEscapeNonPrintables(true)
          .SetPrefixToken(0)
          .SetQuote('"')
          .SetStream(&stream)
          .SetSourceSize(count)
          .SetLanguage(lldb::eLanguageTypeSwift);
      return StringPrinter::ReadBufferAndDumpToStream<
          StringPrinter::StringElementType::UTF8>(options);
```

Any ideas?

milseman: Argh, I figured out that I need `.SetBinaryZeroIsTerminator(false)`, which is completely confounding as what is the point of a length, and why doesn't small Character have to do this?

milseman: And why is StringPrinter highly unidiomatic and completely devoid of comments?

milseman:
Alright, so I added a test to hopefully catch these failures in the sane testing path.

vedantk (approved):

vedantk:
The added test is helpful, but will exercise stringForPrintObject() (from the swift standard library), not the String data formatter.

dcci:
The last versions of the patch looks good to me. Thanks for taking the time to fix the formatter properly.

milseman:
@vedantk the test's output varied as I changed the code. How is this not running the formatter?

edit: As in, if I remove the `.SetBinaryZeroIsTerminator(false)`, the test fails and the printed string stops before any interior nul character

vedantk:
@milseman Oh, got it. I assumed that the lldb repl used the expression evaluator to handle all of the expressions it gets, which would call stringForPrintObject. That might not be right -- @dcci would know better.

dcci:
I think the expression evaluation path goes through this formatter after several layers of indirection. As such, I think what you're seeing is expected (change your code/change output). lg.

dcci:
@swift-ci please test

jimingham:
There's lots of indirection in choosing and applying formatters to a particular ValueObject, but the formatters themselves (and stringForPrintObject) are just straight post-processing steps after expression evaluation is complete.  They aren't part of expression evaluation per se.  The expression evaluator just produces its result as an lldb_private::ValueObject.  Then the user of both the lldb SBAPI's and the command line can opt in or out of these post processing stages on the resultant ValueObject.  

The stringForPrintObject is part of the Swift "ObjectDescription", which you access by calling ValueObject::GetObjectDescription.  That only happens if you directly request it by:

1) By directly calling it through the SB API equivalent
2) By calling "frame variable" with the -O option
3) By calling "expression" with the -O option (po is an alias for "expression -O --)

The formatters are on by default whenever we print a Value (in "expression" and "frame var") or access its children.  You can opt out of this on the command line by passing the --raw options to either of these commands.

Jim

milseman:
@swift-ci please test

vedantk:
@milseman As apple/swift#14755 was merged, I can force-merge this for you if you'd prefer.

milseman:
@shahmishal @aschwaighofer  err, crap. The other branch merged so this should merge too to unblock CI. 

If failures start happening to NSString.test, please merge this early

milseman:
And please do a squash and merge if you do

PR #462, opened 2018-03-27
Reimplement GetDynamicTypeAndAddress_Class using the Remote AST library. 

DougGregor:
`GetDynamicTypeAndAddress_Class` hard-coded a fetch of an "isa" pointer and
used the fairly-heavy `GetMetadataPromise` abstraction to immediately
retrieve the type from the remote AST library. Replace both with direct
uses of remote ASTs.

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1433,58 +1433,47 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Class(
 
   AddressType address_type;
   lldb::addr_t class_metadata_ptr = in_value.GetPointerValue(&address_type);
-  if (auto objc_runtime = GetObjCRuntime()) {
-    if (objc_runtime->IsTaggedPointer(class_metadata_ptr)) {
-      Value::ValueType value_type;
-      return objc_runtime->GetDynamicTypeAndAddress(
-          in_value, use_dynamic, class_type_or_name, address, value_type,
-          /* allow_swift = */ true);
-    }
-  }
   if (class_metadata_ptr == LLDB_INVALID_ADDRESS || class_metadata_ptr == 0)
     return false;
   address.SetRawAddress(class_metadata_ptr);
 
-  size_t base_depth = BaseClassDepth(in_value);
-
-  lldb::addr_t class_instance_location;
-  if (in_value.IsBaseClass())
-    class_instance_location = in_value.GetPointerValue();
-  else
-    class_instance_location = in_value.GetValueAsUnsigned(LLDB_INVALID_ADDRESS);
-  if (class_instance_location == LLDB_INVALID_ADDRESS)
-    return false;
-  Status error;
-  lldb::addr_t class_metadata_location =
-      m_process->ReadPointerFromMemory(class_instance_location, error);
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1508,6 +1508,40 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Class(
...
+                                         *swift_ast_ctx->GetASTContext(),
+                                         GetMemoryReader());
+
+    swift::Type expected_type =
+      GetSwiftType(class_type_or_name.GetCompilerType());
+    swift::remote::RemoteAddress instance_address(in_value.GetPointerValue());
+    auto metadata_address =
+      remote_ast.getHeapMetadataForObject(instance_address);
+    if (!metadata_address) {
+      llvm::errs() << "Remote AST failed to get metadata address for type "
```
jimingham: We don't use llvm::errs.  This information should go into the types log.  Do something like:

      Log *log(GetLogIfAllCategoriesSet(LIBLLDB_LOG_TYPES));
      if (log)
        log->Printf(...);

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1508,6 +1508,40 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Class(
...
+                   << expected_type.getString() << "\n";
+      llvm_unreachable(
```
jimingham: Don't use llvm_unreachable here.  This operation, could fail, for instance, because the stack or heap was corrupted, and you certainly wouldn't want lldb aborting in that instance.  Finding dynamic types is an operation it's okay to fail at.  Just emit something useful to the types log and return false.

jimingham (changes requested):
It isn't appropriate to abort because you couldn't reconstruct a dynamic type.

source/Target/SwiftLanguageRuntime.cpp
DougGregor: Okay! I'll use that if I'm going to put in any logging that's meant to be merged.

source/Target/SwiftLanguageRuntime.cpp
DougGregor: Sorry it wasn't clear from my commit message/Do Not Merge designation; I wanted a hard failure here because it would indicate a case where LLDB knew how to resolve something that Remote AST couldn't, which would be a bug we need to fix in RemoteAST.

DougGregor:
A clean test here implies that we can replace the implementation of `GetDynamicTypeAndAddress_Class` with a call to RemoteAST's `getHeapMetadataForObject()`. I'll rework this PR to do that.

DougGregor:
@swift-ci please test

DougGregor:
@jimingham this is what the actual change looks like, building on top of https://github.com/apple/swift-lldb/pull/465.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@jimingham what do you think?

DougGregor:
@swift-ci please test

jimingham:
Ah, okay.  No problem.  So long as this was only a test...

PR #465, opened 2018-03-27
[Swift] Simplify the extraction of the offset of a member variable.

DougGregor:
Eliminate `MemberVariableOffsetResolver`, which introduced a significant
amount of overhead to allow one to delay the offset computation for a
member variable, but its only use was immediate. Instead of creating a
new Swift remote AST context for each (Swift AST context, Type) pair,
keep one Swift remote AST context per Swift AST context and use that
to immediately query the remote AST context for the member offset.

While here, replace a handcoded fetch of the "isa" pointer of a class
instance with a call to `RemoteASTContext::getHeapMetadataForObject()`.

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -47,10 +47,6 @@ public:
   class MetadataPromise;
   typedef std::shared_ptr<MetadataPromise> MetadataPromiseSP;
 
-  class MemberVariableOffsetResolver;
-  typedef std::shared_ptr<MemberVariableOffsetResolver>
-      MemberVariableOffsetResolverSP;
-
   //------------------------------------------------------------------
   // Static Functions
   //------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

dcci:
whoah.. thanks. This is much easier now.

jimingham:
The other thing the MemberVariableOffsetResolver did was provide a map of type_system+type+ivar-name -> offset, which your patch eliminates.

You don't actually see that cache getting used because we also cache member offsets in the SwiftAST Context (see SwiftASTContext::GetCachedMemberInfo) but at least for the lldb_private::Module SwiftASTContext, that is actually wrong.  The lldb_private Modules are shared among all the targets & debugging sessions in the lldb process, and you might end up debugging two programs that - due to differences in other libraries they load - end up with different ivar offsets.  So you have to keep the ivar offsets in a cache that is process specific, not in one that is module specific.

So I think we will end up needing this map back.

PR #470, opened 2018-03-27
[SwiftASTContext] Remove dead code. NFCI.

dcci:


dcci:

jimingham:
All the FindXType(s)OrDecl(s) functions have a FindType equivalent.  It seems a little odd to remove the "Type" version from just one of them.

dcci:
As I discover more dead code, I'm going to remove it.

dcci:
But as a general practice, I don't think we should leave dead code in tree just for the sake of internal API symmetry.

PR #474, opened 2018-03-28
Revert "Reimplement GetDynamicTypeAndAddress_Class using the Remote AST library. "

DougGregor:
Reverts apple/swift-lldb#462. This broke a test that we'd somehow missed pre-merge.

### DIFF
--- a/lit/SwiftREPL/NSString.test
+++ b/lit/SwiftREPL/NSString.test
@@ -6,7 +6,7 @@
 import Foundation
 var aString = "patatino" as NSString
 // NSSTRING: Welcome to Swift
-// NSSTRING-NEXT: aString: NSString = "patatino"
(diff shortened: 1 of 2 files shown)


DougGregor:

DougGregor:

dcci:
Which test did this break? If it's not part of swiftpr, we should consider adding it to the set.

DougGregor:
`TestSwiftDictionaryNSObjectAnyObject`

jimingham:
There was some code in the section you deleted that handled types backed by ObjC tagged pointers specially (SwiftLanguageRuntime.cpp:1436 and following.  The failing test is testing that we can print a dictionary filled with NSNumbers, which are usually tagged pointers.  Maybe that's what is missing from the RemoteAST side.

DougGregor:
Good point! I’ll try it out

PR #475, opened 2018-03-28
[SwiftTestSuite] A couple of fixes/improvements for our testing.

dcci:
1) Fixup the missing `import` that caused the trivial ABI test
to fail. llvm.org has the `import` bit so it's probably the result
of a merge issue/conflict.

2) Add a dictionary test that Doug found to `swiftpr` category.
It's sensitive to swift frontend changes, so it amkes sense to
adding it here.

PR #476, opened 2018-03-28
Explicitly import subprocess

jimingham:
For some reason on one of our bots subprocess wasn't already
imported. Do so explicitly.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@328715 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/packages/Python/lldbsuite/test/lang/cpp/trivial_abi/TestTrivialABI.py
+++ b/packages/Python/lldbsuite/test/lang/cpp/trivial_abi/TestTrivialABI.py
@@ -9,6 +9,7 @@ import os
 import time
 import re
 import lldb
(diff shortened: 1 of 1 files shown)


jimingham:

dcci:
please abort this one, it's been fixed already by https://github.com/apple/swift-lldb/pull/475 
(see your inbox for details)

dcci:
Also, this particular fix is wrong. `decorators` should import subprocess and not this very specific test. The way this is fixed creates an unneeded divergence with llvm.org (which already had the import subprocess bits in the right place)

jimingham:
Right you are.  Working too quickly...

PR #478, opened 2018-03-28
[cmake] Set the default test compiler to the clang built in-tree

vedantk:
The cmake build tries to make sure the `clang` target is available
before setting the test compiler to a clang binary built in-tree.

For reasons we don't understand, the `clang` target is not always
available. This is despite running the exact same build-script and
cmake invocations on the same bot.

As a workaround, in swift-lldb, this change unconditionally makes the
default test compiler be the clang that's built in-tree.

rdar://38935299


vedantk:
@swift-ci test

gottesmm:
LGTM. My one suggestion would be to put in a warning if the build would fail.

gottesmm:
*would have failed.

gottesmm:
the reason I am suggesting that is it could be useful for your explorations.

vedantk:
I added some logging for when the clang target goes missing per @gottesmm's suggestion.

vedantk:
@swift-ci test

gottesmm (approved):
LGTM!

vedantk:
@swift-ci test

PR #479, opened 2018-03-29
Implement Swift tuple type queries via AST and remote AST APIs

DougGregor:
For tuple types, re-implement the static queries `GetNumChildren`, `GetNumFields`, 
`GetIndexOfChildMemberWithName` and `GetIndexOfChildWithName` for
 tuple types using AST-level information.

For tuple layout queries, return "invalid offset" when asked for the offset
without a corresponding execution context (e.g., in `GetFieldAtIndex`)
and use the RemoteAST-based interface (`GetMemberVariableOffset`) when
asked the question with a corresponding execution context (e.g., in
`GetChildCompilerTypeAtIndex`). This gets LLDB out of the business of
performing type layout for tuples.

This also eliminates the caching of tuple type information within
`SwiftASTContext`, where it doesn't belong; all of the information we
need is already in the AST or cached within `SwiftLanguageRuntime`.


DougGregor:
@swift-ci please test

PR #480, opened 2018-03-29
Prevent double release of mach ports

fredriss:
Summary:
When a MIG routine returns KERN_FAILURE, the demux function will release any OOL resources like ports. In this case, task_port and thread_port will be released twice, potentially resulting in use after free of the ports.

I don't think we can test this in any useful way
rdar://problem/37331387

Reviewers: jasonmolenda

Subscribers: lldb-commits

Differential Revision: https://reviews.llvm.org/D45011

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@328761 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 8f022f17fe6a1579c1ec53c0f3f369fdcbd1a71b)


fredriss:

PR #483, opened 2018-03-29
Reimplement GetDynamicTypeAndAddress_Class using the Remote AST library.

DougGregor:
GetDynamicTypeAndAddress_Class hard-coded a fetch of an "isa" pointer and
used the fairly-heavy "GetMetadataPromise" abstraction to immediately
retrieve the type from the remote AST library. Replace both with direct
uses of remote ASTs.

Leave in the special case for tagged pointers; remote AST may need
some special support for them before we can remove it from LLDB.


DougGregor:

PR #485, opened 2018-03-30
 [Swift AST Context] Eliminate the member info cache.

DougGregor:
This cache is now unused, having been replaced by three related changes:

1) Structural types now compute the requested information on-demand. (Start doing
so for existential types).
2) Struct and class types only cache their stored properties (at the
AST context level), to provide O(1) index-to-stored-property access.
3) Stored property offsets are cached within the language runtime, so
they are target-specific.

Fixes rdar://problem/38926124.

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -45,7 +45,6 @@ class TypeInfo;
 }
 }
 
-struct CachedMemberInfo;
 class DWARFASTParser;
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

jimingham (approved):
Removing the complexity of these caches seems good to me.  The translation from setting cache elements to getting them on demand looks faithful.  This seems fine, thanks for doing it!

PR #487, opened 2018-03-31
 Implement `SwiftASTContext::GetFieldAtIndex` using `GetChildCompilerTypeAtIndex`

DougGregor:
These two methods did largely the same thing. `GetFieldAtIndex()`
generally cannot look at the execution context (which means it can't
really give offsets for anything), but `GetChildCompilerTypeAtIndex()`
already handles a NULL execution context and/or instance. Replace the
former with a call to the latter, eliminating some code duplication.

There's also a little cleanup here to stop relying on the hard-coded string
`instance_type` to find the instance within an existential; we have enough
data to retrieve it by index.

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -503,6 +503,10 @@ public:
     bool m_is_objc;
     bool m_is_anyobject;
     bool m_is_errortype;
+
+    /// The member index for the error value within an error
+    /// existential.
+    static constexpr unsigned error_instance_index = 0;
   };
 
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

dcci:
@DougGregor unless you have objections, I'm going to steal this from you. It's very close, and I would really like to get this through the finish line in the next week or so, if possible.

DougGregor:
Please go ahead and steal it! I don't expect to have time to nudge it along.

dcci:
This will be superseded by other commits.

PR #488, opened 2018-03-31
 Improve inspection of Swift existential types.

DougGregor:
Provide better type information for the children of existential types, e.g.,
* If there is a superclass bound, use that type for the instance object
  within the existential
* When printing protocol witness table entries, include the protocol name
  with each entry
* When the child is an instance type, use Any.Type (rather than
  RawPointer) for the type.

We get output more like this now:

```
(a.PointSuperclass & PointUtils) $R4 = {
  (a.PointSuperclass) instance = 0x00007fc618f06780 {
    (Swift.Int) superData = {
      (Int64) _value = 17
    }
  }
  (Builtin.RawPointer) witness_table_PointUtils = 0x0000000106d92168
}
```

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -503,6 +503,10 @@ public:
     bool m_is_objc;
     bool m_is_anyobject;
     bool m_is_errortype;
+
+    /// The member index for the error value within an error
+    /// existential.
+    static constexpr unsigned error_instance_index = 0;
   };
 
   static bool GetProtocolTypeInfo(const CompilerType &type,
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

jimingham:
The only thing that gives me a little pause about this is that by replacing the lookup of the instance type field using "instance_type" name with a lookup by index a m_num_payload_words (SwiftLanguageRuntime.cpp:1512) you are putting the knowledge of where this field is in two places, whereas before it was only in one.  If this is something you are not likely to change, that doesn't so much bother me.

Other than that,  this looks fine.

DougGregor:
Good point; I've moved the knowledge into `ProtocolInfo`.

DougGregor:

DougGregor:

DougGregor:

DougGregor:

PR #489, opened 2018-04-03
 Adapt LLDB to swift compiler changes in inout type handling.

adrian-prantl:
The swift compiler no longer mangled inout types and instead uses
DWARF expressions to express indirectness.

<rdar://problem/39125153>


adrian-prantl:
https://github.com/apple/swift/pull/15694
@swift-ci test

jimingham:
Should we also pull out the code that was detecting the inout or @lvalue and doing the extra dereference?  That should be dead code now, and will only do harm if we trip across it later on.

adrian-prantl:
https://github.com/apple/swift/pull/15694
@swift-ci test

adrian-prantl:
https://github.com/apple/swift/pull/15694
@swift-ci test

PR #490, opened 2018-04-03
Create one SwiftASTContext per Module if the scratch context fails

adrian-prantl:
because of irreconcilable module import errors.

When a Swift program contains several dylibs with conflicting
ClangImporter options building the common scratch SwiftASTContext
fails and the expression evaluator is unusable. This patch detects the
situation where an expression fails because of a module import error
and then switches into a mode where one scratch context per
lldb::Module is created, using only the ClangImporter flags for the
module which ideally should have been tested by the Swift compiler
that built that module. This also bifurcates the persistent expression
state, such that each lldb::Module has its own set of return
variables, but it is still a strict improvement over not having access
to the expression evaluator at all.

rdar://problem/38686915

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
@@ -125,7 +125,8 @@ void SwiftUserExpression::ScanContext(ExecutionContext &exe_ctx, Status &err) {
     // we need to make sure the Target's SwiftASTContext has been setup BEFORE
     // we do any Swift name lookups
     if (m_target) {
-      SwiftASTContext *swift_ast_ctx = m_target->GetScratchSwiftASTContext(err);
+      SwiftASTContext *swift_ast_ctx = m_target->GetScratchSwiftASTContext(
+          err, *frame);
       if (!swift_ast_ctx) {
         if (log)
           log->Printf("  [SUE::SC] NULL Swift AST Context");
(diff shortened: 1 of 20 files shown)


adrian-prantl:
@jimingham Could you please review this change? If we decide that we want to move forward with this, we should probably change the upstream interfaces to also take an ExecutionContextScope, so we can get rid of the silly `if (language == Swift)` conditions.

adrian-prantl:
@swift-ci test

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
```diff
@@ -448,17 +443,20 @@ bool SwiftUserExpression::Parse(DiagnosticManager &diagnostic_manager,
...
+      llvm::make_unique<SwiftExpressionParser>(exe_scope, *this, m_options);
 
-  unsigned num_errors = parser->Parse(
-      diagnostic_manager, first_body_line, 
+  unsigned error_code = m_parser->Parse(
+      diagnostic_manager, first_body_line,
       first_body_line + source_code->GetNumBodyLines(), line_offset);
 
-  if (num_errors) {
+  if (error_code == 2) {
```
jimingham, line 453: I don't understand why these changes are in this patch, and I can't find a corresponding change to SwiftExpressionParser::Parse that would motivate them.

jimingham (commented):
There was one change related to FixIts in there I didn't understand.  

Also, I'm not sure why you removed the extra_options from making the ScratchASTContext.  That gets used in the REPL so you can pass Framework paths if you want to import some module you've built into the REPL.  Maybe I missed how you worked around that?

Other than that, the method looks fine.  It will be much cleaner if you just always pass in an execution context when you go to get the ScratchASTContext.  If we're going to do this for real we should push those changes upstream.

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
adrian-prantl: This is the mechanism used to signal that we should tear down the scratch AST context build a per-module one and retry the expression, using the retry mechanism that exists for fix-its.

PR #491, opened 2018-04-03
[SwiftLanguageRuntime] Fix printing of __NSCFString.

dcci:
The internal store went from being a Swift wrapped string:
Swift._NSContiguousString to just a straight up __NSCFString.

This was broken after the recent String changes, but we
didn't catch this because this test wasn't run as part
of the PR workflow. Eventually we won't need to care about
these details in lldb as everything will be hidden inside
RemoteAST, but for now, we need to fix this case.

<rdar://problem/39006341>

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1844,6 +1844,10 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Protocol(
     if (!type.IsValid())
       return false;
 
+    lldb::addr_t class_metadata_ptr = payload0_sp->GetAddressOf();
+    if (class_metadata_ptr && class_metadata_ptr != LLDB_INVALID_ADDRESS)
+      address.SetRawAddress(class_metadata_ptr);
+
     class_type_or_name.SetCompilerType(type.GetPointerType());
     return class_type_or_name.GetCompilerType().IsValid();
   }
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci test and merge

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1844,11 +1844,6 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Protocol(
     if (!type.IsValid())
       return false;
 
-    lldb::addr_t class_metadata_ptr = payload0_sp->GetAddressOf();
-    if (class_metadata_ptr == LLDB_INVALID_ADDRESS || class_metadata_ptr == 0)
-      return false;
-    address.SetRawAddress(class_metadata_ptr);
```
DougGregor: `address` is passed by-reference. Do we want to still `SetRawAddress` when we get a `class_metadata_ptr`, i.e.,

```
if (class_metadata_ptr && class_metadata_ptr != LLDB_INVALID_ADDRESS)
  address.SetRawAddress(class_metadata_ptr);
```

eliminating the early exit but not the successful case?

dcci: Most definitely.

dcci:
@swift-ci test and merge

DougGregor (approved):
Thank you!

dcci:
@swift-ci test and merge

PR #492, opened 2018-04-03
[lldbinline] Reenable inline tests as part of PR testing.

dcci:
We should have more coverage now. We still have one XFAIL,
so this isn't really perfect, but it's much much better than
not running any of these tests at all (hint: they recently
broke).


dcci:

vedantk:
Testing will fail unless https://github.com/apple/swift-lldb/pull/491 is merged first.

vedantk (approved):

dcci:
@swift-ci test and merge

PR #495, opened 2018-04-04
Remove assertions on invalid input

adrian-prantl:
As I'm working on testcases that exercise LLDB's behavior with less than optimal input, I noticed that the following use of lldbassert() seems to be wrong. An assertion should only be used when an internal invariant is violated, but these assertions trigger when the SwiftASTContext has developed fatal errors, which can easily happen in the presence of incomplete debug information.


jimingham (approved):
The original idea was that the test suite should never have a case where we couldn't reconstruct debug info, so we put this in to catch these failures right away.  OTOH, it would have to be a pretty trivial test to fail to reconstruct debug info and not fail shortly thereafter.  And since you are adding tests on how we handle this failure, it makes sense to remove the lldbassert.

adrian-prantl:
@swift-ci test and merge

PR #497, opened 2018-04-04
[SwiftRuntime] Add a test for `language swift refcount`.

dcci:
This was previously untested. As I'm going to make significant
changes here, I figured this should've been exercised somehow.
It's currently broken (prints a strong refcount of zero for
a live object), but hopefully my next commit will fix this.

<rdar://problem/30538363>


dcci:
@swift-ci test and merge

PR #499, opened 2018-04-05
Support template template parameters

fredriss:
Summary:
We would fail to resolve (and thus display the value of) any
templated type which contained a template template argument even
though we don't really use template arguments.

This patch adds minimal support for template template arguments,
but I doubt we need any more than that.

rdar://38464962

Reviewers: clayborg, jingham

Subscribers: JDevlieghere, lldb-commits

Differential Revision: https://reviews.llvm.org/D44613

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@328984 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 4f44b1c6ee7a630e852ec78c6af5d34dae687ff8)


fredriss:

PR #503, opened 2018-04-05
Improve support for debugging Swift code built on a different machine.

adrian-prantl:
The patch does two things:

- If present, applies the target.source-map setting to the
  ClangImporter options of the scratch AST context.

- If present, applies the DBGSourcePathRemapping dictionary from the
  .dSYM bundle to the per-lldb::Module SwiftASTContext.

This allows the ClangImporter to find Clang module maps in a new
location on the machine that LLDB is running on.

This also fixes an oversight in the xcode project file uncovered by
this change: it now sets the libxml2 -D definition for the custom
swift targets, too.

rdar://problem/36032144
(cherry picked from commit 677798b0781387b5f6be39a7d80245b4cce9cacb)

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -865,6 +865,10 @@ protected:
   SwiftEnumDescriptor *GetCachedEnumInfo(void *type);
 
   friend class CompilerType;
+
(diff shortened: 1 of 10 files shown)


adrian-prantl:
@swift-ci test and merge

PR #504, opened 2018-04-05
[dotest] --skip-category should append and not override.

dcci:
<rdar://problem/38566150>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@329358 91177308-0d34-0410-b5e6-96231b3b80d8


dcci:
@swift-ci test and merge

dcci:
@fredriss we should be able to disable the watchpoint tests after this on Linux

PR #508, opened 2018-04-06
[lldb-server] Set a more generous timeout when testing gdbremote.

dcci:
One of our downstream bot is struggling under load,  but this
value should be enough for everyone.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@329426 91177308-0d34-0410-b5e6-96231b3b80d8


dcci:
@swift-ci test and merge

dcci:
This should hopefully calm down the swift Linux bot.

PR #509, opened 2018-04-06
Disable a test on linux till I figure out how to get it to run.

jimingham:
This test isn't finding a .so that we've built.  Disable it on Linux to get the bots green while I figure out what is up.


jimingham:

dcci:
Please refrain from doing this. We're debugging it now.

PR #513, opened 2018-04-06
Split off the float and double returns into a separate test.

jimingham:
This test was getting skipped for not-in-tree debugserver.  That was an
issue with floating point handling.  So I split off the floating point tests
so the rest of the tests could run with the out-of-tree debugserver.


jimingham:

PR #514, opened 2018-04-09
[Swift] Remove unneccessary call to performSILLinking()

slavapestov:


slavapestov:
@swift-ci Please test

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -1857,8 +1857,6 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
     log->PutCString(s.c_str());
   }
 
-  swift::performSILLinking(sil_module.get());
```
dcci, line 1860: I was actually looking at this code over the weekend and wondering whether we were trying to link here. I assume it's been cargo culted from something else in the compiler.

slavapestov: Your intuition is correct. It may have been necessary at some point in the past but is no longer so.

PR #150, opened 2017-02-17
Adapt lldb to the new swift weak reference representation.

jimingham:
This is adapting lldb to the new swift weak reference representation.  This pull request depends on changes from Greg Parker, so he's going to make the changes on his side and then do a joint test.

<rdar://problem/29765836>

PR #515, opened 2018-04-10
[SwiftLanguageRuntime] Fix language refcount.

dcci:
This commit has a bunch of improvements:
1) It moves the logic from wrong ad-hoc calculations in lldb to
an API call to the swift stdlib to retrieve the refcount of
objects.
2) It refactors the path to exit early for improved readability,
while also adding some comments.
3) It prints the `unowned` refcount for objects, previously
absent.

<rdar://problem/30538363>
<rdar://problem/30538399>

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -343,9 +343,6 @@ public:
 
   ConstString GetStandardLibraryBaseName();
 
-  virtual bool GetReferenceCounts(ValueObject &valobj, size_t &strong,
-                                  size_t &weak);
-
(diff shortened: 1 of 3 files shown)


dcci:
@swift-ci test and merge

dcci:
Thanks to @jimingham for teaching me about `GetSyntheticValue()`, btw.

dcci:
@slavapestov this fixes the other half of the problem you reported in https://github.com/apple/swift-lldb/pull/150#issuecomment-338533479 . Greg fixed the object printing representation, but the command was broken. Hopefully not having to deal with hand-made calculations in lldb would ensure us to not having to touch this again. Well, modulo bugs, of course.

PR #521, opened 2018-04-11
Accelerate simple expressions that are just local variable printing

jimingham:
<rdar://problem/35542636>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/expression/simple/TestSimpleExpressions.py
+++ b/packages/Python/lldbsuite/test/lang/swift/expression/simple/TestSimpleExpressions.py
@@ -32,6 +32,7 @@ class TestSimpleSwiftExpressions(TestBase):
         self.do_test()
 
     @decorators.swiftTest
+    @decorators.add_test_categories(["swiftpr"])
     def test_frame_var_accelerator(self):
         self.build()
         self.do_locals_test()
--- a/packages/Python/lldbsuite/test/lang/swift/expression/simple/TestSimpleExpressions.py
+++ b/packages/Python/lldbsuite/test/lang/swift/expression/simple/TestSimpleExpressions.py
@@ -31,14 +31,31 @@ class TestSimpleSwiftExpressions(TestBase):
         self.build()
         self.do_test()
 
+    @decorators.swiftTest
+    def test_frame_var_accelerator(self):
+        self.build()
+        self.do_locals_test()
+
     def setUp(self):
         TestBase.setUp(self)
         self.main_source = "main.swift"
         self.main_source_spec = lldb.SBFileSpec(self.main_source)
 
+    def do_locals_test(self):
+        """Test simple swift expressions"""
+        (target, process, self.thread, breakpoint) = lldbutil.run_to_source_breakpoint(self, 
+                                                                                        "Set breakpoint here", 
+                                                                                        self.main_source_spec)
+        self.frame = self.thread.frames[0]
+        self.assertTrue(self.frame, "Frame 0 is valid.")
+        self.expect("expression --allow-jit 0 -- same_five", substrs=["Bool", "true"])
+        self.expect("expression --allow-jit 0 -- b_struct.b_int", error=True)
+        
+
     def check_expression(self, expression, expected_result, use_summary=True):
         value = self.frame.EvaluateExpression(expression)
-        self.assertTrue(value.IsValid(), expression + "returned a valid value")
+        self.assertTrue(value.IsValid(), expression + "returned a invalid value")
+        
         if self.TraceOn():
             print value.GetSummary()
             print value.GetValue()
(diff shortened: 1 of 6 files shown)


jimingham:
@swift-ci please test

source/Commands/CommandObjectExpression.cpp
```diff
@@ -398,9 +398,50 @@ bool CommandObjectExpression::EvaluateExpression(const char *expr,
...
+    // simple variable access is quite slow.  Accelerate that by first seeing
+    // if the whole expression is a simple identifier, and if so look it up
+    // as a local variable.
+    // Note, Swift's type resolution rules are complex enough that even figuring
+    // out whether the dynamic and static versions of foo.bar are the same is
+    // tricky, so we only handle single identifiers.
+    // Also, we don't do this if we are using the REPL, top level code, 
+    // or if the PlaygroundTransform is enabled.
+    ExpressionResults success = eExpressionSetupError;
+    if (target->GetUseFrameVarToAccelerateExpr()
```
dcci: Can you split the fast path in a different function (and clang-format for consistency) ?

packages/Python/lldbsuite/test/lang/swift/expression/simple/TestSimpleExpressions.py
```diff
@@ -31,14 +31,31 @@ def test_simple_swift_expressions(self):
         self.build()
         self.do_test()
 
+    @decorators.swiftTest
```
dcci, line 34: Can we add this to @swiftpr ?

dcci (commented):
The general idea seems fine (as you already discussed this several times with us), some minors comment inline. Thanks!

dcci:
(also, if you move this to a separate function you can probably leverage early returns to reduce indentations/make the code slightly more readable).

packages/Python/lldbsuite/test/lang/swift/expression/simple/TestSimpleExpressions.py
```diff
@@ -31,14 +31,31 @@ def test_simple_swift_expressions(self):
...
         self.main_source_spec = lldb.SBFileSpec(self.main_source)
 
+    def do_locals_test(self):
+        """Test simple swift expressions"""
+        (target, process, self.thread, breakpoint) = lldbutil.run_to_source_breakpoint(self, 
+                                                                                        "Set breakpoint here", 
+                                                                                        self.main_source_spec)
+        self.frame = self.thread.frames[0]
+        self.assertTrue(self.frame, "Frame 0 is valid.")
+        self.expect("expression --allow-jit 0 -- same_five", substrs=["Bool", "true"])
```
adrian-prantl: Could you please add a comment explaining why --allow-jit is used here? It's not obvious to me and perhaps others can learn something, too :-)

source/Commands/CommandObjectExpression.cpp
```diff
@@ -398,9 +398,50 @@ bool CommandObjectExpression::EvaluateExpression(const char *expr,
...
+                                                          eDynamicCanRunTarget,
+                                                          0,
+                                                          var_sp,
+                                                          error);
+            if(error.Success()
+               && result_valobj_sp
+               && result_valobj_sp->GetError().Success())
+            {
+              success = eExpressionCompleted;
+              result_valobj_sp = result_valobj_sp->Persist();
```
adrian-prantl: Would it make sense to early-exit on success, perhaps by factoring out this and the regular evaluation into a separate function? I always find the control flow in these multi-page functions incredibly difficult to follow otherwise.

source/Commands/CommandObjectExpression.cpp
```diff
@@ -398,9 +398,50 @@ bool CommandObjectExpression::EvaluateExpression(const char *expr,
...
+        && !target->UseScratchTypesystemPerModule()
+        && frame
+        && (options.GetLanguage() == eLanguageTypeSwift
+                  || (options.GetLanguage() == eLanguageTypeUnknown
+                      && frame->GetLanguage() == eLanguageTypeSwift))
+              && options.GetExecutionPolicy() != eExecutionPolicyTopLevel
+              && !options.GetREPLEnabled()
+              && !options.GetPlaygroundTransformEnabled())
+    {
+        if (swift::Lexer::isIdentifier(expr))
```
adrian-prantl: Eventually this should probably be a TypeSystem(or Runtime?)::isSimpleExpression() function rather than hardcoding the Swift functionality here?

adrian-prantl (commented):
I only have a few superficial/stylistic remarks, but otherwise this looks fine.

jimingham:
@swift-ci please test

source/Commands/CommandObjectExpression.cpp
jimingham: This change seems to me more like a workaround for the fact that we don't have a reasonable IR interpreter for Swift.  For C languages I don't think this is actually going to save that much time over IR interpreting the trivial expression.  So I'm not sure I want to enshrine this in the type system.  I'd rather solve the performance problems within the expression parser.

source/Commands/CommandObjectExpression.cpp
jimingham: You can't early exit on success, since you need to do all the work from line 455 following in both cases.  The regular evaluation is two lines of code, I don't think it would be clearer to make that a separate function.  But I'll make the swift local lookup a separate function.

dcci: I guess what we meant was:

```
Status fastPath(...) {
  if (cond1 || cond2 || cond3)
    return NotExecuted
  Evaluate
  return evaluate status
}
```

I really find nested control flow easy to get wrong and super-hard to read.

jimingham: Yes, I'll do that when I extract this code into a separate function.

jimingham:
Addressed review comments.  I'll try the test again, since the last attempt seems to have stalled.

jimingham:
@swift-ci please test

jimingham:
@swift-ci please test

dcci (approved):
Looks much better now. Thanks!

dcci:
@jimingham I thought you merged some variation of this. Am I correct?

jimingham:
No, it wasn't merged anywhere.  There were still some subtle differences between what you got from frame var and from the expression parser, and as the range of things we could accelerate narrowed, it didn't seem worth the complexity.

Jim

dcci:
I see, thanks!

PR #523, opened 2018-04-12
Don't assume backing thread shares protocol ID.

JDevlieghere:
When we're dealing with virtual (memory) threads created by the OS
plugins, there's no guarantee that the real thread and the backing
thread share a protocol ID. Instead, we should iterate over the memory
threads to find the virtual thread that is backed by the current real
thread.

Differential revision: https://reviews.llvm.org/D45497

rdar://36485830

### DIFF
--- a/include/lldb/Target/ThreadList.h
+++ b/include/lldb/Target/ThreadList.h
@@ -102,6 +102,8 @@ public:
 
   lldb::ThreadSP GetThreadSPForThreadPtr(Thread *thread_ptr);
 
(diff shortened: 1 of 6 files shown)


JDevlieghere:
@swift-ci please test

JDevlieghere:
This was reverted upstream; I'll update the PR once it's re-landed.

JDevlieghere:
@swift-ci please test

JDevlieghere:
@swift-ci please test

PR #527, opened 2018-04-13
 Update the SummaryProvider for Data.

adrian-prantl:
Update the SummaryProvider for Data.  …
The current implementation of Data makes it much easier to get at the
length of the underlying storage.

This uncovered an unrelated bug, so I'm only re-enabling part of the
original test.

<rdar://problem/36649559>


adrian-prantl:
@swift-ci test and merge

PR #533, opened 2018-04-17
Change PlatformPosix::DoLoadImage to use a UtilityFunction.

jimingham:
That way we won't have to compile a new expression every time we want
dlopen a library.

<rdar://problem/32626584> 

Differential Revision: https://reviews.llvm.org/D45703

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@330214 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #536, opened 2018-04-18
Create one SwiftASTContext per Moudle if the scratch context fails

adrian-prantl:
because of irreconcilable module import errors.

When a Swift program contains several dylibs with conflicting
ClangImporter options building the common scratch SwiftASTContext
fails and the expression evaluator is unusable. This patch detects the
situation where an expression fails because of a module import error
and then switches into a mode where one scratch context per
lldb::Module is created, using only the ClangImporter flags for the
module which ideally should have been tested by the Swift compiler
that built that module. This also bifurcates the persistent expression
state, such that each lldb::Module has its own set of return
variables, but it is still a strict improvement over not having access
to the expression evaluator at all.

rdar://problem/38686915
(cherry picked from commit 114eb9915431a2fdada35cfa2a3a755634d4efd0)

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -124,9 +124,11 @@ public:
 
   /// Create a SwiftASTContext from a Module.  This context is used
   /// for frame variable ans uses ClangImporter options specific to
-  /// this lldb::Module.
+  /// this lldb::Module.  The optional target is to create a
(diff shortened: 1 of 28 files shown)


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

PR #537, opened 2018-04-18
[LIT] Have lit run the lldb test suite

JDevlieghere:
This is the first in what will hopefully become a series of patches to
replace the driver logic in dotest.py with LIT. The motivation for this
change is that there's no point in maintaining two driver
implementations. Since all of the LLVM projects are using lit, this is
the obvious choice.

Obviously the goal is maintain full compatibility with the functionality
offered by dotest. As such we won't be removing anything until that
point has been reached.

This patch is the initial attempt (referred to as v1) to run the lldb
test suite with lit. To do so we introduced a custom LLDB test format
that invokes dotest.py with a single test file.

Differential revision: https://reviews.llvm.org/D45333

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@330275 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/lit/Suite/lit.cfg
+++ b/lit/Suite/lit.cfg
@@ -1,0 +1,29 @@
+# -*- Python -*-
+
+# Configuration file for the 'lit' test runner.
+
+import os
+
+import lit.formats
+
+# name: The name of this test suite.
+config.name = 'lldb-Suite'
+
+# suffixes: A list of file extensions to treat as test files.
+config.suffixes = ['.py']
+
+# test_source_root: The root path where tests are located.
+# test_exec_root: The root path where tests should be run.
+config.test_source_root = os.path.join(config.lldb_src_root, 'packages','Python','lldbsuite','test')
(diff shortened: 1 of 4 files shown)


JDevlieghere:

dcci (approved):

dcci:
Looks like there are a bunch of failures on Linux, Jonas. I recommend to run the full testsuite on macosx as well.

JDevlieghere:
Jup, the full test suite fails on macOS too.

JDevlieghere:
This makes sense actually, when we run the Swift tests the build script passes a bunch of options to `(lldb-)dotest(.py)`. We'll need to address these features in lit and change the build script to make this work.

dcci:
@JDevlieghere As we decided we're going to do this work on `master-next`, should we close this?

JDevlieghere:
@dcci yep

PR #539, opened 2018-04-18
Change the default for target.use-all-compiler-flags to true

adrian-prantl:
This is a safe thing to do now since we have a reliable fallback
mechanism in the module-specific scract AST contexts. The main
difference for the end user that are affected by this change will be
that they don't get a confusing error message befopre the first
expression is evaluated.

<rdar://problem/39243524>


adrian-prantl:
@swift-ci test and merge

PR #541, opened 2018-04-18
[CMake] Set the default test compiler deterministically

vedantk:
For swift-lldb, switch the default testing behavior so we use the
in-tree compilers by default.

Also, stop setting the cached LLDB_TEST_C_COMPILER variable multiple
times. Just set it one time to the compiler we want to test with.

This address a non-deterministic issue which sometimes caused
LLDB_TEST_C_COMPILER to be set to the empty string. My theory for how
this happened is:

  1. We set the cached variable LLDB_TEST_C_COMPILER for the first time
  in the top-level cmake file, and then

  2. We set the variable again in lit/CMakeLists.txt.

The second set did not use the FORCE modifier, so the empty string
should *never* have been assigned to LLDB_TEST_C_COMPILER, but perhaps
there was a bug in the version of cmake installed on the Linux bots
where we sporadically saw this issue.

At any rate, this PR should address the non-determinism by removing the
multiple assignments to the cached variable.

rdar://39537655, rdar://38935299

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,27 +72,8 @@ option(LLDB_INCLUDE_TESTS "Generate build targets for the LLDB unit tests."
   ${LLVM_INCLUDE_TESTS})
 if(LLDB_INCLUDE_TESTS)
   # BEGIN - Swift Mods
-  # The clang target is sometimes unavailable (rdar://38935299). Instead of
-  # failing the build when the target is missing, make the in-tree compiler the
-  # default compiler to test.
-
   set(LLDB_DEFAULT_TEST_C_COMPILER "${LLVM_BINARY_DIR}/bin/clang${CMAKE_EXECUTABLE_SUFFIX}")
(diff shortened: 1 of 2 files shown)


vedantk:
@swift-ci test

adrian-prantl (approved):
That looks like a step in the right direction to me. We don't want any automagic behavior.

gottesmm:
@vedantk can you elaborate more in the commit message on the exact failure. This: "remove some non-deterministic set commands which caused us to set bogus paths to test compilers." has no meaning to me.

dcci:
lg with Michael's comments addressed. This should have been painful to find, but glad you got to the bottom of it.

gottesmm:
@vedantk I was just thinking... we should look and see what cmake linters there are. It seems like something like this could be caught by a linter.

vedantk:
@gottesmm I've updated the message in the PR description. I'll update the commit message before merging so we don't have to run tests again. +1 on setting up a cmake linter as part of the validation tests, or something (akin to the python linting we do).

gottesmm:
For the cmake linting... I filed: https://bugs.swift.org/browse/SR-7473

gottesmm:
Also, +1 on the commit message. It is really clear and precise! Nice work!

PR #543, opened 2018-04-18
[CMake] Set the default test compiler deterministically

vedantk:
For swift-lldb, switch the default testing behavior so we use the
in-tree compilers by default.

Also, stop setting the cached LLDB_TEST_C_COMPILER variable multiple
times. Just set it one time to the compiler we want to test with.

This address a non-deterministic issue which sometimes caused
LLDB_TEST_C_COMPILER to be set to the empty string. My theory for how
this happened is:

We set the cached variable LLDB_TEST_C_COMPILER for the first time
in the top-level cmake file, and then

We set the variable again in lit/CMakeLists.txt.

The second set did not use the FORCE modifier, so the empty string
should never have been assigned to LLDB_TEST_C_COMPILER, but perhaps
there was a bug in the version of cmake installed on the Linux bots
where we sporadically saw this issue.

At any rate, this PR should address the non-determinism by removing the
multiple assignments to the cached variable.

rdar://39537655, rdar://38935299

(cherry picked from commit ac3f72dc7a6fa3de46530b45cc74631d26c448bd)

 Conflicts:
	CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,16 +71,10 @@ add_subdirectory(tools)
 option(LLDB_INCLUDE_TESTS "Generate build targets for the LLDB unit tests."
   ${LLVM_INCLUDE_TESTS})
 if(LLDB_INCLUDE_TESTS)
-  # TARGET clang works for a non-standalone build.
-  #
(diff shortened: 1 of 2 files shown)


vedantk:
@swift-ci test and merge

PR #545, opened 2018-04-18
[Swift] Only apply 'open' to classes and potentially-overridable decls

jrose-apple:
And don't apply it to anything marked `final`.

The test here *would* catch the issue here once the Swift AST verifier is updated to enforce the above conditions, but the AST verifier isn't run after the AST has been manipulated to add `open`.

No intended user-visible effect.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
@@ -837,9 +837,16 @@ void SwiftASTManipulator::MakeDeclarationsPublic() {
       if (swift::ValueDecl *value_decl =
               llvm::dyn_cast<swift::ValueDecl>(decl)) {
         auto access = swift::AccessLevel::Public;
-        if (swift::isa<swift::ClassDecl>(value_decl) ||
-            swift::isa<swift::ClassDecl>(value_decl->getDeclContext())) {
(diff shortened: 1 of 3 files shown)


jrose-apple:
@swift-ci Please test

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
```diff
@@ -837,9 +837,10 @@ void SwiftASTManipulator::MakeDeclarationsPublic() {
       if (swift::ValueDecl *value_decl =
               llvm::dyn_cast<swift::ValueDecl>(decl)) {
         auto access = swift::AccessLevel::Public;
-        if (swift::isa<swift::ClassDecl>(value_decl) ||
-            swift::isa<swift::ClassDecl>(value_decl->getDeclContext())) {
-          access = swift::AccessLevel::Open;
+        if (llvm::isa<swift::ClassDecl>(value_decl) ||
```
dcci: Can you add a comment here? The surrounding isn't really much commented but we should try improving this situation moving forward.

jrose-apple:
@swift-ci Please test

jrose-apple:
@swift-ci Please test

jimingham (approved):
Makes sense.

dcci (approved):

jrose-apple:
See also apple/swift#15996

PR #548, opened 2018-04-19
Fix a thinko in getting the persistent_expression_state.

jimingham:
For swift it was getting the value but then not setting it
to anything.


jimingham:

PR #551, opened 2018-04-20
Fix a crash when resolving overloads of C++ virtual methods

adrian-prantl:
The isOverload() method needs to account for situations where the two
methods being compared don't have the same number of arguments.

rdar://problem/39542960

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@330450 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 53113aaa3194cd8dbfa106ad9460e124db6f2570)


adrian-prantl:
@swift-ci test and merge

PR #554, opened 2018-04-20
[dotest] --skip-category should append and not override.

dcci:
<rdar://problem/38566150>


dcci:
@swift-ci test and merge

PR #563, opened 2018-04-25
[REPL] Make sure threads don't interrupt each other.

dcci:
Fixes sleep() in the REPL, which before just returned immediately.

<rdar://problem/39398038>

### DIFF
--- a/source/Expression/REPL.cpp
+++ b/source/Expression/REPL.cpp
@@ -295,6 +295,16 @@ void REPL::IOHandlerInputComplete(IOHandler &io_handler, std::string &code) {
       expr_options.SetColorizeErrors(colorize_err);
       expr_options.SetPoundLine(m_repl_source_path.c_str(),
                                 m_code.GetSize() + 1);
+
+      // There is no point in trying to run REPL expressions on just the
+      // current thread of the program, since the REPL tracks the states of
+      // individual threads as you would in a regular debugging session.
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci test and merge

source/Expression/REPL.cpp
```diff
@@ -295,6 +295,12 @@ void REPL::IOHandlerInputComplete(IOHandler &io_handler, std::string &code) {
       expr_options.SetColorizeErrors(colorize_err);
       expr_options.SetPoundLine(m_repl_source_path.c_str(),
                                 m_code.GetSize() + 1);
+
+      // We don't want other threads to interfere and block the thread
```
jimingham: That comment isn't quite right.  Maybe something like:

// There is no point in trying to run REPL expressions on just the current thread of the program, since the REPL
// track the states of individual threads as you would in a regular debugging session.  Interrupting the
// process to switch from one thread to many has observable effects (for instance it causes anything waiting in
// the kernel to be interrupted).  Since we aren't getting any benefit from doing this in the REPL, let's not.

jimingham (changes requested):
Can you fix the comment, I don't think it is quite right.

dcci:
Sure!

dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

PR #564, opened 2018-04-25
[REPL] Make sure threads don't interrupt each other.

dcci:
Fixes sleep() in the REPL, which before just returned immediately.

<rdar://problem/39398038>


dcci:
@swift-ci test and merge

PR #566, opened 2018-04-25
[REPL] More explicativec comment for the option setting.

dcci:
Thanks to Jim for pointing out!.


dcci:
@swift-ci test and merge

PR #571, opened 2018-04-26
[DataFmt] Fix formatting of bridged CFStrings

vedantk:
The formatter for Swift strings called CreateValueObjectFromAddress in
order to construct a ValueObject for bridged strings. In an unknown set
of cases, this API call can accidentally dereference the payload, giving
an invalid ValueObject that the NSString summary provider doesn't
understand.

Address the issue by making sure the address is never dereferenced when
constructing the ValueObject for bridged strings. As a follow-up, there
is a radar to investigate why the accidental dereference occurs
(r://39741576)

rdar://39710387

### DIFF
--- a/lit/SwiftREPL/CFString.test
+++ b/lit/SwiftREPL/CFString.test
@@ -1,0 +1,11 @@
+// Test that CFString works in the REPL.
+// REQUIRES: darwin
+
+// RUN: %lldb --repl < %s | FileCheck %s
+// CHECK: Welcome to Swift
+
+import Foundation
+
+let s = CFStringCreateWithCString(nil, "abcdefghijklmnop", kCFStringEncodingASCII)!
+let s2 = s as String
+// CHECK: s2: String = "abcdefghijklmnop"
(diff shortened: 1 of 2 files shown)


vedantk:
@swift-ci test and merge

vedantk:
@swift-ci test and merge

vedantk:
@swift-ci test and merge

jimingham:
Your comment doesn't make sense.  ValueObjectCreateFromAddress doesn't randomly add a dereference in some cases.  It always makes a Value of the incoming type starting at the address given.

What really seems to be the problem here is that what we previously thought was a pointer to the backing NSObject is really a pointer to a box containing the backing NSObject.  Your fix really just adds that extra level of dereference.  You could have gotten the same effect by reading a pointer from the address you found and passing that to CreateValueFromAddress.  

We need to understand why we misunderstood the layout of these classes.  If it was just something that changed in the layout, then that's fine, add the dereference however is convenient.  But the problem was never some oddity in ValueObjectCreateFromAddress.  It was a misunderstanding of the object layout.

vedantk:
You're right. I've made a not to get rid of the comment and document the ...FromAddress API.

PR #572, opened 2018-04-26
[DataFmt] Fix formatting of bridged CFStrings (4.2)

vedantk:
The formatter for Swift strings called CreateValueObjectFromAddress in
order to construct a ValueObject for bridged strings. In an unknown set
of cases, this API call can accidentally dereference the payload, giving
an invalid ValueObject that the NSString summary provider doesn't
understand.

Address the issue by making sure the address is never dereferenced when
constructing the ValueObject for bridged strings. As a follow-up, there
is a radar to investigate why the accidental dereference occurs
(r://39741576)

rdar://39710387
(cherry picked from commit f945d62)

### DIFF
--- a/lit/SwiftREPL/CFString.test
+++ b/lit/SwiftREPL/CFString.test
@@ -1,0 +1,11 @@
+// Test that CFString works in the REPL.
+// REQUIRES: darwin
+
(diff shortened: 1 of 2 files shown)


vedantk:
@swift-ci test and merge

PR #574, opened 2018-04-27
[debugserver] Fix handling of the 'g' packet

fredriss:
LLDB doesn't use this packet so we never hit this, but it looks like
some other projects talk to debugserver and are hitting an assert
(https://github.com/derekparker/delve/issues/1015).

We had an off by 1 in the accounting of the FPU structure sizes.
I added a test that basically just check that 'g' doesn't return
an error (currently it assert in debug builds). I didn't make
it an lldb-server test because it looks like lldb-server doesn't
implement the g packet.

rdar://39060083

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@331004 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/packages/Python/lldbsuite/test/tools/lldb-server/TestGdbRemoteGPacket.py
+++ b/packages/Python/lldbsuite/test/tools/lldb-server/TestGdbRemoteGPacket.py
@@ -1,0 +1,30 @@
+from __future__ import print_function
+
+
+import gdbremote_testcase
+from lldbsuite.test.decorators import *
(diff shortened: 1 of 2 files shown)


fredriss:
@swift-ci test and merge

karolz-ms:
@fredriss Is this going to be included in XCode 9.4 release?

fredriss:
This went to the swift-4.2-branch, so it'll be released alongside swift 4.2 (https://swift.org/blog/4-2-release-process/)

PR #575, opened 2018-04-27
Fix a thinko in the iteration over StructuredDataPlugin Create functi…

jimingham:
…ons.

The code was grabbing the first plugin, and then never getting
another one.

<rdar://problem/39779438>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@331012 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #581, opened 2018-04-27
Formatters need to obey their incoming StringPrinter options.

jimingham:
<rdar://problem/39802589>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/po/sys_types/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/po/sys_types/main.swift
@@ -13,6 +13,7 @@ import Cocoa
 
 func main() {
   var num = 22
+  //% self.expect("po num", substrs=["\\n", "\""], matching=False) # Make sure po doesn't escape non-printables.
   var str = "Hello world" //% self.expect("po num", substrs = ['22'])
   var arr = [1,2,3,4] 
   var nsarr = NSMutableArray(array: arr) //% self.expect("po str", substrs = ['Hello world'])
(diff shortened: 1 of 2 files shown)


jimingham:

packages/Python/lldbsuite/test/lang/swift/po/sys_types/main.swift
```diff
@@ -13,6 +13,7 @@ import Cocoa
 
 func main() {
   var num = 22
+  //% self.expect("po num", substrs=["\\n", "\""], matching=False) # Make sure po doesn't escape non-printables.
```
dcci, line 16: Thanks for fixing this!
I'm afraid we still have an issue/regression which started happening when the data formatter changed (also for SmallStrings)

Assume:

```
   1    func main() -> Int {
   2      let tinky = "p\n"
   3      var patatino = false
-> 4      return 123
   5    }
```

```
(lldb) po tinky
"p\n"
```

I would expect here `po` printing `p` (and then a newline). Is that correct?

jimingham:
> On Apr 30, 2018, at 1:29 PM, Davide Italiano <notifications@github.com> wrote:
> 
> @dcci commented on this pull request.
> 
> In packages/Python/lldbsuite/test/lang/swift/po/sys_types/main.swift:
> 
> > @@ -13,6 +13,7 @@ import Cocoa
>  
>  func main() {
>    var num = 22
> +  //% self.expect("po num", substrs=["\\n", "\""], matching=False) # Make sure po doesn't escape non-printables.
> 
> Thanks for fixing this!
> I'm afraid we still have an issue/regression which started happening when the data formatter changed (also for SmallStrings)
> 
> Assume:
> 
>    1    func main() -> Int {
>    2      let tinky = "p\n"
>    3      var patatino = false
> -> 4      return 123
>    5    }
> 
> (lldb) po tinky
> "p\n"
> 
> I would expect here po printing p (and then a newline). Is that correct?

Yes, po should render the string, not show it's contents.  We chased this down a bit, and the string that got to the printer had a \ and an n as separate characters.  That's not right, '\n' needs to be preserved as one character till the printing stage.

Jim

PR #585, opened 2018-04-30
Fix expression parser to not accept any type whose basename matches f…

adrian-prantl:
…or a type that must exist at root level

This patch fixes an issue where we weren't looking for exact matches in the expression parser and also fixed the type lookup logic in the Module.cpp. Tests added to make sure we don't regress.

Differential Revision: https://reviews.llvm.org/D46128

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@331227 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit c1ad1efd92650b261c2080d5e9cbad21e388504f)


adrian-prantl:
@swift-ci test and merge

PR #586, opened 2018-04-30
Remove premature caching of the global variables list in CompileUnit.

adrian-prantl:
This fixes a bug where

  (lldb) target var g_ptr

would populate the global variables list with exactly one entry
because SymbolFileDWARF::ParseVariables() was invoked with a list of
DIEs pre-filtered by name, such that a subsequent call to

  (lldb) fr var --show-globals

would only list that one variable, because CompileUnit::m_variables
was already initialized, fooling CompileUnit::GetVariableList().

CompileUnit::GetVariableList() grabs the *complete* list of variables
via (SymbolFileDWARF, ...)::ParseVariablesForContext and that still
calls CompileUnit::SetVariableList(variables) which acts as the
caching mechanism.

Differential Revision: https://reviews.llvm.org/D46220

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@331230 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit bc92d947a5445e2656babfacb90509bd03daa0cb)

### DIFF
--- a/packages/Python/lldbsuite/test/lang/c/global_variables/TestGlobalVariables.py
+++ b/packages/Python/lldbsuite/test/lang/c/global_variables/TestGlobalVariables.py
@@ -39,6 +39,12 @@ class GlobalVariablesTestCase(TestBase):
         environment = self.registerSharedLibrariesWithTarget(
             target, self.shlib_names)
 
+        # Test that static initialized variables can be inspected without process.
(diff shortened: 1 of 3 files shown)


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test

PR #587, opened 2018-05-01
 Move the persistent variable counter into Target

adrian-prantl:
Adopt swift plugin for upstream persistent variable handling changes.  …
Moving persistent variable counters into the target has the
side-effect that the numbering is now shared between all the scratch
SwiftASTContext objects.

rdar://problem/39299889


adrian-prantl:
@swift-ci test and merge

PR #592, opened 2018-05-02
Make the command-line REPL not trap C++ exceptions.

jimingham:
Assume that in the REPL, any C++ code that gets called from
swift will catch all the exceptions it throws.  This is a
reasonable expectation, since there's no way to sensibly throw
past a swift frame.

<rdar://problem/39700160>


jimingham:

PR #593, opened 2018-05-02
 Don't import modules from an Objective-C context into a Swift expression.

adrian-prantl:
This doesn't work in general because PerformAutoImport calls
ast->GetModule() which only succeeds if the module has been loaded
beforehand. In an Objective-C context there is no master .swiftmodule
that would pre-import the clang modules so this fails.
    
By disabling the feature we get a more reasonable error message.
   
<rdar://problem/39696005>


adrian-prantl:
@swift-ci test and merge

PR #594, opened 2018-05-02
Fix the .experimental. settings feature so that we don't return an error

jasonmolenda:
Fix the .experimental. settings feature so that we don't return an error
if an experimental setting has been removed/is missing.

Add tests for the .experimental. settings behaviors -- that they correctly
forward through to the real setting if it has become a real setting,
that they don't generate errors when a settig has been removed.

As Pavel notes in https://reviews.llvm.org/D45348, the way I'm suppressing
errors in the setting is not completely correct - if any of the setting
path components include "experimental", a missing setting would be declared
a non-error.  So

settings set target.experimental.setting-that-does-not-exist true

would not generate an error, which is correct.  But as Pavel notes, 

settings set setting-does-not-exist.experimental.run-stopped true

should generate an error because the unknown name occurs before the
"experimental".  The amount of change to do this correctly hasn't
thrilled me, so I'm leaving this as-is for now.

<rdar://problem/39223054> 
Differential Revision: https://reviews.llvm.org/D45348

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@331315 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 73723d5b2da4f3eff4153953ba5da06d860ca5f1)

### DIFF
--- a/packages/Python/lldbsuite/test/settings/TestSettings.py
+++ b/packages/Python/lldbsuite/test/settings/TestSettings.py
@@ -524,3 +524,41 @@ class SettingsCommandTestCase(TestBase):
                              "target.process.extra-startup-command",
                              "target.process.thread.step-avoid-regexp",
                              "target.process.thread.trace-thread"])
+
+    # settings under an ".experimental" domain should have two properties:
+    #   1. If the name does not exist with "experimental" in the name path,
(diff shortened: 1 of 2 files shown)


jasonmolenda:
@swift-ci test and merge

PR #596, opened 2018-05-02
Fixes for Swift-side changes to support inlinable autolinking [4.2]

slavapestov:
These two patches are already on the stable branch, and go with https://github.com/apple/swift/pull/16301 on the swift-4.2-branch.


slavapestov:
@swift-ci Please test

PR #598, opened 2018-05-02
This test fails with inject-local-variables set to false.

jimingham:
We will address that issue but the test still tests useful things,
so instead of xfailing it till we figure out why, I'm setting
inject-local-variables to true by hand in the test.

<rdar://problem/39908723>


jimingham:

PR #599, opened 2018-05-02
 Add support for dwarf expressions on global variables.

adrian-prantl:
The Swift compiler emits a DW_OP_deref for global variables in a fixed
size buffer, which is correct unless the variable uses inline
storage. This makes the majority of resilient types in Foundation work
as global variables.  The correct solution would be for LLDB to poke
at the runtime to figure out whether the storage is inline or not, but
until then this is the next best thing.

<rdar://problem/39767528>

### DIFF
--- a/include/lldb/Core/Value.h
+++ b/include/lldb/Core/Value.h
@@ -229,6 +229,9 @@ public:
 
   static const char *GetContextTypeAsCString(ContextType context_type);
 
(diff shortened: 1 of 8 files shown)


adrian-prantl:
test with https://github.com/apple/swift/pull/16318
@swift-ci test

PR #602, opened 2018-05-03
cherry-pick r331462.

adrian-prantl:
This is a change that only affects Swift and is NFC for the language
plugins on llvm.org. In Swift, we can have global variables with a
location such as DW_OP_addr <addr> DW_OP_deref. The DWARF expression
evaluator doesn't know how to apply a DW_OP_deref to a file address,
but at the very end we convert the file address into a load address.

This patch moves the file->load address conversion to right after the
result of the DW_OP_addr is pushed onto the stack so that a subsequent
DW_OP_deref (and potentially other operations) can be interpreted.

rdar://problem/39767528

Differential revision: https://reviews.llvm.org/D46362

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@331462 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit ef04d19bbb3bd3475db2d76b066611725bc632a4)

### DIFF
--- a/include/lldb/Core/Value.h
+++ b/include/lldb/Core/Value.h
@@ -229,6 +229,9 @@ public:
 
   static const char *GetContextTypeAsCString(ContextType context_type);
 
+  /// Convert this value's file address to a load address, if possible.
+  void ConvertToLoadAddress(SymbolContext sc);
+
   bool GetData(DataExtractor &data);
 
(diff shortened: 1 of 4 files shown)


adrian-prantl:
@swift-ci test and merge

PR #605, opened 2018-05-04
The on-ios-device command line lldb has an optimization where

jasonmolenda:
The on-ios-device command line lldb has an optimization where
when it and the inferior process both have the same shared cache
(a conglomeration of all libraries at the same fixed address for
all processes), lldb will read data out of its own memory to speed
things up.  The shared cache has a UUID, so lldb currently checks
that the UUID of its own shared cache matches that of the inferior.

This change adds one refinement to that -- it checks that the UUID
is the same and that the base address of the shared cache is the
same.  And only uses its local shared cache if they are both identical.

This involved using a different style of SPI with dyld to get lldb's
shared cache load address, but it's not especially difficult.

One unattractive part of the change is that I'm using the real
underlying types of task_t and kern_return_t instead of picking
them up from mach/mach.h.  The defines that get picked up there (a
lot from machine.h but others too) conflict with llvm/Support/MachO.h
even when I have mach.h included before our SafeMachO.h which
undefines most of the defines before including llvm/Support/MachO.h.
I'll need to augment the #undefs in SafeMachO.h to get this to
compile cleanly, but that'll be another day.

<rdar://problem/39868238>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@331497 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 83c5f870a1ed2f680162525fba501fdb85bd1935)

### DIFF
--- a/source/Plugins/ObjectFile/Mach-O/ObjectFileMachO.cpp
+++ b/source/Plugins/ObjectFile/Mach-O/ObjectFileMachO.cpp
@@ -2438,17 +2438,17 @@ size_t ObjectFileMachO::ParseSymtab() {
           // by
           // reading the memory for the __LINKEDIT section from this process.
 
-          UUID lldb_shared_cache(GetLLDBSharedCacheUUID());
-          UUID process_shared_cache(GetProcessSharedCacheUUID(process));
+          UUID lldb_shared_cache;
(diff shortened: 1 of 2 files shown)


jasonmolenda:
@swift-ci test and merge

PR #606, opened 2018-05-04
[SwiftASTContext] Print fatal error if there is one

JDevlieghere:
When we error out because of a swift error, we should print the fatal
error if it exists. I was debugging a Swift issue without realizing my
reproducer was outdated. This seems like something we'd want to
communicate to the end user.

In my particular case, this now prints:

  warning: Swift error in module Foo: the swift module file format is
  too old to be used by the version of the swift compiler in LLDB.
  Debug info from this module will be unavailable in the debugger.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1311,7 +1311,8 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
         StreamString ss;
         module_sp->GetDescription(&ss, eDescriptionLevelBrief);
         if (module_swift_ast->HasFatalErrors())
-          ss << ": " << module_swift_ast->GetFatalErrors().AsCString();
+          ss << ": "
+             << module_swift_ast->GetFatalErrors().AsCString("unknown error");
 
         target.GetDebugger().GetErrorFile()->Printf(
             "warning: Swift error in module %s.\n"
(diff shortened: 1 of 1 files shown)


JDevlieghere:
@swift-ci please test

adrian-prantl (commented):
Does this affect the output of any of the lang/swift/clangimporter testcases? Could we test it?

JDevlieghere:
None of the existing tests seem to be checking this particular error. Do you have any suggestions other than checking in a binary created with an old version of the swift compiler?

adrian-prantl:
Would something like `echo garbage > my.swiftmodule` work?

dcci (approved):
This is already an improvement on what we have. It would be great to test but I'm not particularly worried about it [it's an error message after all] (if Adrian's suggestion works great, otherwise I would check this one as is).

vedantk (approved):
I agree with Davide. We should test this if it can be done in a low-effort / maintainable way.

JDevlieghere:
Adrian, can you elaborate on how you'd get that "garbage" module into LLDB and to the point that it would trigger this error? I can't seem to find any test to use as inspiration.

adrian-prantl:
Instead of letting the swift driver run in build system mode, you'd have to separate out compile and link steps and then intercept the .swiftmodule before the .dSYM is generated.

You can run the swift compiler with -### for inspiration.

JDevlieghere:
Alright, I'll look into that over the weekend/next week.

jimingham:
I have seen a not inconsiderable number of reports where HasFatalErrors is true but the error string is empty.  I think in that case it's clearer to print some string like "unknown error" than just having a blank bit in the output.  Can you check for the case where the error string is empty and put in something like this instead?

JDevlieghere:
Apparently the linker is doing some sanity checking on the swift module. I'm all for having tests but this + the lack of other tests that manually invoke the different stages of the driver separately make me believe this doesn't really quality as "low-effort / maintainable" anymore. If anyone has another idea please do let me know!

```
ld: warning: ignoring file main.swiftmodule, file was built for unsupported file format (...) which is not the architecture being linked (x86_64): main.swiftmodule
```

JDevlieghere:

PR #608, opened 2018-05-04
 Honor the search paths of all swiftmodules in one lldb::Module.

adrian-prantl:
Prior to this commit LLDB would only load the search paths from the
first serialized AST in the current dylib, after this commit it
concatenates the options from all serialized ASTs in the current
dylib. This improved the debuggability of projects using static
archives at the expense of potential clang module import failures when
the options from within one module are conflicting.

<rdar://problem/39833129>


adrian-prantl:
@swift-ci test and merge

PR #612, opened 2018-05-04
 Add a test for importing a c module into the Swift REPL

adrian-prantl:
Because of limitations in our test driver an its inability to pass
extra flags around this test is wedged into a Makefile at the moment.

<rdar://problem/37836639>


adrian-prantl:
@swift-ci test and merge

PR #613, opened 2018-05-06
[SwiftASTContext] Print fatal error if there is one

JDevlieghere:
Cherry-picking changes from #606 to swift-4.2-branch.


JDevlieghere:

PR #618, opened 2018-05-09
Calculate dynamic type of Optional's payload whenever possible

vedantk:
When dumping the value of an Optional, we may not calculate the dynamic
type of the payload. This patch makes it so that we always calculate
this dynamic type. The user-visible change occurs for code like this:

  let d: NSObject? = MyClass() as NSObject

Before this patch, lldb would dump the value of 'd' as:

  d: NSObject? = 0x0000000100200060

After this patch, it would instead print:

  d: NSObject? = 0x0000000100200060 {
    ObjectiveC.NSObject = {}
  }

rdar://39736324

### DIFF
--- a/source/Plugins/Language/Swift/SwiftOptional.cpp
+++ b/source/Plugins/Language/Swift/SwiftOptional.cpp
@@ -42,7 +42,6 @@ std::string lldb_private::formatters::swift::SwiftOptionalSummaryProvider::
 // retrieve the value of the Some case..
 static PointerOrSP
 ExtractSomeIfAny(ValueObject *optional,
-                 lldb::DynamicValueType dynamic_value = lldb::eNoDynamicValues,
                  bool synthetic_value = false) {
   if (!optional)
     return nullptr;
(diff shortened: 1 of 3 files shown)


vedantk:
As this is all unfamiliar territory for me, I'd appreciate any feedback on whether/not this is sane, and will wait on that before kicking off CI.

source/Plugins/Language/Swift/SwiftOptional.cpp
```diff
@@ -42,7 +42,6 @@ std::string lldb_private::formatters::swift::SwiftOptionalSummaryProvider::
 // retrieve the value of the Some case..
 static PointerOrSP
 ExtractSomeIfAny(ValueObject *optional,
-                 lldb::DynamicValueType dynamic_value = lldb::eNoDynamicValues,
                  bool synthetic_value = false) {
```
dcci, line 45: All the callers of `ExtractSomeIfAny`, always pass `true` to `synthetic_value`. Can we get rid of the extra argument?

dcci (commented):
Probably is just historical, but I don't see why we might want the static version of an object. I'm not deeply familiar with this code either. Jim?

source/Plugins/Language/Swift/SwiftOptional.cpp
vedantk: Yes, I have that cleanup (and another for IsNonTriviallyManagedReferenceType) ready locally, but I'd rather do it in a different PR.

jimingham:
So the default value for dynamic_value in ExtractSomeIfAny is never used so far as I can tell.  All the callers pass in the ValueObjects's value for this.  That's because somebody might do:

(lldb) frame variable -d no-dynamic-values SomeType

and we should obey that setting.  So I don't think it is right to override the incoming setting.  However, fetching the dynamic type should be the default for frame variable, so I'd want to know how the simple "frame variable" commands in the test you're fixing end up with a value object whose setting for dynamic values is no dynamic.

jimingham (changes requested):
I don't think it is right to override the incoming ValueObject's dynamic type lookup setting.  Rather I'd like to figure out why a plain "frame variable" ends up passing in noDynamicValues when the default is to fetch dynamic values.

vedantk:
> I don't think it is right to override the incoming ValueObject's dynamic type lookup setting. Rather I'd like to figure out why a plain "frame variable" ends up passing in noDynamicValues when the default is to fetch dynamic values.

@jimingham Makes sense, I'll keep looking. I'd be curious to know if you have any hints re: how we determine whether or not a swift ValueObject's type should be dynamically set?

jimingham:
For frame variable it should either come from an explicit option setting, or it should look up the target property:

target.prefer-dynamic-value

and get the value from there.

vedantk:
Hm, target.prefer-dynamic-value really does default to eDynamicDontRunTarget. So that's a dead-end.

What's happening is that we pass the ValueObject for "d" into SwiftLanguageRuntime::CouldHaveDynamicValue, and it says "false". This happens because "d", as an enum, is considered to have a fully realized type by SwiftASTContext::IsPossibleDynamicType. Is that the correct logic?

ISTM that there isn't a clear enough distinction between an enum having a dynamic type, and one of its ADT payloads having a dynamic type.

ExtractSomeIfAny decides whether or not to calculate the dynamic type of a payload based on whether the parent enum is dynamic, but ISTM it never will be.

@jimingham Perhaps the right fix is to set `is-possibly-dynamic` on the payload's ValueObject. Is the way to do that to modify SwiftASTContext::IsPossiblyDynamicValue to check if the parent is an enum, and to return true if so?

jimingham:
I wouldn't muck with whether a value is possibly dynamic when it isn't, then we'll do work that it probably going to fail, and might cause unnecessary errors.  And anyway, we know that's wrong...

Seems to me that what's getting conflated is "this value object was requested with eDynamicDontRunTarget" and "this value object or one of its parents actually has a dynamic value".  The latter implies the former, but not the other way 'round.

It looks like we rely on propagating this information about the request to callers along a side chain (e.g. passing the appropriate value to use_dynamic in:

StackFrame::GetValueObjectForFrameVariable(const VariableSP &variable_sp,
                                           DynamicValueType use_dynamic) {

when we make the thing, and then at some point we lose the side value and have to fall back on:

valobj.GetDynamicValueType()

which only relays what was requested if the object happens to have been dynamic.

I think the actual right fix is to record the requested dynamic behavior in the value object when it gets made, and then refer to that not whether it or any of its parents actually ARE dynamic values to determine what to do with its children. 

 I don't know whether you could get away with recording the requested type and returning it in  ValueObject::GetDynamicValueTypeImpl (which is currently returning eNoDynamicValues).  We'd also have to check the uses of GetDynamicValueType to see which of them really mean "I or one of my parents was a DynamicValue" and which just mean 'I should fetch dynamic values if available..."  There a handful of uses and I don't understand all the call sights without some thinking.

Another short term hack would be to consult the target setting in the calls to ExtractSomeIfAny.  That way we'll mostly give dynamic values - which is actually what people usually want, but if they explicitly say "No dynamic values" by setting the setting it will not disobey that.

vedantk:
> Another short term hack would be to consult the target setting in the calls to ExtractSomeIfAny. That way we'll mostly give dynamic values - which is actually what people usually want, but if they explicitly say "No dynamic values" by setting the setting it will not disobey that.

@jimingham I've updated the PR to do this (PTAL) because I'm unsure about what a more general fix would look like.

> I think the actual right fix is to record the requested dynamic behavior in the value object when it gets made, and then refer to that not whether it or any of its parents actually ARE dynamic values to determine what to do with its children.

Yep, I agree with this, I'm just getting lost in the specifics. For starters, what *is* the correct/requested dynamic behavior? Is it sometimes the target prefer-dynamic setting? Or is it the result of SwiftLanguageRuntime::CouldHaveDynamicValue?

If it's just the latter, please see my earlier question about whether or not the result of CouldHaveDynamicValue is correct for "d" (the Optional<NSObject> described in the commit description):

> What's happening is that we pass the ValueObject for "d" into SwiftLanguageRuntime::CouldHaveDynamicValue, and it says "false". This happens because "d", as an enum, is considered to have a fully realized type by SwiftASTContext::IsPossibleDynamicType. Is that the correct logic?

> I don't know whether you could get away with recording the requested type and returning it in ValueObject::GetDynamicValueTypeImpl (which is currently returning eNoDynamicValues). We'd also have to check the uses of GetDynamicValueType to see which of them really mean "I or one of my parents was a DynamicValue" and which just mean 'I should fetch dynamic values if available..." There a handful of uses and I don't understand all the call sights without some thinking.

It sounds like this is really worth doing, possibly by introducing a new method on ValueObject to distinguish between the "is this thing dynamic" and the "is this thing, or one of its parents, dynamic" cases. I can file a bug so we don't lose track of it.

jimingham:
Until we get the value object to record what the user requested when they made it, once you get to the children you're just guessing, as you've found out.  

If I'm reading the code correctly, a value object will only return true to CouldHaveDynamic if the root value object was made requesting dynamic values.  So a "true" there means you should also fetch dynamic values for the child.  

But if that returns false then you don't know.  At that point the best you can do at present is check the target setting.  

If you combine these two tests, then the case you will get wrong is when the target setting was "no-dynamic-values" but user asked for dynamic values on a particular ValueObject (AND nothing in the path from the root to the child actually could have a dynamic value.)  That's not too bad, and seems the best you are going to do until we get the root value object to record whether dynamic values were quested.  

The latter does seem like something we should do but not for this patch.

vedantk:
@jimingham Is the approach in this patch OK for now? I can clone the bug for an audit of the process by which we determine whether/not to get an object's dynamic value.

jimingham:
I think it would be better to leave ExtractSomeIfAny the way it is, and then if the incoming value is no-dynamic-value, check the target setting and override it.  You want to first check CouldHaveDynamicValue anyway, and this gets you that part for free.  Put a comment in the override saying this is a temporary hack, and then when we add "ValueObject::RequestedDynamicValue" or whatever, you can remove the override and have the callers pass that in instead.

vedantk:
If we were to leave ExtractSomeIfAny unmodified, we'd have to duplicate logic to pass in `use_dynamic` in its callers, and it's unclear to me why that's an improvement.

Don't we eventually want ExtractSomeIfAny to check "ValueObject::RequestedDynamicValue"? If so, why should the callers be responsible for passing in a `use_dynamic` option? Put another way, why should callers of ExtractSomeIfAny be able to override the result of a hypothetical "ValueObject::RequestedDynamicValue"?

jimingham:
Okay, that makes sense.  But in ExtractSomeIfAny, first check the CouldHaveDynamicValue and then the target setting if it is false.  And put in a FIXME saying that this is an incomplete calculation, and we store the requested dynamic setting in the root value object and do the right thing with that.

vedantk:
@jimingham I've added the extra check and fixme.

vedantk:
@swift-ci test

jimingham (approved):

jimingham:
That looks fine.

PR #621, opened 2018-05-09
[stable] Fix TSan reports for Swift access races, and add a testcase.

kubamracek:
We currently fail to report "Swift access races" (race between mutating methods on a struct) as such, and we only show a generic "Race on a library object". Let's fix that and add a test case.

### DIFF
--- a/packages/Python/lldbsuite/test/functionalities/tsan/swift-access-race/Makefile
+++ b/packages/Python/lldbsuite/test/functionalities/tsan/swift-access-race/Makefile
@@ -1,0 +1,7 @@
+LEVEL = ../../../make
+
+SWIFT_SOURCES := main.swift
+
+include $(LEVEL)/Makefile.rules
+
+SWIFTFLAGS += -sanitize=thread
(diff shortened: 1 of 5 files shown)


kubamracek:
Note that this requires <https://reviews.llvm.org/D46661> to land first.

adrian-prantl:
Could you recreate the pull request against swift-4.2-branch? swift-4.2 is automerged into stable, which is automerged into upstream-with-swift.

adrian-prantl:
That is, unless the dependency doesn't go into swift-4.2-branch.

kubamracek:
Yes, this is for 4.2. Here's the new PR: <https://github.com/apple/swift-lldb/pull/622>

adrian-prantl:
Can this also be triggered from Objective-C? It would be nice if this could go into llvm.org instead.

kubamracek:
The testcase needs to be Swift-specific, and it doesn't make much sense to split the implementation and the test. I'll upstream the changes in TSanRuntime.{h,cpp} after this lands.

PR #622, opened 2018-05-09
[4.2] Fix TSan reports for Swift access races, and add a testcase.

kubamracek:
We currently fail to report "Swift access races" (race between mutating methods on a struct) as such, and we only show a generic "Race on a library object". Let's fix that and add a test case.

### DIFF
--- a/packages/Python/lldbsuite/test/functionalities/tsan/swift-access-race/TestTsanSwiftAccessRace.py
+++ b/packages/Python/lldbsuite/test/functionalities/tsan/swift-access-race/TestTsanSwiftAccessRace.py
@@ -1,0 +1,79 @@
+# TestTsanSwift.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""
+Test Swift support of TSan.
+"""
+import lldb
+import lldbsuite.test.decorators as decorators
+import lldbsuite.test.lldbtest as lldbtest
+import lldbsuite.test.lldbutil as lldbutil
+import os
+import unittest2
+import json
+
+
+class TsanSwiftAccessRaceTestCase(lldbtest.TestBase):
+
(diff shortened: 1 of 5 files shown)


kubamracek:
Note that this requires https://reviews.llvm.org/D46661 to land first.

adrian-prantl:
Can this also be triggered from Objective-C? It would be nice if this could go into llvm.org instead.

adrian-prantl:
I guess I missed the "Swift" in the title :-)

kubamracek:
The testcase needs to be Swift-specific, and it doesn't make much sense to split the implementation and the test. I'll upstream the changes in TSanRuntime.{h,cpp} after this lands.

packages/Python/lldbsuite/test/functionalities/tsan/swift-access-race/main.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+// -----------------------------------------------------------------------------
+import Foundation
+
```
adrian-prantl: isn't DispatchQueue part of Dispatch and not Foundation?

packages/Python/lldbsuite/test/functionalities/tsan/swift-access-race/TestTsanSwiftAccessRace.py
```diff
@@ -0,0 +1,79 @@
...
+    def test_tsan_swift(self):
+        self.build()
+        self.do_test()
+
+    def setUp(self):
+        lldbtest.TestBase.setUp(self)
+        self.main_source = "main.swift"
+        self.main_source_spec = lldb.SBFileSpec(self.main_source)
+
+    def do_test(self):
```
adrian-prantl, line 41: A comment explaining what this test is testing would be nice.

kubamracek:
Updated the PR.

kubamracek:
@adrian-prantl any more comments?

adrian-prantl (approved):
LGTM with outstanding comments addressed.

kubamracek:
Okay, the compiler-rt patch is now in swift-4.2-branch.

kubamracek:

PR #623, opened 2018-05-10
Run GetTypeSystemForLanguage(Swift) in parallel for every module.

fredriss:
This can shave quite a bit of time from the first expression
evaluation in a session. GetTypeSystemForLanguage() on a module
runs the import of all the referenced clang modules, which means
it will likely rebuild all of them.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1338,6 +1338,11 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
   }
 
   if (target.GetSwiftCreateModuleContextsInParallel()) {
+    // The first call to GetTypeSystemForLanguage() on a module will
+    // trigger the import (and thus most likely the rebuild) of all
+    // the Clang modules that were imported in this module. This can
+    // be a lot of work (potentially ten seconds per module), but it
+    // can be performed in parallel.
     llvm::ThreadPool pool;
     for (size_t mi = 0; mi != num_images; ++mi) {
       auto module_sp = target.GetImages().GetModuleAtIndex(mi);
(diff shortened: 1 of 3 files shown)


fredriss:
@swift-ci please test

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1336,6 +1337,17 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
     handled_sdk_path = true;
   }
 
+  if (target.GetSwiftCreateModuleContextsInParallel()) {
```
adrian-prantl, line 1340: Could you please add a comment explaining what this block is doing here?

source/Target/Target.cpp
```diff
@@ -4224,6 +4227,12 @@ FileSpecList &TargetProperties::GetSwiftModuleSearchPaths() {
   return option_value->GetCurrentValue();
 }
 
+bool TargetProperties::GetSwiftCreateModuleContextsInParallel() const {
```
jimingham: Could you put this in the target.experimental properties instead?  I don't think this is a setting we should keep around forever, and if you put it in experimental we can remove it without it breaking people's .lldbinit files.

jimingham (changes requested):
Just where the property goes.

source/Target/Target.cpp
```diff
@@ -3635,6 +3635,8 @@ static PropertyDefinition g_properties[] = {
     {"swift-module-search-paths", OptionValue::eTypeFileSpecList, false, 0,
      nullptr, nullptr,
      "List of directories to be searched when locating modules for Swift."},
+    {"swift-create-module-contexts-in-parallel", OptionValue::eTypeBoolean, false, true,
+     nullptr, nullptr, "Create modules AST context in parallel."},
```
adrian-prantl: `Create the per-module Swift AST contexts in parallel.` ?

adrian-prantl (approved):
Thanks!

fredriss:
@swift-ci please test

fredriss:
@swift-ci test

adrian-prantl:
@swift-ci test

jimingham (approved):
looks good to me.

PR #626, opened 2018-05-10
Really turn off all the ObjC runtime ISA expression logs if not verbose.

jimingham:
We apparently only turned off one of them.


jimingham:

dcci:
Should we have the same change upstream? If we do already, I apologize for the noise.

jimingham:
My bad.  I'll cancel this an do it through llvm.org.

dcci:
IIRC Adrian did a similar change so you might end up just back porting it.
Regardless, thanks for doing this :) [as I wanted to do it for a while]

PR #627, opened 2018-05-11
Add a lock to PlatformPOSIX::DoLoadImage

fredriss:
Multiple threads could be calling into DoLoadImage concurrently,
only one should be allowed to create the UtilityFunction.

This is discussed upstream in https://reviews.llvm.org/D46733, but
this seems fine for now.

rdar://problem/40140766


fredriss:
@swift-ci test and merge

PR #631, opened 2018-05-14
 Avoid adding duplicate configuration macros to clang importer options.

adrian-prantl:
When concatenating the clang options from all dylibs we don't unique
configuration macros which can lead to more unnecessary module
configurations to be built.

<rdar://problem/40220286>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1320,6 +1320,10 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
   std::shared_ptr<SwiftASTContextForExpressions> swift_ast_sp(
       new SwiftASTContextForExpressions(target));
 
+  Log *log(GetLogIfAllCategoriesSet(LIBLLDB_LOG_TYPES));
+  if (log)
+    log->Printf("SwiftASTContext::CreateInstance(Target)");
+  
   if (!arch.IsValid())
     return TypeSystemSP();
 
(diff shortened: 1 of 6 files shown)


adrian-prantl:
@swift-ci test

adrian-prantl:
This actually may dereference one byte too many.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1591,7 +1591,10 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
                      osi < ose; ++osi) {
                   const char *clang_argument =
                       ast_context->GetClangArgumentAtIndex(osi);
-                  swift_ast_sp->AddClangArgument(clang_argument, true);
+                  bool force = true;
+                  if (clang_argument && clang_argument[0] == '-' && clang_argument[1] == 'D')
+                    force = false;
+                  swift_ast_sp->AddClangArgument(clang_argument, force);
```
jrose-apple: Don't forget that people write macros as both `-DFOO` and `-D FOO`. And don't forget about `-U` as well! (But I think it's okay to not worry about -D and -U fighting across modules.)

adrian-prantl: Thanks for pointing this out. It turns out the passing `-Xcc -D -Xcc SPACE` to the swift compiler causes LLDB to fail creating a ClangImporter entirely...

adrian-prantl:
@swift-ci test

adrian-prantl:
Ha. Looks like my force-push obliterated @jrose-apple review comments...

fredriss:
Looks like this should work

fredriss (approved):

adrian-prantl:
@swift-ci nominate
Explanation: When concatenating the clang options from all dylibs we don't unique configuration macros which can lead to more unnecessary module configurations to be built.
Scope: Debugging of Swift applications (all platforms)
Radar: rdar://problem/35578590 / rdar://problem/40220286
Risk: Low. This is a pointed fix in the clang importer options handling in LLDB.
Testing: Regression test in the LLDB test suite.

PR #632, opened 2018-05-15
[DataFmt] 32-bit String: Handle boxed NSStrings and _SwiftStringStorage

vedantk:
Add support for two more layouts of 32-bit Swift.String:

1) The ADT payload inside of the String contains a boxed NSString
   pointer Note that this isn't always true, the runtime may choose not to
   wrap the pointer with one level of indirection, without a clear
   indication in the documented set of control bits.

2) The payload contains a _SwiftStringStorage instance.

rdar://40112101

### DIFF
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -18,6 +18,8 @@
 #include "lldb/Utility/DataBufferHeap.h"
 #include "lldb/Utility/Status.h"
 #include "lldb/Target/SwiftLanguageRuntime.h"
+#include "llvm/ADT/Optional.h"
+#include "llvm/ADT/StringRef.h"
 
 // FIXME: we should not need this
 #include "Plugins/Language/CPlusPlus/CxxStringTypes.h"
@@ -28,6 +30,7 @@ using namespace lldb;
 using namespace lldb_private;
 using namespace lldb_private::formatters;
 using namespace lldb_private::formatters::swift;
+using namespace llvm;
 
 bool lldb_private::formatters::swift::Character_SummaryProvider(
     ValueObject &valobj, Stream &stream, const TypeSummaryOptions &options) {
@@ -138,6 +141,40 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
       StringPrinter::ReadStringAndDumpToStreamOptions());
 }
 
+/// Get an Obj-C object's class name, if one is present.
+static Optional<StringRef> getObjC_ClassName(ValueObject &valobj,
+                                             Process &process) {
+  ObjCLanguageRuntime *runtime =
+      (ObjCLanguageRuntime *)process.GetLanguageRuntime(
+          lldb::eLanguageTypeObjC);
+  if (!runtime)
+    return None;
+
+  ObjCLanguageRuntime::ClassDescriptorSP descriptor(
+      runtime->GetClassDescriptor(valobj));
+
+  if (!descriptor.get() || !descriptor->IsValid())
+    return None;
+
+  ConstString class_name_cs = descriptor->GetClassName();
+  return class_name_cs.GetStringRef();
+}
+
+/// If valobj is a _SwiftRawStringStorage instance, retrieve the payload address
+/// of the character data and determine whether the string is in UTF-16.
+static bool GetRawStringStoragePayload(Process &process, ValueObject &valobj,
+                                       uint64_t &payloadAddr, bool &isUTF16) {
+  CompilerType ty = valobj.GetCompilerType();
+
+  auto objCName = getObjC_ClassName(valobj, process);
+  if (!objCName || !objCName->contains("_SwiftStringStorage"))
+    return false;
+
+  isUTF16 = !objCName->endswith("UInt8_");
+  payloadAddr = valobj.GetValueAsUnsigned(0) + 16;
+  return true;
+}
+
 bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
     ValueObject &valobj, Stream &stream,
     const TypeSummaryOptions &summary_options,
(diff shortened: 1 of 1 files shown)


vedantk:
@swift-ci test

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -138,6 +141,38 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
       StringPrinter::ReadStringAndDumpToStreamOptions());
 }
 
+/// Get an Obj-C object's class name, if one is present.
+static Optional<StringRef> getObjC_ClassName(ValueObject &O, Process &P) {
```
jimingham: Don't use one character upper case variable names in lldb.  Function arguments should be short lower-cased words.

jimingham: This also seems like a generally useful bit of boiler-plate wrapping.  Maybe it should go in the ObjC Language Runtime class?

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -138,6 +141,38 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
+  if (!descriptor.get() || !descriptor->IsValid())
+    return None;
+
+  ConstString class_name_cs = descriptor->GetClassName();
+  return class_name_cs.GetStringRef();
+}
+
+/// If O is a _SwiftRawStringStorage instance, retrieve the payload address
+/// of the character data and determine whether the string is in UTF-16.
+static bool GetRawStringStoragePayload(Process &P, ValueObject &O,
```
jimingham: Same comment about names.  Local variables and arguments are all lower case, _ separated words.

jimingham: And try to use the same word everywhere for the same thing, we generally use "process" and "valobj".

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -231,17 +266,29 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
 
-    // Warning: Using ValueObject::CreateValueObjectFromAddress can create an
-    // invalid ValueObject here in an unknown set of cases. Until this is fixed,
-    // use CreateValueObjectFromData (rdar://39741576).
+    bool success = false;
+    ValueObjectSP nsstring;
+
+    // Swift usually (but not always) wraps the NSString with one level of
+    // pointer indirection. So, first, dereference & format the string.
     DataExtractor DE(&startAddress, process->GetAddressByteSize(),
```
jimingham: DE really isn't a good name either...  I think I missed this in an earlier review.

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -231,17 +266,29 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
                      process->GetByteOrder(), process->GetAddressByteSize());
-    ValueObjectSP nsstring = ValueObject::CreateValueObjectFromData(
+    nsstring = ValueObject::CreateValueObjectFromData(
         "nsstring", DE, valobj.GetExecutionContextRef(), id_type);
-
     if (nsstring)
```
jimingham: Sometimes a function that produces a ValueObjectSP will fail by returning an empty shared pointer.  But we try not to do that, since it carries no information.  More often failure is signified by `nsstring->GetError().Fail()`  You should check that here as well as whether the pointer is empty.

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -231,17 +266,29 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
-      return NSStringSummaryProvider(*nsstring.get(), stream, summary_options);
-    return false;
+      success =
+          NSStringSummaryProvider(*nsstring.get(), stream, summary_options);
+
+    // We may have an NSString pointer inline, so try formatting it directly.
+    if (!success) {
```
jimingham: Is there really no way to tell whether this is boxed or inline?  Your workaround is okay; the chances that the first words of an inlined string are a valid pointer to another string being fairly small...  But it would be nice if the stdlib could provide a more definite way for us to figure this out.

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -322,7 +369,8 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
       // Small strings aren't emitted yet.
       return false;
     } else {
-      // The associated value in the selected enum case is an NSString.
+      // The associated value in the selected enum case is either an NSString,
```
jimingham, line 374: Does the variant case not tell you which of these two possibilities was chosen?  It bugs me that we have to guess.

jimingham (changes requested):
A few trivial naming changes.  Then it seems like the stdlib is making us guess how this data is laid out.  If that's necessary for now, then it's okay to get this in, but we should circle back and get them to really tell us.  OTOH, then runtime must have some definitive way of figuring this out, so it would be better if we can tap into that.

source/Plugins/Language/Swift/SwiftFormatters.cpp
vedantk: I think making GetClassDescriptor() return an Optional descriptor would do the job. I'd rather leave that refactor as a follow-up, as it requires changing upstream code.

source/Plugins/Language/Swift/SwiftFormatters.cpp
vedantk: I've been staring at the bits inside of these strings and going through the scant set of docs, and can't find any indicator of whether the NSString is boxed. :/.

vedantk:
@jimingham I've made the suggested changes and fixed the code comments. In the previous iteration, I had the comments for the boxed path and the unboxed path swapped.

vedantk:
@swift-ci Please test

jimingham (approved):
This is okay for now.  But can file a request for the stdlib folks to give us some better way of determining which of the various flavors the data is actually using, that will make this more supportable.

vedantk:
Thanks! I filed: r://40267572

PR #633, opened 2018-05-15
[DataFmt] 32-bit String: Handle boxed NSStrings and _SwiftStringStorage

vedantk:
Add support for two more layouts of 32-bit Swift.String:

1) The ADT payload inside of the String contains a boxed NSString
   pointer Note that this isn't always true, the runtime may choose not to
   wrap the pointer with one level of indirection, without a clear
   indication in the documented set of control bits.

2) The payload contains a _SwiftStringStorage instance.

rdar://40112101
(cherry picked from commit ee1e26e2d64160367d3c45d3f57a591bc36bdb66)


vedantk:
@swift-ci test and merge

PR #634, opened 2018-05-16
Calculate the dynamic type of Optional's payload whenever possible

vedantk:
When dumping the value of an Optional, we may not calculate the dynamic
type of the payload. This patch makes it so that we always calculate
this dynamic type. The user-visible change occurs for code like this:

  let d: NSObject? = MyClass() as NSObject

Before this patch, lldb would dump the value of 'd' as:

  d: NSObject? = 0x0000000100200060

After this patch, it would instead print:

  d: NSObject? = 0x0000000100200060 {
    ObjectiveC.NSObject = {}
  }

rdar://39736324
(cherry picked from commit 21e6432acc12674ba9d2eed03b2a51313c5440a5)

### DIFF
--- a/lit/SwiftREPL/OptionalWithDynamicType.test
+++ b/lit/SwiftREPL/OptionalWithDynamicType.test
@@ -1,0 +1,32 @@
+// Test that we print the dynamic type of the object contained within an
+// optional.
+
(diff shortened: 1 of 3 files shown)


vedantk:
@swift-ci test

PR #635, opened 2018-05-17
Disable parallel extraction of DIEs in SymbolFileDWARF::Index()

fredriss:
We shouldn't really be using that codepath anyway.
In some cases under the REPL, the DIE indexing code would deadlock.
The reason is that the thread asking for the indexing holds the
Module lock and some workers that it spawns will want the lock too.
For now just disabled the parallel extraction of the DIEs until we
find a more principled solution.

rdar://problem/38461035


fredriss:
@swift-ci please test

PR #637, opened 2018-05-18
 Set the minimum deployment target of the Swift REPL to the Host OS version

adrian-prantl:
Otherwise the minimum deployment target is picked up from the stub,
which may be lower.

rdar://problem/40279463

### DIFF
--- a/lit/SwiftREPL/DeploymentTarget.test
+++ b/lit/SwiftREPL/DeploymentTarget.test
@@ -1,0 +1,11 @@
+// Test that the REPL can call a *really* new function.
+// REQUIRES: darwin
+
+// RUN: echo -n "@available(macOS " >%t.swift
+// RUN: sysctl kern.osproductversion | awk '{printf $2}' >>%t.swift
+// RUN: echo  ", *) func newAPI() -> Int { return 11+12 }" >>%t.swift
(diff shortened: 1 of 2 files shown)


adrian-prantl:
@swift-ci test and merge

source/Utility/ArchSpec.cpp
```diff
@@ -1447,8 +1447,14 @@ void ArchSpec::PiecewiseTripleCompare(
...
-  os_version_different = (me.getOSMajorVersion() != them.getOSMajorVersion());
-
+  unsigned me_major, me_minor, me_patch;
+  me.getOSVersion(me_major, me_minor, me_patch);
+  unsigned them_major, them_minor, them_patch;
+  them.getOSVersion(them_major, them_minor, them_patch);
+  os_version_different = (me_major != them_major) ||
+                         (me_minor != them_minor) ||
+                         (me_patch != them_patch);
+
```
adrian-prantl: Note that I did not find any other callers of this function.

adrian-prantl: There are, however, plenty of calls to Target::SetArchitecture().

adrian-prantl:
Reimplemented it in a far less invasive fashion.

adrian-prantl:
@swift-ci test and merge

PR #639, opened 2018-05-18
[swift-pr] Include TestSwiftGenericExpression as part of standard pr …

dcci:
…testing.

This would've caught problem triggered by a change in swiftc.


dcci:

PR #640, opened 2018-05-18
Fix _NSCFBoolean data formatter.

JDevlieghere:
In r265181 the test for the NSCFBoolean data formatter was removed.
Later, in r279353 and r279446 a new implementation was provided for the
formatter, which I believe never worked (and this wasn't caught because
the test was never re-enabled).

This commit fixes the bug and re-enables the old test case.

Differential revision: https://reviews.llvm.org/D47014

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@332700 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit e2b500258d6d5352c48c308449e0fc0981b30923)


JDevlieghere:
rdar://36336178

JDevlieghere:

PR #641, opened 2018-05-18
Skip images without a serialized Swift AST when collecting header

adrian-prantl:
search options for the scratch context.

This is both a performance optimization and as a side effect can avoid
spurious warning messages.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1578,58 +1578,55 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
           return;
 
         SymbolVendor *sym_vendor = module_sp->GetSymbolVendor();
+        if (!sym_vendor)
+          return;
 
-        if (sym_vendor) {
-          std::vector<std::string> module_names;
+        std::vector<std::string> module_names;
+        SymbolFile *sym_file = sym_vendor->GetSymbolFile();
+        if (!sym_file)
+          return;
 
-          SymbolFile *sym_file = sym_vendor->GetSymbolFile();
-          if (sym_file) {
-            Status sym_file_error;
-            SwiftASTContext *ast_context =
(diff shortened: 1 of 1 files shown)


adrian-prantl:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1037,6 +1037,16 @@ static bool DeserializeAllCompilerFlags(SwiftASTContext &swift_ast,
   return false;
 }
 
+/// Return whether this module contains any serializes Swift ASTs.
```
jimingham: serialized

jimingham (commented):
What happens if I'm in the fallback mode so I don't have a Target-wide scratch SwiftASTContext, and I stop in a module that returns false from HasSwiftModules and do:

(lldb) expr -l swift -- func $foo() { print("I wonder what this will do?") }

source/Symbol/SwiftASTContext.cpp
adrian-prantl: This is the normal behavior:
```
(lldb) expr -l swift -- func $foo() { print("I wonder what this will do?") }
(lldb) expr -l swift -- $foo
() -> () $R0 = 0x00000001012d6000
```

This is what we get in fallback mode:

```
(lldb) expr -l swift -- func $foo() { print("I wonder what this will do?") }
(lldb) expr -l swift -- $foo() 
error: Couldn't lookup symbols:
  __lldb_expr_4.$foo() -> ()
(lldb) expr -l swift -- 1
(Int) $R4 = 1
(lldb) expr -l swift -- func $foo() { print("I wonder what this will do?") }
(lldb) expr -l swift -- 1
(Int) $R6 = 1
(lldb) expr -l swift -- $foo
error: Couldn't lookup symbols:
  __lldb_expr_7.$foo() -> ()
(lldb) expr -l swift -- $R6
(Int) $R8 = 1
```

This behavior is, however, not affected by this patch. This was already broken.

PR #642, opened 2018-05-18
Add a testcase where two dylibs contain conflicting definitions in a …

adrian-prantl:
…bridging header.


adrian-prantl:
@swift-ci test and merge

PR #644, opened 2018-05-19
Improve wording of warning message.

adrian-prantl:


adrian-prantl:
@swift-ci test

source/Target/Target.cpp
```diff
@@ -2158,17 +2158,17 @@ TypeSystem *Target::GetScratchTypeSystemForLanguage(
                                     .GetFilename()
                                     .AsCString();
             if (m_use_scratch_typesystem_per_module)
-              errs->Printf(
-                  "\nnote: Swift header search options for %s conflict with "
-                  "options found in other modules. Symbols from other modules "
-                  "may not be available.",
-                  module_name);
-            else
+              errs->Printf("\nnote: Swift ClangImporter options for %s "
```
fredriss: Is ClangImporter really something we want to expose? Wouldn't it be better to just say "compiler options"?

adrian-prantl:
@swift-ci test and merge

source/Target/Target.cpp
adrian-prantl: It might actually be more accurate, too.

source/Target/Target.cpp
```diff
@@ -2158,17 +2158,17 @@ TypeSystem *Target::GetScratchTypeSystemForLanguage(
...
-              errs->Printf(
-                  "\nnote: Swift header search options for %s conflict with "
-                  "options found in other modules. Symbols from other modules "
-                  "may not be available.",
-                  module_name);
-            else
+              errs->Printf("\nnote: Swift compiler options for %s "
+                           "conflict with options found in other modules;\n"
+                           "      Switching to an isolated expression context "
+                           "for %s.\n", module_name, module_name);
```
fredriss: I'm also wondering if this is informative enough. I think we should add something like:
"..isolated expression context for %s, you won't have access to old $R variables anymore."

adrian-prantl: I wanted to strike a balance between an informative and concise message. That's why I removed the message in the curly braces below from this path. My thinking was that "isolated" implies this. If you think it's important enough to include I can back that change out again.

fredriss: I don't think people will understand what isolated means here. Maybe not even expression context. What about the less accurate but (I think more understandable):
"Switching to a new expression evaluator for %s, old $R variables are lost."

fredriss: or "are not accessible anymore"

adrian-prantl:
@swift-ci test and merge

PR #646, opened 2018-05-21
 Use a backwards-compatible way of retreiving the host os version.

adrian-prantl:
It looks like the sysctl was introduced in 10.13.

<rdar://problem/40419341>


adrian-prantl:
@swift-ci test and merge

PR #647, opened 2018-05-21
[SwiftASTContext] Fix the definition of IsFullyRealized.

dcci:
IRGen always expects a substituted type when asking for TypeInfo,
otherwise the behaviour is undefined (assertions hit/crashes).
IsFullyRealized is supposed to guard us against passing
not-fully-realized types down to the compiler, but the check was
incomplete. This fixes SR-7728.

<rdar://problem/40417941>

### DIFF
--- a/lit/SwiftREPL/SwiftTypeLookup.test
+++ b/lit/SwiftREPL/SwiftTypeLookup.test
@@ -1,0 +1,6 @@
+// Test that we don't crash when looking up types.
+
+// RUN: %lldb --repl < %s | FileCheck %s --check-prefix=TYPE
+
+:type lookup Swift
+// TYPE: Optional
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci please test

jimingham:
This seems fine in itself.  

Along with hasArchetype & hasTypeParameter, Types.h has a few other bit accessors that sound worrisome, like "hasUnresolvedType"...  Are we sure hasArchetype & hasTypeParameter are the only two that are going to cause trouble?

jckarter:
`hasUnresolvedType` should only show up in un-type-checked ASTs. Hopefully if you've gotten this far you've gotten past the type checker, so it shouldn't be a problem.

jimingham:
Davide used "type lookup Swift" as the test case for this.  IIRC he said that this path was crashing.

"type lookup" dumps raw types, and doesn't go through the type checker per se.  But it does do some munging to types (in the SwiftLanguage::SwiftTypeScavanger).  But I don't think that amounts to a full-on type check...

dcci:
@jimingham thanks. I don't think we can get here with a type for which `hasUnresolvedType()` is true. [I tried to put an unreachable and run the testsuite and other tests]. I'm going to audit all the flags and see if we need something else as a follow-up (also, try to come up with test cases if we need it).

Also, thanks to @jckarter for discussing the fix with me.

PR #648, opened 2018-05-21
 Add SDK fallbacks for AppleTV and watchOS.

adrian-prantl:
<rdar://problem/37255462>


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
This is the fallback path for when the serialized SDK path does not exist on the machine running LLDB.

jasonmolenda (commented):
looks good, I was a little confused by the is_simulator lambda but I see that it's only used for the ios/watchos/tvos cases.

PR #649, opened 2018-05-22
Work around some odd instruction single-step behavior on macOS.

jimingham:
We've seen some cases on macOS where you go to instruction single
step (over a breakpoint), and single step returns but the instruction
hasn't been executed (and the pc hasn't moved.)  The ThreadPlanStepOverBreakpoint
used to handle this case by accident, but the patches to handle two adjacent
breakpoints broke that accident.

This patch fixes the logic of ExplainsStop to explicitly handle the case where
the pc didn't move.  It also adds a WillPop that re-enables the breakpoint we
were stepping over.  We never want an unexpected path through the plan to
fool us into not doing that.

I have no idea how to make this bug happen.  It is very inconsistent when it
occurs IRL.  We really need a full MockProcess Plugin before we can start to write
tests for this sort of system hiccup.

<rdar://problem/38505726>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@332922 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/source/Target/ThreadPlanStepOverBreakpoint.cpp
+++ b/source/Target/ThreadPlanStepOverBreakpoint.cpp
@@ -69,30 +69,47 @@ bool ThreadPlanStepOverBreakpoint::DoPlanExplainsStop(Event *event_ptr) {
     // contained a breakpoint.
     StopReason reason = stop_info_sp->GetStopReason();
 
+    Log *log(lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_STEP));
+
+    if (log)
+      log->Printf("Step over breakpoint stopped for reason: %s.", 
+          Thread::StopReasonAsCString(reason));
+  
     switch (reason) {
-    case eStopReasonTrace:
-    case eStopReasonNone:
-      return true;
-    case eStopReasonBreakpoint:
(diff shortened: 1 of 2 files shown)


source/Target/ThreadPlanStepOverBreakpoint.cpp
```diff
@@ -69,30 +69,47 @@ bool ThreadPlanStepOverBreakpoint::DoPlanExplainsStop(Event *event_ptr) {
...
+        // But make sure we don't set ourselves to auto-continue or we'll wrench
+        // control away from the plans that can deal with this.
+        // Be careful, however, as we may have "seen a breakpoint under the PC
+        // because we stopped without changing the PC, in which case we do want
+        // to re-claim this stop so we'll try again.
+        lldb::addr_t pc_addr = m_thread.GetRegisterContext()->GetPC();
+
+        if (pc_addr == m_breakpoint_addr) {
+          if (log)
+            log->Printf("Got breakpoint stop reason but pc: %" PRIu64
```
jasonmolenda, line 103: This should log in hex, but pc: 0x%" PRIx64

jimingham:
@swift-ci please test

jasonmolenda:
Looks good to me.

jimingham:

PR #650, opened 2018-05-22
 Avoid building test on Linux that isn't run there. 

adrian-prantl:
rdar://problem/40437419


adrian-prantl:
@swift-ci test and merge

PR #653, opened 2018-05-23
[SwiftASTContext] Call collectLinkLibraries for all the modules.

dcci:
Unbreaks `import Cocoa`.

<rdar://problem/40458863>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -3551,7 +3551,11 @@ void SwiftASTContext::LoadModule(swift::ModuleDecl *swift_module,
         all_dlopen_errors.GetData());
   };
 
-  swift_module->collectLinkLibraries(addLinkLibrary);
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3551,7 +3551,11 @@ void SwiftASTContext::LoadModule(swift::ModuleDecl *swift_module,
         all_dlopen_errors.GetData());
   };
 
-  swift_module->collectLinkLibraries(addLinkLibrary);
+  swift_module->forAllVisibleModules(
+      {}, [&](swift::ModuleDecl::ImportedModule import) {
+        import.second->collectLinkLibraries(addLinkLibrary);
+        return true;
+      });
```
jrose-apple, line 3558: Ahh…the correct fix is to call `collectLinkLibraries` on the SourceFile being compiled, not the module. Sorry for our mistake!

PR #655, opened 2018-05-23
[SwiftASTContext] Fix the definition of IsFullyRealized.

dcci:
IRGen always expects a substituted type when asking for TypeInfo,
otherwise the behaviour is undefined (assertions hit/crashes).
IsFullyRealized is supposed to guard us against passing
not-fully-realized types down to the compiler, but the check was
incomplete. This fixes SR-7728.

<rdar://problem/40417941>


dcci:
@swift-ci please test

PR #657, opened 2018-05-23
[SwiftASTContext] Call collectLinkLibraries for all the modules.

dcci:
Unbreaks `import Cocoa`.

<rdar://problem/40458863>


dcci:
@swift-ci please test

PR #658, opened 2018-05-23
Destroy RemoteASTContexts before associated swift::ASTContexts

vedantk:
This fixes a use-after-free caught by assertions in swift (and by ASan).

A RemoteASTContext must be destroyed before its swift::ASTContext can be
destroyed: otherwise, its destructor will try to operate on freed
memory.

Thanks to John McCall for pointing this out!

rdar://40491481


vedantk:
@swift-ci test

rjmccall (approved):
LGTM.

dcci (approved):

PR #659, opened 2018-05-24
Destroy RemoteASTContexts before associated swift::ASTContexts (4.2)

vedantk:
This fixes a use-after-free caught by assertions in swift (and by ASan).

A RemoteASTContext must be destroyed before its swift::ASTContext can be
destroyed: otherwise, its destructor will try to operate on freed
memory.

Thanks to John McCall for pointing this out!

rdar://40491481
(cherry picked from commit 22c70c376dbff8c0cf606bdb082a8b47bc2ebff5)


vedantk:
This should go into 4.2, not stable (again).

PR #664, opened 2018-05-30
[ObjC] Fix the formatter for NSOrderedSet.

dcci:
<rdar://problem/40622096>


dcci:
@swift-ci please test

vedantk (approved):
LGTM. As a follow-up, it would be nice to determine whether we're missing coverage for NSCountedSet. Based on the deleted code, it looks like this coverage could be missing.

Other than that, it'd also be nice to find some way to reduce repetition of hardcoded mangled names. That would be fine as a follow-up too.

PR #665, opened 2018-05-30
[FileSpec] Re-implmenet removeLastPathComponent

JDevlieghere:
When reading DBGSourcePathRemapping from a dSYM, we remove the last two
path components to make the source lookup more general. However, when
dealing with a relative path that has less than 2 components, we ended
up with an invalid (empty) FileSpec.

This patch changes the behavior of removeLastPathComponent to remove the
last path component, if possible. It does this by checking whether a
parent path exists, and if so using that as the new path. We rely
entirely on LLVM's path implementation to do the heavy lifting.

We now also return a boolean which indicates whether the operator was
successful or not.

Differential revision: https://reviews.llvm.org/D47495

rdar://37791687

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@333540 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 23454e85f0ef4916fc9282deddcbaa378d06c62d)

### DIFF
--- a/include/lldb/Utility/FileSpec.h
+++ b/include/lldb/Utility/FileSpec.h
@@ -549,7 +549,14 @@ public:
   void AppendPathComponent(llvm::StringRef component);
   void AppendPathComponent(const FileSpec &new_path);
 
-  void RemoveLastPathComponent();
+  //------------------------------------------------------------------
+  /// Removes the last path component by replacing the current path with its
+  /// parent. When the current path has no parent, this is a no-op.
+  ///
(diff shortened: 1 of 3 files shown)


JDevlieghere:

JDevlieghere:
Looks like m_style is not available yet.

JDevlieghere:

dcci:
After this change all the bots started timing out, so I'm going to roll back this.

PR #666, opened 2018-05-31
Restore the old remoteLastPathComponent behavior

dcci:
This basically reverts the last two commits on FileSpec as they caused bots timeout.
Jonas, if you want to look closely, the logs are available (TestMiniDump* was the one failing, which we, unfortunately, don't run as part of the PR testing cycle).


dcci:
@swift-ci test and merge

PR #671, opened 2018-06-01
Adapt Swift type metadata lookup to τ_depth_index convention.

adrian-prantl:
This is the LLDB work to adapt the new compiler convention to use
depth and index instead of name to lookup type metadata artificial
variables.

<rdar://problem/38306926>


adrian-prantl:
test with https://github.com/apple/swift/pull/16937
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/16937
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/16937
@swift-ci test

source/Plugins/Language/Swift/SwiftLanguage.cpp
```diff
@@ -872,7 +872,7 @@ SwiftLanguage::GetHardcodedSummaries() {
         return swift_metatype_summary_sp;
       }
       if (valobj.GetName().GetLength() > 12 &&
-          valobj.GetName().GetStringRef().startswith("$swift.type.") &&
+          valobj.GetName().GetStringRef().startswith(u8"$\u03C4_") &&
```
dcci: Would be nice to store the unicode string to a variable, or add a comment for posterity

source/Plugins/SymbolFile/DWARF/DWARFASTParserSwift.cpp
```diff
@@ -119,8 +119,11 @@ lldb::TypeSP DWARFASTParserSwift::ParseTypeFromDWARF(const SymbolContext &sc,
   }
 
   if (!compiler_type && name) {
-    if (name.GetStringRef().startswith("$swift.") ||
-        name.GetStringRef().startswith(SwiftLanguageRuntime::GetCurrentMangledName("_TtBp").c_str())) {  // This is the RawPointerType, need to figure out its name from the AST.
+    if (name.GetStringRef().startswith("$") ||
+        name.GetStringRef().startswith(
+            SwiftLanguageRuntime::GetCurrentMangledName("_TtBp")
```
dcci, line 124: Should we switch to the new mangled name here? 
Also, isn't the second condition made redundant by the first? All the mangled names now start with `$S`

source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
```diff
@@ -320,10 +326,16 @@ void SwiftUserExpression::ScanContext(ExecutionContext &exe_ctx, Status &err) {
               log->Printf("    [SUE::SC] Argument name: %s",
                           template_arg_name.AsCString());
 
-            CompilerType concrete_type =
-                exe_ctx.GetProcessRef()
-                    .GetSwiftLanguageRuntime()
-                    ->GetConcreteType(frame, template_arg_name);
+            CompilerType concrete_type = template_arg_type;
```
dcci: This and the snippet above (line 295) are exactly the same. Can we factor the logic out in a common function?

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -1262,6 +1277,9 @@ MaterializeVariable(SwiftASTManipulatorBase::VariableInfo &variable,
     VariableMetadataVariable *variable_metadata =
         static_cast<VariableMetadataVariable *>(variable.m_metadata.get());
 
+    // FIXME: It would be nice if we could do something like
```
dcci, line 1284: Why we can't?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2406,12 +2430,15 @@ Value::ValueType SwiftLanguageRuntime::GetValueType(
...
-        static_type_flags.AllClear(eTypeIsArchetype)) {
+        static_type_flags.AllClear(eTypeIsGenericTypeParam)) {
+      // FIXME: This branch is not covered by any testcases in the test suite.
       if (is_indirect_enum_case || static_type_flags.AllClear(eTypeIsBuiltIn))
         return Value::eValueTypeLoadAddress;
     }
   }
 
+  // Enabling this makes the inout_variables test hang.
+  //  return Value::eValueTypeScalar;
```
dcci, line 2431: Can you elaborate this one? Is this dead code? if so, can we remove it?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2406,12 +2430,15 @@ Value::ValueType SwiftLanguageRuntime::GetValueType(
     }
 
     if (static_type_flags.AllSet(eTypeIsSwift | eTypeIsPointer) &&
-        static_type_flags.AllClear(eTypeIsArchetype)) {
+        static_type_flags.AllClear(eTypeIsGenericTypeParam)) {
+      // FIXME: This branch is not covered by any testcases in the test suite.
```
dcci, line 2424: Then should we remove it or add a test?

dcci (changes requested):
Thanks for the great piece of work. I have some comments inline.

source/Plugins/SymbolFile/DWARF/DWARFASTParserSwift.cpp
adrian-prantl: It's possible that this is just dead code.

AFAIK all remaining occurrences of _T types in LLDB are hard-coded typenames in the Objective-C Foundation framework that are reserved for Swifty interfaces vended by various of Foundation classes. In this particular case though, I am not so sure.

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: Yes. This is just something I noticed while working on this patch. I'll put an unreachable there and see if it sticks.

adrian-prantl:
test with https://github.com/apple/swift/pull/16937
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/16937
@swift-ci test

adrian-prantl:
@dcci: Can you take another look?

adrian-prantl:
@dcci: Can you take another look?

dcci:
This looks good to me but we should really add a test for type reconstruction in the compiler.

adrian-prantl:
I did a new type reconstruction test over in the last revision of apple/swift#16937.

PR #674, opened 2018-06-01
[SwiftLanguageRuntime] The witness table location is not really used.

dcci:


dcci:

jimingham:
I think you should be able to remove the computation of witness_table_ptr_value as well, since the only thing that gets used for is to throw an error if we can't get its value, which doesn't seem terribly useful.

PR #675, opened 2018-06-02
Fix Module::FindTypes to return the correct number of matches.

fredriss:
In r331719, I changed Module::FindTypes not to limit the amount
of types returned by the Symbol provider, because we want all
possible matches to be able to filter them. In one code path,
the filtering was applied to the TypeList without changing the
number of types that gets returned. This is turn could cause
consumers to access beyond the end of the TypeList.

This patch fixes this case and also adds an assertion to
TypeList::GetTypeAtIndex to catch those obvious programming
mistakes.

Triggering the condition in which we performed the incorrect
access was not easy. It happened a lot in mixed Swift/ObjectiveC
code, but I was able to trigger it in pure Objective C++ although
in a contrieved way.

rdar://problem/40254997

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@333786 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 6dc827b19e54e64786ed1a304516e5fe16f96db0)

### DIFF
--- a/packages/Python/lldbsuite/test/lang/objcxx/class-name-clash/Makefile
+++ b/packages/Python/lldbsuite/test/lang/objcxx/class-name-clash/Makefile
@@ -1,0 +1,7 @@
+LEVEL = ../../../make
+OBJCXX_SOURCES := main.mm myobject.mm
+include $(LEVEL)/Makefile.rules
+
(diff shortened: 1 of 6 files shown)


fredriss:
@swift-ci test and merge

PR #676, opened 2018-06-04
[SwiftLanguageRuntime] Simplify Error dynamic type lookup.

dcci:
We don't really need the witness table pointer, so don't bother
computing it.


dcci:
@swift-ci test and merge

PR #677, opened 2018-06-05
[SwiftLanguageRuntime] Refactor archetype(s) dynamic lookup.

dcci:
-> Use IsStaticallyDetermined().
-> Early returns for better readability.
-> Add an unreacahble on a condition that should never happen
(and silence a compiler warning).


dcci:
@swift-ci test and merge

PR #678, opened 2018-06-05
[SwiftLanguageRuntime] Switch archetype resolution to use RemoteAST.

dcci:
Another step towards removing our homegrown dynamic lookup logic now
that we have infra in the compiler to fix this.


dcci:
@swift-ci please test

PR #679, opened 2018-06-06
[SwiftLanguageRuntime] Simplify existential dynamic lookup.

dcci:
Nobody's using this information now that we switched to RemoteAST.


dcci:
@swift-ci please test

PR #680, opened 2018-06-07
Rollback switch archetype resolution to use RemoteAST.

dcci:
This breaks some corner cases not really covered by the testsuite.
I'm going to add a test next, and then try to debug what went
wrong here.


dcci:
@swift-ci test and merge

PR #681, opened 2018-06-11
Fix MainThreadCheckerRuntime.cpp to recognize properties from functions.

kubamracek:
Fix MainThreadCheckerRuntime.cpp to recognize properties from functions.

<rdar://problem/36557167>

### DIFF
--- a/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
+++ b/source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
@@ -112,6 +112,7 @@ static std::string TranslateObjCNameToSwiftName(std::string className,
       swift::ClassDecl *cls = llvm::dyn_cast<swift::ClassDecl>(VD);
       if (!cls)
         return;
+      cls->loadAllMembers();
       auto funcs = cls->lookupDirect(selectorToLookup, true);
       if (funcs.size() == 0)
         return;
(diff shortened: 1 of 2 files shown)


kubamracek:
@swift-ci please test

source/Plugins/InstrumentationRuntime/MainThreadChecker/MainThreadCheckerRuntime.cpp
```diff
@@ -112,6 +112,7 @@ static std::string TranslateObjCNameToSwiftName(std::string className,
       swift::ClassDecl *cls = llvm::dyn_cast<swift::ClassDecl>(VD);
       if (!cls)
         return;
+      cls->loadAllMembers();
```
jrose-apple: This should not be necessary; `lookupDirect` should handle it. If that's not the case we should change something in Swift, not in MainThreadChecker.

kubamracek: Yes, it is not the case. The lookup started returning wrong results after <https://github.com/apple/swift/pull/13866> landed. Do you have a sense of what's going on?

jrose-apple: Not obviously, but maybe there's a problem deserializing an AccessorDecl without its accompanying property? (See ModuleFile::loadObjCMethods.)

kubamracek:
https://github.com/apple/swift/pull/17150

@swift-ci Please test

kubamracek:
apple/swift#17150

@swift-ci Please test

kubamracek:
@jrose-apple so it this okay now?

jrose-apple:
Yep! I mean, now you're just throwing out the unnecessary code. :-)

kubamracek:
apple/swift#17193

@swift-ci Please test

PR #682, opened 2018-06-11
[SwiftLanguageRuntime] Switch archetype resolution to RemoteAST.

dcci:
While here, add a test to cover a case that regressed (and wasn't
tested).


dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

PR #684, opened 2018-06-12
[Swift] Fix dynamic type discovery for classes adhering to `Error`.

dcci:
This replaces the handmade computation of metadata and dynamic
type with a call to fullFillTypePromise(), which uses RemoteAST
for the resolution (apparently, correctly).

I added a test as this fixes a crash.

Fixes <rdar://problem/41025993>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/class_error/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/class_error/main.swift
@@ -1,0 +1,31 @@
+// Let's try to make sure that frame var prints the dynamic type of a class
+// (and a subclass) conforming to Error correctly (and that we don't crash).
+// This involves resolving the dynamic type correctly in the language runtime.
+
+class MyErr : Error {
+  var x : Int
+
+  init(_ x : Int) {
+    self.x = x
+  }
+}
+
+class MyOtherErr : MyErr {}
+
+func f<T>(_ Pat : T) -> T {
+  return Pat //%self.expect('frame variable -d run -- Pat', substrs=['MyErr', 'x = 23'])
+}
+
+func g<T>(_ Pat : T) -> T {
+  return Pat //%self.expect('frame variable -d run -- Pat', substrs=['MyOtherErr', 'x = 42'])
+}
+
+func main() -> Int {
+  let foo : MyErr = MyErr(23)
+  let goo : MyErr = MyOtherErr(42)
+  let patatino = f(foo)
+  let tinky = g(goo)
+  return 0
+}
+
+let _ = main()
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1953,41 +1953,23 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Promise(
         llvm::dyn_cast_or_null<SwiftASTContext>(var_type.GetTypeSystem());
 
     CompilerType protocol_type(promise_sp->FulfillTypePromise());
+    // Error has a special, NSError-compatible layout.
     if (swift_ast_ctx->IsErrorType(protocol_type)) {
       if (swift_ast_ctx) {
         Status error;
-        // the offset
-        size_t ptr_size = m_process->GetAddressByteSize();
-        size_t metadata_offset = ptr_size + 4 + (ptr_size == 8 ? 4 : 0);
-        metadata_offset += ptr_size + ptr_size + ptr_size;
         lldb::addr_t archetype_ptr_value = in_value.GetValueAsUnsigned(0);
-        lldb::addr_t base_errortype_ptr =
+        lldb::addr_t metadata_ptr =
             m_process->ReadPointerFromMemory(archetype_ptr_value, error);
-        lldb::addr_t static_metadata_ptrptr =
-            base_errortype_ptr + metadata_offset;
-        lldb::addr_t static_metadata_ptr =
-            m_process->ReadPointerFromMemory(static_metadata_ptrptr, error);
         MetadataPromiseSP promise_sp(
-            GetMetadataPromise(static_metadata_ptr, swift_ast_ctx));
-        if (promise_sp) {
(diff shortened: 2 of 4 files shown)


dcci:
@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/class_error/main.swift
```diff
@@ -0,0 +1,31 @@
+// Let's try to make sure that frame var prints the dynamic type of a class
+// (and a subclass) conforming to Error correctly (and that we don't crash).
+// <rdar://problem/41025993>
```
adrian-prantl: Let's add the radar number only in the commit message and be sure to include as much explanation in the comment as needed.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1956,38 +1956,19 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Promise(
     if (swift_ast_ctx->IsErrorType(protocol_type)) {
```
adrian-prantl, line 1957: Out of curiosity: Do you understand why error is handled specially here? I'd be curious to learn why we can't handle all existential types the same way.

packages/Python/lldbsuite/test/lang/swift/class_error/main.swift
```diff
@@ -0,0 +1,31 @@
...
+
+func g<T>(_ Pat : T) -> T {
+  return Pat //%self.expect('frame variable -d run -- Pat', substrs=['MyOtherErr', 'x = 42'])
+}
+
+func main() -> Int {
+  let foo : MyErr = MyErr(23)
+  let goo : MyErr = MyOtherErr(42)
+  let patatino = f(foo)
+  let tinky = g(goo)
```
adrian-prantl, line 27: It's possible that this won't work because of an unrelated bug, but
```
func h<U, V>(_ tuple : (U, V)) -> (U, V) {
  return tuple //%self.expect('frame variable -d run -- Pat', substrs=['MyErr', 'x = 23', 'MyOtherErr', 'x = 42'])
}

h((foo, goo))
```
would be an interesting testcase, too.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1956,38 +1956,19 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Promise(
...
-            address.SetLoadAddress(load_addr, &m_process->GetTarget());
-            return true;
+            GetMetadataPromise(metadata_ptr, swift_ast_ctx));
+        if (!promise_sp)
+          return false;
+        CompilerType clang_type(promise_sp->FulfillTypePromise());
+        if (clang_type.IsValid()) {
+          class_type_or_name.SetCompilerType(clang_type);
+          address.SetLoadAddress(archetype_ptr_value, &m_process->GetTarget());
+          return true;
```
adrian-prantl: I wish I understood better what exactly the old code was doing, but the new code definitely looks *safe*.

source/Target/SwiftLanguageRuntime.cpp
jrose-apple: Error has a special, NSError-compatible layout.

adrian-prantl: Thanks, that makes sense!

adrian-prantl: It might be worth looking at the LLVM IR generated for the testcase on various platforms then, to check whether any of the contortions performed by the old code perhaps make sense on any other targets (like i386 for example).

adrian-prantl, line 1957: Also: let's add `// Error has a special, NSError-compatible layout.` here :-)

jrose-apple: Ah, it's worth double-checking whether that's always true or only on Darwin, too!

packages/Python/lldbsuite/test/lang/swift/class_error/main.swift
dcci: This is failing with 

```
(lldb) frame var tuple
(U, V) tuple = <Unable to determine byte size.>
```
which I think is unrelated to my patch (it actually fails before getting to the resolution).

source/Target/SwiftLanguageRuntime.cpp
dcci: Checked, it looks fine.

dcci: Will do!

dcci (commented):
Jim came up with another testcase which doesn't work (but it didn't work before the patch either), so I'll leave it as a follow up. I'm going to update the patch to address your comments. Thanks for the review!

dcci:
@adrian-prantl updated, can you take another look?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1953,41 +1953,23 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Promise(
...
-          if (clang_type.IsValid() && load_addr != 0 &&
-              load_addr != LLDB_INVALID_ADDRESS) {
-            class_type_or_name.SetCompilerType(clang_type);
-            address.SetLoadAddress(load_addr, &m_process->GetTarget());
-            return true;
+            GetMetadataPromise(metadata_ptr, swift_ast_ctx));
+        if (!promise_sp)
+          return false;
+        CompilerType dynamic_type(promise_sp->FulfillTypePromise());
+        if (dynamic_type.IsValid()) {
```
adrian-prantl, line 1968: What happens when !IsValid? Is it intentional that we don't return false in that case?

dcci: if `!IsValid`, then we just fall off the `if` statement, break, and return `false` at the end of the function, which I think is what we want.

adrian-prantl (approved):

dcci:
@swift-ci test and merge

PR #687, opened 2018-06-13
[Platform] Accept arbitrary kext variants

JDevlieghere:
When loading kexts in PlatformDarwinKernel, we use the BundleID as the
filename to to create shared modules. In GetSharedModule we call
ExamineKextForMatchingUUID for any BundleID it finds that is a match, to
see if the UUID is also a match. Until now we were using
Host::ResolveExecutableInBundle which calls a CoreFoundation API to
obtain the executable. However, it's possible that the executable has a
variant suffix (e.g. foo_development) and these files were ignored.

This patch replaces that call with logic that looks for all the binaries
in the bundle. Because of the way ExamineKextForMatchingUUID works, it's
fine to try to load executables that are not valid and we can just
iterate over the list until we found a match.

Differential revision: https://reviews.llvm.org/D47539

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@334205 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 209989753af17fcc18d1769d2f68be45d2a16feb)

### DIFF
--- a/source/Plugins/Platform/MacOSX/PlatformDarwinKernel.cpp
+++ b/source/Plugins/Platform/MacOSX/PlatformDarwinKernel.cpp
@@ -694,7 +694,7 @@ Status PlatformDarwinKernel::GetSharedModule(
       }
     }
 
-    // Give the generic methods, including possibly calling into 
+    // Give the generic methods, including possibly calling into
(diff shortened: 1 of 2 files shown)


JDevlieghere:

PR #688, opened 2018-06-13
[ObjC] Add dataformatter for NSDecimalNumber

JDevlieghere:
This patch adds a data formatter for NSDecimalNumber. The latter is a
Foundation object used for representing and performing arithmetic on
base-10 numbers that bridges to Decimal.


JDevlieghere:

PR #693, opened 2018-06-15
Fix TestExec on macOS 10.14

fredriss:


fredriss:
@swift-ci test and merge

PR #696, opened 2018-06-16
[Swift] Fix dynamic type discovery for classes adhering to `Error`.

dcci:
This replaces the handmade computation of metadata and dynamic
type with a call to fullFillTypePromise(), which uses RemoteAST
for the resolution (apparently, correctly).

I added a test as this fixes a crash.

Fixes <rdar://problem/41025993>


dcci:
@swift-ci please test

PR #700, opened 2018-06-19
[Swift] Test that we print an array of Any.Type(s) correctly.

dcci:
This was a bug in type reconstruction, but I'm adding a test on
the lldb side to make sure this doesn't regress here.


dcci:
@swift-ci test and merge

PR #701, opened 2018-06-19
Fix the 'tb' alias command

fredriss:
No idea when this broke or if it ever worked. Added a small test
for one-shot breakpoints while I was there.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@334921 91177308-0d34-0410-b5e6-96231b3b80d8


fredriss:
@swift-ci test and merge

PR #706, opened 2018-06-21
Use the scratch Swift AST context to resolve dynamic types.

adrian-prantl:
Dynamic type resolution in RemoteAST might pull in other Swift
modules, so use the scratch context where such operations are legal
and safe.

<rdar://problem/40950542>

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1462,8 +1462,9 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Class(
     return false;
   address.SetRawAddress(class_metadata_ptr);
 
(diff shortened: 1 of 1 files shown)


adrian-prantl:
test with https://github.com/apple/swift/pull/17384
@swift-ci test

dcci (commented):
I think there's at least another place where we use RemoteAST, and that's where we get the offset of member variables.
```
GetMemberVariableOffset
```

It looks like we don't use the scratch AST context there, should we?

dcci:
Also, maybe as a follow-up, but we should consider adding the tests that are failing as `@swift-pr`, so that we catch other problems in the future. Thanks!

adrian-prantl:
I considered that, but GetMemberVariableOffset does not do any form of dynamic type resolution, so my thinking was that it would not run into this situation.

PR #715, opened 2018-06-23
 Make Objective-C types work in frame variable when debugging without...

adrian-prantl:
…a .swiftmodule.

When we failed to look up the type because no .swiftmodule is
present or it couldn't be read, fall back to presenting objects
that look like they might be come from Objective-C as Clang
types. LLDB's Objective-C part is very robust against malformed
object pointers, so this isn't very risky.

rdar://problem/41370330


adrian-prantl:
@swift-ci test and merge

PR #716, opened 2018-06-25
[SwiftASTContext] Stopgap solution to avoid crashing when printing ge…

dcci:
…neric extensions.

The proper fix requires backporting a bunch of stuffs that's not really
worth it. `stable` contains a working version.

<rdar://problem/41343572>


dcci:
@swift-ci please test

PR #718, opened 2018-06-25
[Swift] Link against the swift reflection library.

dcci:
This is the first step to integrate RemoteMirrors in lldb.

<rdar://problem/36477650>

### DIFF
--- a/source/Target/CMakeLists.txt
+++ b/source/Target/CMakeLists.txt
@@ -1,5 +1,13 @@
 include_directories(../Plugins/Process/Utility)
 
+# libswiftReflection is placed in a subdirectory which depends
+# on the host we're currently running on.
+# FIXME: There must be a better way to get at libswiftReflection.a
+string(REPLACE "-" ";" TRIPLE_LIST "${LLVM_HOST_TRIPLE}")
+list(GET TRIPLE_LIST 0 ARCH)
+list(GET TRIPLE_LIST 2 HOST_OS_VERSION)
+string(REGEX MATCHALL "[a-z]+" HOST_OS "${HOST_OS_VERSION}")
+
 add_lldb_library(lldbTarget
   ABI.cpp
   CPPLanguageRuntime.cpp
@@ -67,6 +75,7 @@ add_lldb_library(lldbTarget
     swiftBasic
     swiftFrontend
     swiftRemoteAST
+    swiftReflection-${HOST_OS}-${ARCH}
     lldbBreakpoint
     lldbCore
     lldbExpression


dcci:
@jrose-apple / @gottesmm do you know a better way of getting to the library in the cmake build system? 

@jimingham I need to hook this also up in the Xcode build system. What's the way of telling xcodebuild to link against a library?

adrian-prantl (approved):
We discussed this extensively before :-)

jrose-apple:
You don't actually need the architecture, just the platform. I would expect you'd be able to use the CMake target name, `swiftReflection-PLATFORM`, but maybe that doesn't work.

dcci:
@jrose-apple that doesn't seem to work. I noticed that the unittests use
```
swiftReflection${SWIFT_PRIMARY_VARIANT_SUFFIX}
```

but that doesn't work either.

dcci:

gottesmm:
This is the wrong way to do this. You should be using the export that you guys are already using from cmake.

dcci:
force stopped the merge to discuss this with @gottesmm who's going to propose a better approach of handling this.

gottesmm:
Thanks!

gottesmm:
(for those listening in, I am going to show Davide how to do it tomorrow when I am in the office).

dcci:
Will update this once the cmake bits land: https://github.com/apple/swift/pull/17525

gottesmm:
CMake bits landed and everything looks good.

dcci:
@swift-ci test and merge

dcci:
(thanks @gottesmm for the help with the cmake bits on the swift side).

dcci:
@gottesmm this is still broken for Linux. I'll take a closer look. Thanks for your help!

source/Target/CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
 include_directories(../Plugins/Process/Utility)
 
+# libswiftReflection is placed in a subdirectory which depends
+# on the host we're currently running on.
+# FIXME: There must be a better way to get at libswiftReflection.a
+string(REPLACE "-" ";" TRIPLE_LIST "${LLVM_HOST_TRIPLE}")
```
Rostepher, line 6: The error message in the failing bot seems to indicate that `TRIPLE_LIST` is empty when the next line (`list(GET...)`) is called, but only on Linux.

dcci: The issue is more complicated than that.
Linux doesn't really include the arch when building.

```
davide@rad:~/work/build$ find . -name "libswiftReflection*"
./Ninja-ReleaseAssert/swift-linux-x86_64/lib/swift/linux/x86_64/libswiftReflection.a
./Ninja-ReleaseAssert/swift-linux-x86_64/lib/swift/linux/libswiftReflection.a
```

Even though I change `CmakeLists.txt` to just link `swiftReflection`, the library is not found, as it's not exported correctly.

dcci:
This is obsolete now.

PR #719, opened 2018-06-26
 Remove Foundation dependency from Swift testcase.

adrian-prantl:
rdar://problem/41444994


adrian-prantl:
@swift-ci test and merge

PR #720, opened 2018-06-26
[Swift] Add a test for printing a generic function in an extension

dcci:
While I'm here, I deleted some dead code.


dcci:
@swift-ci test and merge

PR #723, opened 2018-06-28
Avoid initializing the Swift IRGenModule for a given AST context concurrently

fredriss:
rdar://problem/40290306


fredriss:
@swift-ci test and merge

PR #724, opened 2018-06-28
Weaken test for generic typealias demangling.

DougGregor:
As part of rdar://problem/41444286, we no longer produce generic
typealias mangling in the compiler. Weaken this test so it works
either with or without generic typealias mangling.


DougGregor:

dcci:
lg, thanks.

PR #726, opened 2018-06-28
Fixed file completion for paths that start with '~'

Teemperor:
We didn't add the remaining path behind the '~' to the completion string,
causing it to just complete directories inside the user home directory. This
patch just adds the directory of the remaining path if there is one.

Fixes rdar://problem/40147002

Original lldb-commit r334978. Includes the test case from r335905.


dcci:
@swift-ci please test

dcci:
@swift-ci please merge

PR #731, opened 2018-07-02
patch to get latest tensorflow lldb to compile in google's piper

marcrasi:
@allevato previously patched some `const char *`s to be `std::string`s to get lldb to compile in google's internal builds (https://github.com/apple/swift-lldb/commit/5db6779b98764451e31bafa5ee9b475a7f95e5e4). A new `const char *` has appeared, so here's a patch to make it into a `std::string`.


dan-zheng (approved):
👍

PR #732, opened 2018-07-02
[SwiftASTContext] Register the request functions when creating a cont…

dcci:
…ext.

Not doing that results in the state not properly set up, and an
assertion hit in the compiler. No test as several are already failing
on the bots, so we have coverage.

<rdar://problem/41738966>


dcci:

dcci:
@DougGregor could you please take a look and let me know whether this makes sense to you?

DougGregor (approved):
LGTM, thank you!

PR #737, opened 2018-07-03
 Disable the EnableTargetOSChecking language option for all Swift AST

adrian-prantl:
…contexts.

LLDB started to understand the minimum deployment target embedded in
the Mach-O load commands in r332067. This makes it more likely to hit
the module-too-new warning in a multi-dylib project where any of the
dylibs have a newer minimum deployment target than the main
executable. Since this is only a warning emitted by the serialized
module loader, there isn't much to be gained to enable this warning as
error in LLDB. This patch disables it.

rdar://problem/41652444


adrian-prantl:
@swift-ci test and merge

PR #739, opened 2018-07-04
Disable the EnableTargetOSChecking language option for all Swift AST …

adrian-prantl:
…contexts.

LLDB started to understand the minimum deployment target embedded in
the Mach-O load commands in r332067. This makes it more likely to hit
the module-too-new warning in a multi-dylib project where any of the
dylibs have a newer minimum deployment target than the main
executable. Since this is only a warning emitted by the serialized
module loader, there isn't much to be gained to enable this warning as
error in LLDB. This patch disables it.

rdar://problem/41652444
(cherry picked from commit a72289ceb4f23d6c7e7b667b70cb84eac4dffa61)

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/deployment_target/TestSwiftDeploymentTarget.py
+++ b/packages/Python/lldbsuite/test/lang/swift/deployment_target/TestSwiftDeploymentTarget.py
@@ -1,0 +1,54 @@
+# TestSwiftDeploymentTarget.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2018 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+
+import lldbsuite.test.lldbinline as lldbinline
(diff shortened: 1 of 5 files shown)


adrian-prantl:
@swift-ci test and merge

fredriss (approved):

packages/Python/lldbsuite/test/lang/swift/deployment_target/TestSwiftDeploymentTarget.py
```diff
@@ -0,0 +1,54 @@
...
+    mydir = TestBase.compute_mydir(__file__)
+
+    @decorators.skipUnlessDarwin
+    @decorators.skipIf(macos_version=["<", "10.11"])
+    @decorators.swiftTest
+    @decorators.add_test_categories(["swiftpr"])
+    def test_swift_deployment_target(self):
+        self.build()
+        # Create the target
+        target = self.dbg.CreateTarget(self.getBuildArtifact("a.out"))
```
jimingham, line 33: You can replace the code from line 32 to line 51 with:

`(_, _, self.thread, _) = lldbutils.run_to_source_breakpoint('break here', lldb.SBFileSpec('main.swift'))`

jimingham (approved):
It would be good to use the packaged form of the startup sequence for dotest tests.  At some point, we'll get all of them working this way, then any global changes we need to make to startup can all go in one place.

PR #741, opened 2018-07-04
[Swift] Fix new test for review feedback [4.2]

slavapestov:
* Description: This is a follow-up to https://github.com/apple/swift-lldb/pull/738. @adrian-prantl pointed out that the test was using an idiom that would prevent it from failing if the wrong output was produced.

* Risk: None, this PR just updates the test.

* Reviewed by: @adrian-prantl


slavapestov:
@swift-ci Please test

adrian-prantl:
Thanks!

PR #742, opened 2018-07-04
Use in-tree dsymutil for running the test suite.

JDevlieghere:


JDevlieghere:

davezarzycki:
Hi @JDevlieghere – This broke unified builds on Linux. Was that intentional?
```
CMake Error at tools/lldb/test/CMakeLists.txt:169 (add_dependencies):
  The dependency target "dsymutil" of target "lldb-dotest" does not exist.
```

PR #744, opened 2018-07-05
Fix TestLoadUsingPaths on linux

jimingham:
we need to explicitly link the test program with -ldl for the dlopen
function to be available.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@335956 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #747, opened 2018-07-06
[TestLoadUsingPaths] Forward port a couple of fixes from 4.2 to stable

dcci:
These appear to be on 4.2 but not on stable as they haven't been auto-merged.
@jimingham as the bots are red, I'm going ahead and merge this, as it should. be a no-brainer, but I would appreciate if you can take a look and ACK the back(forward?)port from 4.2 to stable.

<rdar://problem/41897465>


dcci:

PR #750, opened 2018-07-07
Use "Swift.Optional" to wrap expressions in weak self context.

jimingham:
We were using Optional, but that could collide with other uses.  So make
it explicit in our wrapper.

<rdar://problem/41525749>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/expression/optional_amibiguity/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/expression/optional_amibiguity/Makefile
@@ -1,0 +1,15 @@
+LEVEL = ../../../../make
+SRCDIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+SWIFT_OBJC_INTEROP := 1
+
+SWIFT_SOURCES := main.swift
+
+include $(LEVEL)/Makefile.rules
+
+a.out: main.swift
+	$(SWIFTC) $(SWIFTFLAGS) -sdk "$(SWIFTSDKROOT)" $^ -emit-executable \
+            -o a.out \
+	    -emit-module -Xcc -I$(SRCDIR)
+
+clean::
+	rm -rf *.swiftmodule *.swiftdoc *.dSYM *~ lib*.dylib a.out *.o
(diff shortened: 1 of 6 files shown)


jimingham:
Adrian, I cobbed the test from your example, can you check I did it right?  Other than that the patch is pretty non-controversial.

jimingham:
@swift-ci please test

adrian-prantl:
(as discussed offline) we could do this even simpler without a bridging header.

jimingham:
This is the version w/o the bridging header.  The test still fails with current swift, and succeeds with the patch, so it still captures the failure.

jimingham:

dcci:
Howdy, this has broken the Linux bots (I haven't investigated why).
As this is a Linux test, you might consider adding the `@swiftpr` decorator.

dcci:
FWIW, this is failing to compile. I wonder if it ever worked on Linux?
https://gist.github.com/dcci/aec643b54379fd5d443cfb196b26a24d

dcci:
More explicitly, this is the problem:

```
/home/buildnode/jenkins/workspace/oss-lldb-incremental-linux-ubuntu-16_10/lldb/packages/Python/lldbsuite/test/lang/swift/expression/optional_amibiguity/Optional.h:1:9: error: 'Foundation/Foundation.h' file not found
12:37:15 #import <Foundation/Foundation.h>
12:37:15         ^
```

@jrose-apple do we support that header on Linux?

jrose-apple:
Nope. No Objective-C support on Linux. You could change the test to use a struct instead.

dcci:
Thanks, Jordan. I'll go ahead and fix this.

jrose-apple:
Why didn't it fail in PR testing?

dcci:
This didn't fail in PR testing because it wasn't decorated with `@swiftpr`. I think it's an easy thing to forget/overlook and the best way to catch this is peer-review (or, in the worst case, bots failures).

dcci:
Admittedly, for `lldb` PR testing we should run the full set of tests instead of the `@swiftpr` one. I think the reason why it didn't happen was because it might add a bit of overhead, but we should consider having `smoke test` and `full test` for lldb as well. What do you think?

dcci:
@jrose-apple I looked at the original report (https://bugs.swift.org/browse/SR-8120) and my understanding is that the test relies on the existence of an Obj-C protocol named `Optional`.
Given there's no `Obj-C` on Linux, maybe we should skip this test there altogether?

jrose-apple:
The test relies on the existence of a type named Optional that isn't Swift.Optional. It doesn't actually have anything to do with ObjC specifically, let alone protocols.

dcci:
Sorry, I was wrong and you're right. https://github.com/apple/swift-lldb/pull/752

PR #751, opened 2018-07-07
[4.2] [Playground] Replaced "$builtin" with "__builtin" in the playground preamble.

cwakamo:
This aligns with Swift, which is changing the instrumenter to use "__builtin" instead.

This is the swift-4.2-branch equivalent of #293, and is tied to apple/swift#17822. I'll add Swift 4.2 nomination information after it passes testing.

This is part of rdar://problem/36031860.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/playgrounds/PlaygroundsRuntime.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/playgrounds/PlaygroundsRuntime.swift
@@ -53,19 +53,19 @@ class LogRecord {
 }
 
 @_silgen_name ("playground_log_hidden") public func builtin_log_with_id<T>(_ object : T, _ name : String, _ id : Int, _ sl : Int, _ el : Int, _ sc : Int, _ ec: Int) -> AnyObject? {
-  return LogRecord(api:"$builtin_log", object:object, name:name, id:id, range : SourceRange(sl:sl, el:el, sc:sc, ec:ec))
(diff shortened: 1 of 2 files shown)


cwakamo:
Please test with this PR:

apple/swift#17822

@swift-ci Please test

cwakamo:
@dcci @jimingham If one of y'all would be able to re-review this for Swift 4.2 nomination purposes, I'd greatly appreciate it!

dcci (approved):
This is fine.

cwakamo:
• Explanation: Currently, the playground transform uses symbols like $builtin_log_with_id and others as the functions the compiler inserts calls to when running the transform. This means that playgrounds must be compiled with $-identifiers enabled, but this is suboptimal and effectively means that playgrounds must be built in a debugger-like context. To support more easily building with the playground transform with an unmodified compiler, as well as to bring the playground transform in alignment with the related PC macro, this change switches the playground transform to using functions named with a __builtin_* prefix instead of a $builtin_* prefix. This change also updates lldb to use __builtin_* when setting up the playground transform instead of $builtin_*.
• Scope of Issue: Affects the playground transform and lldb playground expressions.
• Origination: The original design of playgrounds used $builtin_* because it was only invoked from the debugger.
• Risk: Low
• Reviewed By: Swift changes reviewed by @jrose-apple, and LLDB changes reviewed by @dcci and @jimingham
• Testing: Update the Swift and LLDB test suites to use __builtin_* and observed no regressions.

PR #752, opened 2018-07-09
[Swift/Optional] Fix this test on Linux and make it simpler.

dcci:
-> We don't really need a module or the Obj-C bridging, just
another object called `Optional` to trigger the bug.

-> We now add this test to @swiftpr so that it gets run as
part of the standard PR pipeline.

Thanks to Jordan/Jim for discussing this with me.

<rdar://problem/41525749>


dcci:

PR #753, opened 2018-07-10
Report the exact stop location when "check_function_name" match fails.

jimingham:
Also always check that our "step" operations are completing and not getting interrupted.

We are seeing this test fail intermittantly on Linux, so I'm adding a little more
diagnostic to the fail to help see what actually went wrong.


jimingham:

PR #755, opened 2018-07-11
[testsuite] Implement a category to skip libstdcxx tests

dcci:
On systems where it's not supported.
As far as I understand Linux is the only systems which now ships
with libstdcxx (maybe NetBSD?, but I'm not entirely sure of the
state of lldb on the platform).
We could make this more fine grained looking for the header as
we do for libcxx. This is a little tricky as there's no such
thing as /usr/include/c++/v1, but libstdcxx encodes the version
number in the path (i.e. /usr/include/c++/5.4). I guess we might
match a regex, but it seems fragile to me.

Differential Revision:  https://reviews.llvm.org/D49110

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@336724 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/packages/Python/lldbsuite/test/dotest.py
+++ b/packages/Python/lldbsuite/test/dotest.py
@@ -1170,6 +1170,23 @@ def checkLibcxxSupport():
     print("Libc++ tests will not be run because: " + reason)
     configuration.skipCategories.append("libc++")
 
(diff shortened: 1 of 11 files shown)


dcci:
@swift-ci test and merge

PR #756, opened 2018-07-11
[ObjCRuntime] Add support for obfuscation in tagged pointers.

dcci:
This is the default in MacOS Mojave. No testcases, as basically
we have a lot of coverage (and the testsuite fails quite a bit
without this change in Beta 3).

Thanks to Fred Riss for helping me with this patch (fixing
bugs/nondeterminism).

<rdar://problem/38305553>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@336607 91177308-0d34-0410-b5e6-96231b3b80d8


dcci:

PR #746, opened 2018-07-06
Unbreak non darwin unified builds

davezarzycki:
Fix regression from #742

PR #758, opened 2018-07-11
Revert "Unbreak non darwin unified builds"

JDevlieghere:
The change in #746 is not correct. The goal is to use the in-tree dsymutil, which on swift-llvm still has the llvm-prefix. This reverts Dave's change and renames the dependency instead.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -15,11 +15,7 @@ function(add_python_test_target name test_script args comment)
     )
 endfunction()
 
-if (CMAKE_SYSTEM_NAME MATCHES "Darwin")
-  set(LLDB_TEST_DEPS lldb)
-else()
-  set(LLDB_TEST_DEPS lldb llvm-dsymutil)
-endif()
+set(LLDB_TEST_DEPS lldb)
 
 # darwin-debug is an hard dependency for the testsuite.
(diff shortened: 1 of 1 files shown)


JDevlieghere:
@swift-ci please test

davezarzycki:
Unless I'm missing something, this pull request only appears to revert #746. Where/what is the change that "renames the dependency instead"?

JDevlieghere:
It's not you, it's me. I was looking at Github's revert feature in the web interface but decided to just do it manually. When I came back to the repo the "create PR" suggested Github's branch instead of mine. Anyway, I pushed that commit to this branch.

JDevlieghere:
@swift-ci please test

davezarzycki (approved):
Verified. Unified builds continue to work on Red Hat Fedora 28 after this change.

davezarzycki:
PS – Reverts (and cherry-picks and rebases) aren't formally tracked in the commit schema, therefore I wouldn't go out of my way to use use `git revert` or whatever GitHub provides. As far as git is concerned, you could be using the classic `diff ... | patch -R ...` approach for the revert.

JDevlieghere:
Yeah, you're totally right. It was mostly out of curiosity, but once again I would've been better of using my regular workflow... :-)

dcci:
Go ahead.

dcci (approved):

PR #760, opened 2018-07-11
[SwiftExpressionParser] Always inject the DWARF name.

dcci:
Here we accidentally ended up injecting the demangled one, so
AST type checking was failing. Fixes several failures on
master-next/upstream-with-swift.

<rdar://problem/42047420>


jimingham:
It's a little interesting why we couldn't resolve the type for the demangled name, but that seems more error-prone.  Anyway, we know we will get it right for the unqualified name, so this is a good change.

PR #761, opened 2018-07-11
[upstream-with-swift] Fix a mismerge which caused build failures.

dcci:


dcci:
Down to 88 failures from 140
```
===================
Test Result Summary
===================
Failure:              69
Error:                18
Exceptional Exit:      1
```

PR #767, opened 2018-07-13
Fix the libcxx set, multiset, vector and bitset formatters to work on…

jimingham:
… references.

The synthetic child providers for these classes had a type expression that matched
pointers & references to the type, but the Front End only worked on the actual object.

I fixed this by adding a way for the Synthetic Child FrontEnd provider to request dereference,
and then had these formatters use that mode.

<rdar://problem/40849836>

Differential Revision: https://reviews.llvm.org/D49279

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@337035 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/include/lldb/DataFormatters/TypeSynthetic.h
+++ b/include/lldb/DataFormatters/TypeSynthetic.h
@@ -212,6 +212,19 @@ public:
       return *this;
     }
 
+    bool GetFrontEndWantsDereference() const {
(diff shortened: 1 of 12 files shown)


jimingham:
@swift-ci please test

jimingham:
@swift-ci please test

PR #769, opened 2018-07-15
[Swift] Exclude some tests from PR testing.

dcci:
These doesn't seem to be really stable, so we're temporarily
moving them out of the standard PR cycle to avoid delays in
merges (also, for the swift compiler).

SR-8245
<rdar://problem/42145676>


dcci:
@swift-ci test and merge

PR #770, opened 2018-07-15
[4.2] Update Fix-It test to rely on a Fix-It that isn't going away.

DougGregor:
The auto-insertion of '!' is going away with
rdar://problem/42081852. Switch to a different Fix-It (the removal of
an extraneous '!') instead. This is in support of https://github.com/apple/swift/pull/17955.


DougGregor:
@swift-ci please test

PR #771, opened 2018-07-16
[SwiftLanguageRuntime] Be more defensive in face of broken ASTcontexts.

dcci:
If we have errors, there's no point in keep going, so just stop.

Partially backports a change already on stable.

<rdar://problem/42048145>


dcci:
@swift-ci test and merge

fredriss (approved):

fredriss:
This handles my crash

jimingham:
There was another use in FindFunctionPointersInCall that gets the swift_ast and checks that it is not null, but doesn't check !HasErrors.

And GetBridgedSyntheticChildProvider gets types from the scratch SwiftASTContext and only checks !null, it doesn't check !HasFatalErrors.

If we're being complete you might want to fix those two uses as well.

PR #772, opened 2018-07-17
Disable EnableTargetOSChecking in the scratch AST context.

fredriss:
This is not needed in the general case, as this flag is set by the
expression parser. There is one case where it matters though: if the
Swift code is loaded by a main executable that has a deploymnent target
lower than 10.9, then we will fail to import the standard library
module in SwiftASTContext::CreateInstance().

I'm still working on a testcase for this, but I wanted to get the trivial
fix in for users to test.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1345,6 +1345,8 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
   if (!arch.IsValid())
     return TypeSystemSP();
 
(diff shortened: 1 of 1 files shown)


fredriss:
@swift-ci test and merge

jimingham (approved):
IIUC, the intention is that we always turn off the OS checking because we know the program managed to run, so if we are tripping on check, lldb likely is in error.  So doing it right when we make the target scratch AST context is appropriate.

fredriss:
Yes, this is correct. And moreover, we set this flag for basically all other uses of the scratch context through the expression evaluation, so it will only impact loading the standard library in CreateInstance.

PR #775, opened 2018-07-17
Dev/rdar 41652444 disable deployment target check stable

fredriss:


fredriss:
@swift-ci test and merge

PR #776, opened 2018-07-17
[CMake] Have check-lldb-unit use CMAKE_CURRENT_BINARY_DIR

JDevlieghere:
llvm-lit uses `map_config` directives (generated at configuration-time) to
make it possible to pass a test path relative to the source instead of
the build folder (CMAKE_CURRENT_BINARY_DIR).

However, this doesn't work in the case of swift where the build
directory layout prevents llvm-lit from knowing about lldb and its test
paths. This caused check-lldb-unit to fail in this configuration, while
everything was working as expected upstream. This patch fixes the issue
by passing a path relative to the build directory. This was already the
case for check-lldb-lit.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@337291 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit e71634d556cf23871b427126d15adf9fc9c05283)

### DIFF
--- a/lit/CMakeLists.txt
+++ b/lit/CMakeLists.txt
@@ -41,12 +41,12 @@ set(LLDB_TEST_DEPS
 if(NOT LLDB_BUILT_STANDALONE)
   list(APPEND LLDB_TEST_DEPS FileCheck not yaml2obj)
 endif()
-  
+
 # lldb-server is not built on every platform.
 if (TARGET lldb-server)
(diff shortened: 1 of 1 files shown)


JDevlieghere:

weliveindetail:
Thanks for investigating. Will double-check in a bit.

weliveindetail (commented):
Fixed it!

PR #777, opened 2018-07-17
Make these tests c++ tests so they can be skipped on systems that don…

jimingham:
…'t support those tests.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@337058 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #782, opened 2018-07-19
Fix TestAttachDenied on macOS Mojave

dcci:
TestAttachDenied tries to attach to a process that is ptracing itself and
verifies that we error out. Starting with macOS Mojave, processes need
an entitlement to be able to ptrace. This commit adds the entitlement for
the test binary when building on Darwin.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@337029 91177308-0d34-0410-b5e6-96231b3b80d8


dcci:
@swift-ci test and merge

PR #783, opened 2018-07-20
No empty search paths

jimingham:
The new dlopen along a list of paths code doesn't work if any of the paths are the empty string.  Filter them out in the dlopen code, and also remove a thinko in the GetLibrarySearchPaths from SwiftASTContext.cpp that was adding empty paths to the search path list.


jimingham:

PR #784, opened 2018-07-20
No empty paths

jimingham:
The new "load using paths" code didn't handle empty paths correctly, and the code in SwiftASTContext.cpp:GetLibrarySearchPaths errantly inserted empty strings into the path list.  That had the effect of truncating the search before the RuntimeLibraryPath was used, and on Linux that meant you couldn't import GlibC and a few other standard modules into the REPL.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -3413,9 +3413,8 @@ static void
 GetLibrarySearchPaths(std::vector<std::string> &paths,
                       const swift::SearchPathOptions &search_path_opts) {
   paths.clear();
-  paths.resize(search_path_opts.LibrarySearchPaths.size() + 1);
-  std::copy(search_path_opts.LibrarySearchPaths.begin(),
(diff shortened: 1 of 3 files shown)


jimingham:
@swift-ci please test

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3413,9 +3413,8 @@ static void
 GetLibrarySearchPaths(std::vector<std::string> &paths,
                       const swift::SearchPathOptions &search_path_opts) {
   paths.clear();
-  paths.resize(search_path_opts.LibrarySearchPaths.size() + 1);
-  std::copy(search_path_opts.LibrarySearchPaths.begin(),
-            search_path_opts.LibrarySearchPaths.end(), paths.begin());
+  paths.assign(search_path_opts.LibrarySearchPaths.begin(),
```
fredriss, line 3416: why not simply
`paths = search_path_opts.LibrarySearchPaths`
?

jimingham: Dunno, that's what Shafik suggested.

PR #786, opened 2018-07-21
Delay passing the persistent_state till after making the expression c…

jimingham:
…ould invalidate it.

In the SwiftExpressionParser, we set the persistent_state associated with the currently
active scratch AST context into a couple of delegate objects that we will use at the
end of the expression evaluation.  But in the course of making an expression parser in the context
of a new module, some conflicting imports could invalidate the scratch AST context, leaving
the persistent_state pointers as dangling pointers.

So this patch delays getting and passing the persistent_state to the delegates till after
making the new SwiftExpressionParser, at which point we know the ones we calculate are the
ones we really intend to use.

Before this patch, I was getting a crash in:

lang/swift/clangimporter/objcmain_conflicting_dylibs_bridging_headers

<rdar://problem/42410482>

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftUserExpression.cpp
@@ -363,15 +363,6 @@ bool SwiftUserExpression::Parse(DiagnosticManager &diagnostic_manager,
                                  "couldn't start parsing (no target)");
     return false;
   }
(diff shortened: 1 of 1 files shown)


jimingham:
@swift-ci please test

fredriss:
LGTM and this could be the source of a number of hard to debug crashes, so let's take it.

adrian-prantl (commented):
I think this looks reasonable.

PR #790, opened 2018-07-25
Get the running OS version from the process rather than the platform.

fredriss:
It turns out the OS information returned by the platform can't be trusted
because the platform code is bogus (rdar://problem/42595503). Get the OS
version information from the process instead where it is accurate. This
commit also fixes the test to pass an older deployment target to the
compilation command so that it actually tests the feature we are trying
to exercise.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/availability/TestAvailability.py
+++ b/packages/Python/lldbsuite/test/lang/swift/availability/TestAvailability.py
@@ -16,6 +16,21 @@ def getOSName(os):
     if os == 'watchos': return 'watchOS'
     return os
 
+def getArch(os):
+    if os == 'macosx': return 'x86_64'
+    if os == 'ios': return 'arm64'
+    if os == 'tvos': return 'arm64'
+    if os == 'watchos': return 'armv7k'
+    return os
+
+def getTriple(os, version):
+    return getArch(os) + '-apple-' + os + version
+
+def getOlderVersion(major, minor):
+    if minor != 0:
+        return '%d.%d' % (major, minor-1)
+    return '%d.%d' % (major-1, minor)
+
 class TestAvailability(TestBase):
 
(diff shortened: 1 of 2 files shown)


fredriss:
@swift-ci test and merge

scripts/Xcode/repos/FALLBACK
```diff
@@ -1,20 +1,20 @@
 {
-    "regexs" : [".*github.*apple.*swift-lldb.*llvm-swift5-transition.*"],
+    "regexs" : [".*github.*apple.*swift-lldb.*stable.*"],
     "repos" : [
         {"name": "llvm",
          "vcs": "git",
          "url": "https://github.com/apple/swift-llvm.git",
-         "ref": "swift-5.0-branch"},
+         "ref": "stable"},
```
dcci: Do you really want to commit this file?

scripts/Xcode/repos/FALLBACK
```diff
@@ -1,20 +1,20 @@
 {
-    "regexs" : [".*github.*apple.*swift-lldb.*llvm-swift5-transition.*"],
+    "regexs" : [".*github.*apple.*swift-lldb.*stable.*"],
```
jrose-apple: Was this change supposed to get into this commit?

scripts/Xcode/repos/FALLBACK
fredriss: oops, no

fredriss:
@swift-ci test and merge

jasonmolenda:
Should we check that major/minor/patch (at least major/minor) are not equal to -1 which is what the Process base class will return if a subclass doesn't override?  If we know what the oldest iOS version we support with this Xcode, we could install it on a device and make sure qHostInfo includes the version #.  This codepath is only executed on darwin targets so old iOS devices are the only concern I have.

PR #791, opened 2018-07-25
Get the running OS version from the process rather than the platform.

fredriss:
It turns out the OS information returned by the platform can't be trusted
because the platform code is bogus (rdar://problem/42595503). Get the OS
version information from the process instead where it is accurate. This
commit also fixes the test to pass an older deployment target to the
compilation command so that it actually tests the feature we are trying
to exercise.

rdar://problem/42351790

(cherry picked from commit dbf58532c3a277690d28a88a52c4a564fd4fc2d8)


fredriss:
@swift-ci test and merge

PR #793, opened 2018-07-25
An additional fix to ObjectFileMachO's minimum version

jasonmolenda:
An additional fix to ObjectFileMachO's minimum version
number reporting; handle the case where we're given a
minimum OS version of 0 correctly. This is the same
as the change I made in r335556 for ObjectFileMachO::GetSDKVersion
but it also needs to be made to the similar code in
ObjectFileMachO::GetMinimumOSVersion.

rdar://problem/41372699


jasonmolenda:
@swift-ci please test

fredriss:
Looks good

PR #796, opened 2018-07-26
[lit] Relax NSString test by not checking the REPL welcome message, NFC

vedantk:
This NSString-in-REPL test was checking the text of the REPL welcome
message, which is a little brittle, and was broken by a change which
added a newline.

It seems simpler to not check the welcome text at all.

https://ci.swift.org/job/oss-lldb-incremental-osx-cmake/1112


vedantk:

PR #798, opened 2018-07-29
Add a test case for LLDB using -debug-prefix-map.

allevato:
As requested by @dcci to follow-up on https://github.com/apple/swift/pull/17665 (where `-debug-prefix-map` was implemented).

This is my first time diving into LLDB's test cases, so I tried to base it on existing examples. Feedback appreciated!

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/debug_prefix_map/TestSwiftDebugPrefixMap.py
+++ b/packages/Python/lldbsuite/test/lang/swift/debug_prefix_map/TestSwiftDebugPrefixMap.py
@@ -1,0 +1,77 @@
+# TestSwiftDebugPrefixMap.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""
+Test that LLDB correctly finds source when debug info is remapped.
+"""
+import lldb
+from lldbsuite.test.lldbtest import *
+import lldbsuite.test.decorators as decorators
+import lldbsuite.test.lldbutil as lldbutil
+import os
+import os.path
+import shutil
+import unittest2
+
+
+class TestSwiftDebugPrefixMap(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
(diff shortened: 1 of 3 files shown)


packages/Python/lldbsuite/test/lang/swift/debug_prefix_map/TestSwiftDebugPrefixMap.py
```diff
@@ -0,0 +1,77 @@
...
+        self.assertTrue(target, VALID_TARGET)
+
+        # Don't allow ANSI highlighting to interfere with the output.
+        self.runCmd('settings set stop-show-column none')
+        self.expect('breakpoint set -l 13', substrs=['foo'])
+        self.expect('source list -l 13', substrs=['return x + y - z'])
+        self.expect('run', substrs=['return x + y - z'])
+
+
+if __name__ == '__main__':
```
dcci: Do you need this?

allevato: Unfortunately I can't tell through GitHub's UI which part "this" is referring to; the comment only shows up on line 73 but the e-mail cited all the way up to line 59.

If you're referring to lines 59-70 (or even 67-70), the goal there is to ensure that the source file is found based on the prefix map; if the flag didn't work correctly, then the commands at line 69-70 wouldn't return the correct source text.

Or were you referring to something else?

dcci:
I'm referring to everything after the `if name == __main__`.
My understanding is that lldb tests should never be run in that mode.

allevato:
There are some in the test/lang/swift directory that don't do that, like [this one](https://github.com/apple/swift-lldb/blob/stable/packages/Python/lldbsuite/test/lang/swift/closures/TestSwiftClosures.py) but there are also several that do:

https://github.com/apple/swift-lldb/blob/stable/packages/Python/lldbsuite/test/lang/swift/conditional_breakpoints/TestSwiftConditionalBreakpoint.py
https://github.com/apple/swift-lldb/blob/stable/packages/Python/lldbsuite/test/lang/swift/cross_module_extension/TestSwiftCrossModuleExtension.py
https://github.com/apple/swift-lldb/blob/stable/packages/Python/lldbsuite/test/lang/swift/path_with_colons/TestSwiftPathWithColons.py

It's not obvious to me what the difference, if any, is among those, however.

dcci:
I think it's just historical and redundant. You can remove it.

allevato:
Done.

dcci:
@allevato thanks, I'm going to merge this.

dcci:

allevato:
Thank you!

PR #799, opened 2018-07-30
Move the LLDB_TEST_SWIFT to test/CMakeLists.txt where it is used

fredriss:
It looks like the LLDB_TEST_SWIFT option never actually worked. We never
noticed as master-next was broken, but daa2eb72666e78be7 removed the
fallback code that seemed to have kept the swift tests running.

This commit also removes LLDB_TEST_CLANG which doesn't seem to be used
anywhere.


fredriss:
I'm merging this even though I'm not sure it's the right way to fix the issue. It gets upstream-with-swift to run swift tests again.

JDevlieghere:
This looks fine to me.

PR #800, opened 2018-07-30
Sync-up the installed Clang resource directory and the code looking f…

fredriss:
…or it.

All Swift tests where failing because they couldn't build the ClangImporter. This patch gets
things going again even though I'm really not sure it's the correct solution.


fredriss:
I'm merging this as it fixes all the Swift tests.

PR #802, opened 2018-07-31
[SwiftREPL] Catch up with changes upstream.

dcci:
lldb failing to parse arguments now returns an exit code of 1,
update the test to reflect the new world order.


vedantk:
Thanks!

PR #804, opened 2018-08-01
Avoid crashing when printing global variables without a process.

adrian-prantl:
This fixes a regression introduced the big generic type representation overhaul.


adrian-prantl:
@swift-ci test and merge

PR #805, opened 2018-08-02
 Pass in the current LLDB module whenever a Swift scratch context is r...

adrian-prantl:
…equested.

Using the scratch context for dynamic type resolution has uncovered a
couple of omissions in the original fallback mechanism
implementation. This patch adds a few convenience accesors that make
it easy to always get access to the most appropriate Swift scratch
context in all situations.

The testcase causes the scratch context to get destroyed by a
conflict that is triggered via dynamic type resolution via "frame variable".

rdar://problem/42554384


adrian-prantl:
@swift-ci test

adrian-prantl:
@swift-ci test and merge

PR #806, opened 2018-08-02
Enable the per-module scratch AST context fallback in dynamic type re…

adrian-prantl:
…solution.

After switching dynamic type resolution over to the scratch context,
we also need to make sure to support the fallback mechanism there.

This is a smaller, more targeted version of the same patch that went
into the stable branch.

rdar://problem/42554384


adrian-prantl:
@swift-ci test

PR #807, opened 2018-08-02
[SwiftLanguageRuntime] Switch Error dynamic type discovery to RemoteAST.

dcci:
<rdar://problem/37479968>

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1640,173 +1640,34 @@ bool SwiftLanguageRuntime::IsValidErrorValue(
 bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_ErrorType(
     ValueObject &in_value, lldb::DynamicValueType use_dynamic,
     TypeAndOrName &class_type_or_name, Address &address) {
-  // layout of error type
-  // pointer to -------> SwiftError {
-  // --------------
-  // CFRuntimeBase
-  // CFIndex
-  // CFStringRef
-  // CFDictionaryRef
-  // --------------
-  // Metadata
-  // WitnessTable
-  // hashable Metadata
-  // hashable WitnessTable
-  // --------------
-  // tail allocated actual object data *
-  // }
-  // * for a struct, it's the inline data
-  // * for a class, it's the inline pointer-to-the-data (aka, the swift class
-  // instance)
-  SwiftErrorDescriptor error_descriptor;
-  if (!IsValidErrorValue(in_value, &error_descriptor))
+  lldb::addr_t existential_address = in_value.GetAddressOf();
+  if (!existential_address || existential_address == LLDB_INVALID_ADDRESS)
     return false;
 
-  Status error;
   CompilerType var_type(in_value.GetStaticValue()->GetCompilerType());
-  size_t ptr_size = m_process->GetAddressByteSize();
   SwiftASTContext *swift_ast_ctx =
       llvm::dyn_cast_or_null<SwiftASTContext>(var_type.GetTypeSystem());
-  if (!swift_ast_ctx)
+  if (!swift_ast_ctx || swift_ast_ctx->HasFatalErrors())
     return false;
 
-  switch (error_descriptor.m_kind) {
-  case SwiftErrorDescriptor::Kind::eNotAnError:
+  auto &remote_ast = GetRemoteASTContext(*swift_ast_ctx);
+  swift::remote::RemoteAddress remote_existential(existential_address);
+  auto dyn_result = remote_ast.getDynamicTypeAndAddressForExistential(
+      swift::remote::RemoteAddress(remote_existential),
+      GetSwiftType(in_value.GetCompilerType()));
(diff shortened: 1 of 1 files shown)


source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1599,176 +1599,41 @@ bool SwiftLanguageRuntime::IsValidErrorValue(
...
-  if (!swift_ast_ctx)
+  lldb::addr_t existential_address = instance_type_sp->GetValueAsUnsigned(0);
+  if (existential_address == 0 || existential_address == LLDB_INVALID_ADDRESS)
     return false;
 
-  switch (error_descriptor.m_kind) {
-  case SwiftErrorDescriptor::Kind::eNotAnError:
+  // Dynamic type resolution in RemoteAST might pull in other Swift modules, so
+  // use the scratch context where such operations are legal and safe.
+  SwiftASTContext *swift_ast_ctx = GetScratchSwiftASTContext();
```
dcci: @adrian-prantl Adrian, I saw you changed things here a bit, should this move to your new utility function?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1599,176 +1599,41 @@ bool SwiftLanguageRuntime::IsValidErrorValue(
...
-  } break;
-  }
+  auto &remote_ast = GetRemoteASTContext(*swift_ast_ctx);
+  swift::remote::RemoteAddress remote_existential(existential_address);
+  auto dyn_result = remote_ast.getDynamicTypeAndAddressForExistential(
+      swift::remote::RemoteAddress(remote_existential),
+      GetSwiftType(in_value.GetCompilerType()));
+  if (!dyn_result)
+    return false;
+  auto swift_dynamic_type = dyn_result.getValue().first.getPointer();
```
jimingham, line 1660: Shouldn't you use swift_dynamic_type in the next line rather than doing the calculation again?

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: the correct new call is probably `in_value.GetScratchSwiftASTContext();`

dcci:
For the folks listening at home (if any). This turned out to be slightly trickier than expected.

The issue is that the way the lldb expression parser works is somehow in constrast with the way the newly introduced API works.

`frame var` is fine, as we hand to RemoteAST a pointer to target memory and then the state machine reads the metadata and happily solve a dynamic type and project the address of the value.

OTOH, `expr` "dematerializes" a variable (in lldb lingo), i.e. makes a frozen copy of the object in host memory. This is a fundamental assumption in the way lldb expression parser works, as it's expected that return variables are kind-of immutable objects that can be saved for later use. As RemoteAST needs a target pointer to do its job, the code as-is doesn't work.

Something that @jimingham suggested is that of anticipating the resolution of dynamic types for result variables, i.e. when we have target memory still available (pre-dematerialization).

jimingham (approved):
Assuming RemoteAST now does all the work you've deleted for us, this is great.  I had one trivial comment

source/Target/SwiftLanguageRuntime.cpp
dcci: Yes, I should. I'll fix that up once we figure out what to do with the expression parser problem :) [and before merging].

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1640,173 +1640,34 @@ bool SwiftLanguageRuntime::IsValidErrorValue(
...
-        }
-      }
-    }
-  } break;
-  }
-
-  return false;
+  auto swift_dynamic_type = dyn_result.getValue().first.getPointer();
+  CompilerType result_type(swift_ast_ctx,
+                           dyn_result.getValue().first.getPointer());
```
adrian-prantl, line 1662: This looks dangerous, because we're mixing `swift_ast_ctx` with a type pointer from dyn_result. Dynamic types always use the scratch ast context and `swift_ast_ctx` seems to come from the static type which may be a pointer into any of the module swift ast contexts.

adrian-prantl (changes requested):

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: The pattern to import a type from one context into the scratch context is:
```
  auto scratch_ctx = target.GetScratchSwiftASTContext(error, lldb_module);
  if (!scratch_ctx)
    return base_type;
  base_type = scratch_ctx->ImportType(base_type, error);
```

adrian-prantl: cf. SwiftLanguageRuntime::DoArchetypeBindingForType()

dcci: I think this is a good point. I would like to note that this is broken in the current version. Not that it matters or that we shouldn't fix it.

dcci:
Subsumed.

PR #809, opened 2018-08-07
[REPL] Set the correct rpath when installing.

dcci:
$ORIGIN represents the location of the REPL at runtime,
whenever the binary is located. So, this makes sure
that when we generate the toolchain on one machine and
transfer to another, the repl keeps working, as it needs
to `dlopen` libSwiftCore, which sits in a path relative
to the executable itself.

<rdar://problem/42863249>

### DIFF
--- a/tools/repl/swift/CMakeLists.txt
+++ b/tools/repl/swift/CMakeLists.txt
@@ -5,6 +5,7 @@ if (LLDB_BUILD_FRAMEWORK)
 elseif( CMAKE_SYSTEM_NAME MATCHES "Linux" )
   # Set the correct rpath to locate libswiftCore
   set(CMAKE_EXE_LINKER_FLAGS "-Wl,-rpath,${CMAKE_BINARY_DIR}/../swift-linux-x86_64/lib${LLVM_LIBDIR_SUFFIX}/swift/linux -Wl,-ldl")
+  set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib/swift/linux:${CMAKE_INSTALL_RPATH}")
 endif()
 
 add_lldb_tool(repl_swift


dcci:
@swift-ci please test

dcci:
@aciidb0mb3r this should fix the issue you saw on the downloaded linux toolchain
@shahmishal if you want to give this a try, go ahead.

vedantk (approved):
Thanks!

adrian-prantl:
Seems plausible.. what is `$ORIGIN`?

dcci:
`$ORIGIN` means "this executable", as in "the path at runtime for the dummy repl executable".
The dynamic loader, i.e. `ld.so` on Linux, expands this variable (and others) if they appear in the tags specified through `-rpath`.

The trick is that the static linker (`ld.bfd` or `ld.gold`) emits a `DT_RPATH/DT_RUNPATH` tag (depending on the version) containing a path where the dynamic linker should look for dynamic libraries to `dlopen()`.

I found `man ld.so` and `man dlopen` pretty enlightening to find out where the issue was. If you have other questions about this patch in particular or the mechanism though, don't hesitate to ask!

adrian-prantl:
Thanks for the explanation!

gottesmm:
@dcci Is it possible to add a test to the integration test suite that would catch this?

gottesmm:
@dcci Is it possible to add a test to the integration test suite that would catch this?

dcci:
@gottesmm I thought about this, but it's a little tricky, because it will require inspecting a binary and seeing that `rpath` is correct. Maybe we can rely on `objdump` (or llvm-objdump) being there and built and grep for `DT_RPATH` being there and containing the correct string for the installed executable? What do you think?

gottesmm:
@dcci That would be fine. I did something similar to this to make sure that on linux we don't mount any libraries w+x.

See: https://github.com/apple/swift-integration-tests/tree/master/test-snapshot-binaries

gottesmm:
Here is where I added llvm-readelf support. I would use llvm-objdump.

gottesmm:
https://github.com/apple/swift-integration-tests/blob/e882c92e1f063f5971f11c4163414fd75356f521/lit.cfg#L142

dcci:
@gottesmm yes, that's pretty much what I had in mind. `llvm-readelf` is fine, as we don't need to disassemble anything but just dump the `DT_RPATH` tags. I'll take a stab at that later tonight on the bus or tomorrow.

gottesmm:
SGTM! Thanks Davide!

dcci:
@gottesmm (and future archeologists), the integration test is here: 
https://github.com/apple/swift-integration-tests/pull/31

PR #810, opened 2018-08-08
[lldb-mi] Disable flakey pexpect tests

JDevlieghere:
Disable two pexpect-based test that are timing out. The real fix for
this is to rewrite the tests using LIT but until then we removed these
two from Darwin testing.


JDevlieghere:

PR #812, opened 2018-08-09
radar42585636-demangling

weliveindetail:
This PR brings https://reviews.llvm.org/D50071 to the current swift-lldb stable.
It depends on https://github.com/apple/swift-llvm/pull/106

Notes:
* Requires a fix in LLVM (see 5b2b8b5 in swift-llvm)
* Depends on llvm::Any, which was introduced after the cut and needs 4 cherry-picks from the upstream LLVM
* Pavel submitted a big reorg in between my commits (see https://reviews.llvm.org/D497404), which caused some merge headache, mostly in (less critical?) Xcode project files

I think this would be nice to have and it's all sorted out now, but if you have any concerns please speak up.

### DIFF
--- a/include/lldb/Core/Mangled.h
+++ b/include/lldb/Core/Mangled.h
@@ -11,18 +11,15 @@
 #define liblldb_Mangled_h_
 #if defined(__cplusplus)
 
+#include "lldb/lldb-enumerations.h"
+#include "lldb/lldb-forward.h"
+
 #include "lldb/Utility/ConstString.h"
-#include "lldb/lldb-enumerations.h" // for LanguageType
-#include "llvm/ADT/StringRef.h"     // for StringRef
 
-#include <stddef.h> // for size_t
+#include "llvm/ADT/StringRef.h"
 
-namespace lldb_private {
-class RegularExpression;
-}
-namespace lldb_private {
-class Stream;
-}
+#include <memory>
+#include <stddef.h>
 
 namespace lldb_private {
 
@@ -238,7 +235,6 @@ public:
       return true;
     return GetDemangledName(language) == name;
   }
-
   bool NameMatches(const RegularExpression &regex,
                    lldb::LanguageType language) const;
 
@@ -299,6 +295,36 @@ public:
   ///     if there is no mangled or demangled counterpart.
   //----------------------------------------------------------------------
   lldb::LanguageType GuessLanguage() const;
+
+  /// Function signature for filtering mangled names.
+  using SkipMangledNameFn = bool(llvm::StringRef, ManglingScheme);
+
+  //----------------------------------------------------------------------
+  /// Trigger explicit demangling to obtain rich mangling information. This is
+  /// optimized for batch processing while populating a name index. To get the
+  /// pure demangled name string for a single entity, use GetDemangledName()
+  /// instead.
+  ///
+  /// For names that match the Itanium mangling scheme, this uses LLVM's
+  /// ItaniumPartialDemangler. All other names fall back to LLDB's builtin
+  /// parser currently.
+  ///
+  /// This function is thread-safe when used with different \a context
+  /// instances in different threads.
+  ///
+  /// @param[in] context
+  ///     The context for this function. A single instance can be stack-
+  ///     allocated in the caller's frame and used for multiple calls.
+  ///
+  /// @param[in] skip_mangled_name
+  ///     A filtering function for skipping entities based on name and mangling
+  ///     scheme. This can be null if unused.
+  ///
+  /// @return
+  ///     True on success, false otherwise.
+  //----------------------------------------------------------------------
+  bool DemangleWithRichManglingInfo(RichManglingContext &context,
+                                    SkipMangledNameFn *skip_mangled_name);
 
 private:
   //----------------------------------------------------------------------
--- a/source/Core/Mangled.cpp
+++ b/source/Core/Mangled.cpp
@@ -235,7 +235,7 @@ void Mangled::Clear() {
 int Mangled::Compare(const Mangled &a, const Mangled &b) {
   return ConstString::Compare(
       a.GetName(lldb::eLanguageTypeUnknown, ePreferMangled),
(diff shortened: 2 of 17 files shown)


include/lldb/Core/Mangled.h
```diff
@@ -11,18 +11,15 @@
...
+#include "llvm/ADT/StringRef.h"
 
-namespace lldb_private {
-class RegularExpression;
-}
-namespace lldb_private {
-class Stream;
-}
+#include <memory>
+#include <stddef.h>
```
weliveindetail, line 23: Had a merge conflict in the lines above, because in the review I was asked to move forward decls to `lldb-forward.h` and remove comments for includes.

source/Core/Mangled.cpp
```diff
@@ -49,13 +35,14 @@
...
+#include "lldb/lldb-enumerations.h"
 
 #include "Plugins/Language/CPlusPlus/CPlusPlusLanguage.h"
 #include "Plugins/Language/ObjC/ObjCLanguage.h"
 
-#include "llvm/ADT/StringRef.h"    // for StringRef
-#include "llvm/Support/Compiler.h" // for LLVM_PRETT...
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Demangle/Demangle.h"
+#include "llvm/Support/Compiler.h"
```
weliveindetail, line 46: Had a merge conflict in the lines above, because I added `lldb/Core/RichManglingContext.h` and removed comments on includes.

source/Symbol/Symtab.cpp
```diff
@@ -283,67 +322,18 @@ void Symtab::InitNameIndexes() {
...
-              }
+        const SymbolType type = symbol->GetType();
+        if (type == eSymbolTypeCode || type == eSymbolTypeResolver) {
+          // Other schemes are not relevant in the Swift use case.
+          bool is_relevant_itanium = !lldb_skip_name(
+              entry.cstring.GetStringRef(), Mangled::eManglingSchemeItanium);
+
+          if (is_relevant_itanium) {
+            if (mangled.DemangleWithRichManglingInfo(rmc, lldb_skip_name)) {
+              RegisterMangledNameEntry(entry, class_contexts, backlog, rmc);
```
weliveindetail, line 333: Had a merge conflict here due to logic changes

fredriss: Why did you add the is_relevant_itanium bool that AFAICS is not present upstream? doesn't the lldb_skip_name argument to DemangleWithRichManglingInfo fulfill the same purpose? If not, why is this condition not on llvm.org?

I have merged your patch to upstream-with-swift a couple days ago, and this is the resolution I came up with:
```
        const SymbolType type = symbol->GetType();
        if (type == eSymbolTypeCode || type == eSymbolTypeResolver) {
          if (mangled.DemangleWithRichManglingInfo(rmc, lldb_skip_name))
            RegisterMangledNameEntry(entry, class_contexts, backlog, rmc);
          else if (SwiftLanguageRuntime::IsSwiftMangledName(name.str().c_str())) {
            lldb_private::ConstString basename;
            bool is_method = false;
            ConstString mangled_name = mangled.GetMangledName();
            if (SwiftLanguageRuntime::MethodName::
                    ExtractFunctionBasenameFromMangled(mangled_name, basename,
                                                       is_method)) {
              if (basename && basename != mangled_name) {
                entry.cstring = basename;
                if (is_method)
                  m_method_to_index.Append(entry);
                else
                  m_basename_to_index.Append(entry);
              }
            }
          }
        }
```

Is this wrong?

weliveindetail: Good point. No it's not wrong. If a name matches `lldb_skip_name` then `DemangleWithRichManglingInfo` has no side effects. I thought about that and then decided to stick to the existing behaviour as close as possible, which means: Only call `SwiftLanguageRuntime::IsSwiftMangledName()` for skipped names and not for other cases where `DemangleWithRichManglingInfo` returns `false` i.e. when `IPD::partialDemangle()` fails.

I think this is good for the release branch, but we can keep your resolution on upstream-with-swift as there's time to fix potential issues.

weliveindetail: Btw. the other way around is not worrying, as Swift names have a `_T` prefix while Itanium uses `_Z`. Thus, IPD will never take on a Swift name and will always reach `SwiftLanguageRuntime::IsSwiftMangledName()`.

weliveindetail:
@swift-ci test and merge

weliveindetail:
Failed due to `RichManglingContext.h:18:10: fatal error: 'llvm/ADT/Any.h' file not found`
Trying again with https://github.com/apple/swift-llvm/pull/107 in there now

weliveindetail:
@swift-ci test and merge

weliveindetail:
From the log at https://ci.swift.org/job/swift-lldb-PR-Linux/808/consoleText
```
/home/buildnode/jenkins/workspace/swift-lldb-PR-Linux/branch-stable/llvm
HEAD is now at 73ae8cb Merge remote-tracking branch 'origin/swift-5.0-branch' into stable
```
Which is 2 days behind:
```
$ git show 73ae8cb
commit 73ae8cb3c57ec97c8fa0ec50b10ad7c046857739
Merge: 843d02f5106 efc35a0bed0
Author: swift-ci <swift-ci@users.noreply.github.com>
Date:   Fri Aug 10 15:39:22 2018 -0700

    Merge remote-tracking branch 'origin/swift-5.0-branch' into stable
```
Waiting a little longer for 2181912.

weliveindetail:
@swift-ci test and merge

weliveindetail:
This gets quite hairy now. In order for the OSX Platform tests to succeed it needs this fix: https://reviews.llvm.org/rL339669 As we did that afterwards on llvm.org, the tests in the original patch use a function that was renamed with the patch I want to move over here..

I will move this over in two separate PRs:
* functional change: https://github.com/apple/swift-lldb/pull/824
* tests: todo

weliveindetail:
Committing all of this in one fat PR: https://github.com/apple/swift-lldb/pull/825

PR #813, opened 2018-08-09
Fix data formatters for Swift Set and Dictionary types

lorentey:
This PR updates lldb's Swift Set and Dictionary data formatters, removing obsolete paths, simplifying code structure and fixing edge cases.

rdar://problem/43045289

Partial list of fixes:
- Fixes summary formatters for Swift sets -- they should be summaried as `<n> values`, like arrays.
    ```
    $ lldb --repl
    1> import Foundation

    // Before: ❌
    2> [1, 2] as Set<Int>
    $R0: Set<Int> = {  // Note: no summary
      [0] = 1
      [1] = 2
    }
    // After: 👍
    2> [1, 2] as Set<Int>
    $R0: Set<Int> = 2 values {
       [0] = 1
       [1] = 2
     }
    ```
- Empty sets and dictionaries using the type-punned singleton storage instance are now formatted correctly. This also works when the storage instance is verbatim bridged to Objective-C.
    ```
    // Before: ❌
    2> Set<Int>()
    $R0: Set<Int> = {} 
    3> $R0 as NSSet
    $R1: _RawNativeSetStorage = {
      Swift._SwiftNativeNSSet = {}
      bucketCount = 1
      count = 0
      ... more gobbledygook ...
    }

    // After: 👍
    2> Set<Int>()
    $R0: Set<Int> = 0 values
    3> $R0 as NSSet
    $R1: _RawNativeSetStorage = 0 values
    ```

- Swift sets and dictionaries that are bridged to Objective-C now use the proper formatter, rather than a raw data dump:
    ```
    // Before: ❌
    2> [1, 2] as Set<Int> as NSSet
    $R0: _SwiftDeferredNSSet<Int> = {
       Swift._SwiftNativeNSSet = {}
       _heapStorageBridged_DoNotUse = nil
       nativeBuffer = {
         _storage = {
           ... more gobbledygook ...
         }
       }
    }
    3> [NSNumber(value: 1), NSNumber(value: 2)] as Set<NSObject> as NSSet
    $R1: _HashableTypedNativeSetStorage<NSObject> = {
      Swift._TypedNativeSetStorage = {
        Swift._RawNativeSetStorage = {
          Swift._SwiftNativeNSSet = {}
          bucketCount = 4
           ... more gobbledygook ...
        }
      }
    }

    // After: 👍
    2> [1, 2] as Set<Int> as NSSet
    $R0: _SwiftDeferredNSSet<Int> = 2 values {
      [0] = 1
      [1] = 2
    }
    3> [NSNumber(value: 1), NSNumber(value: 2)] as Set<NSObject> as NSSet
    $R1: _HashableTypedNativeSetStorage<NSObject> = 2 values {
      [0] = Int64(2)
      [1] = Int64(1)
    }
    ```
- Cocoa NSSet and NSDictionary instances that are verbatim-bridged to Swift are now formatted correctly:
    ```
    // Before: ❌
    2> NSSet(array: [NSNumber(value: 1), NSNumber(value: 2)]) as! Set<NSObject>
    $R0: Set<NSObject> = {}

    // After: 👍
    2> NSSet(array: [NSNumber(value: 1), NSNumber(value: 2)]) as! Set<NSObject>
    $R0: Set<NSObject> = 2 values {
      [0] = Int64(1)
      [1] = Int64(2)
    }
    ```

### DIFF
--- a/source/Plugins/Language/Swift/SwiftDictionary.cpp
+++ b/source/Plugins/Language/Swift/SwiftDictionary.cpp
@@ -26,95 +26,42 @@ using namespace lldb_private;
 using namespace lldb_private::formatters;
 using namespace lldb_private::formatters::swift;
 
-namespace lldb_private {
-namespace formatters {
-namespace swift {
-class SwiftDictionaryNativeBufferHandler
-    : public SwiftHashedContainerNativeBufferHandler {
-public:
-  SwiftHashedContainerBufferHandler::Kind GetKind() {
-    return Kind::eDictionary;
-  }
-
-  static ConstString GetMangledStorageTypeName();
-
-  static ConstString GetDemangledStorageTypeName();
-
-  SwiftDictionaryNativeBufferHandler(ValueObjectSP nativeStorage_sp,
-                                     CompilerType key_type,
-                                     CompilerType value_type)
-      : SwiftHashedContainerNativeBufferHandler(nativeStorage_sp, key_type,
-                                                value_type) {}
-  friend class SwiftHashedContainerBufferHandler;
-
-private:
-};
-
-class SwiftDictionarySyntheticFrontEndBufferHandler
-    : public SwiftHashedContainerSyntheticFrontEndBufferHandler {
-public:
-  SwiftHashedContainerBufferHandler::Kind GetKind() {
-    return Kind::eDictionary;
-  }
-
-  virtual ~SwiftDictionarySyntheticFrontEndBufferHandler() {}
-
-  SwiftDictionarySyntheticFrontEndBufferHandler(lldb::ValueObjectSP valobj_sp)
-      : SwiftHashedContainerSyntheticFrontEndBufferHandler(valobj_sp) {}
-  friend class SwiftHashedContainerBufferHandler;
-
-private:
-};
-
-class DictionarySyntheticFrontEnd : public HashedContainerSyntheticFrontEnd {
-public:
-  DictionarySyntheticFrontEnd(lldb::ValueObjectSP valobj_sp)
-      : HashedContainerSyntheticFrontEnd(valobj_sp) {}
-
-  virtual bool Update();
-
-  virtual ~DictionarySyntheticFrontEnd() = default;
-};
-}
-}
+const DictionaryConfig &
+DictionaryConfig::Get() {
+  static DictionaryConfig g_config{};
+  return g_config;
 }
 
-SyntheticChildrenFrontEnd *
-lldb_private::formatters::swift::DictionarySyntheticFrontEndCreator(
-    CXXSyntheticChildren *, lldb::ValueObjectSP valobj_sp) {
-  if (!valobj_sp)
-    return NULL;
-  return (new DictionarySyntheticFrontEnd(valobj_sp));
+DictionaryConfig::DictionaryConfig()
+  : HashedCollectionConfig() {
+  m_summaryProviderName =
+    ConstString("Swift.Dictionary summary provider");
+  m_syntheticChildrenName =
+    ConstString("Swift.Dictionary synthetic children");
+  m_collection_demangledRegex =
+    ConstString("^Swift\\.Dictionary<.+,.+>$");
+
+  // Native storage class
+  m_nativeStorage_demangledPrefix =
+    ConstString("Swift._HashableTypedNativeDictionaryStorage<");
+  m_nativeStorage_demangledRegex =
+    ConstString("^Swift\\._HashableTypedNativeDictionaryStorage<.+,.+>$");
+
+  // Type-punned empty dictionary
+  m_emptyStorage_demangled
+    = ConstString("Swift._RawNativeDictionaryStorage");
+
+  // Deferred non-verbatim bridged dictionary
+  m_deferredBridgedStorage_demangledPrefix
+    = ConstString("Swift._SwiftDeferredNSDictionary<");
+  m_deferredBridgedStorage_demangledRegex
+    = ConstString("^Swift\\._SwiftDeferredNSDictionary<.+,.+>$");
 }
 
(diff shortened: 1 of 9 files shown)


lorentey:
@swift-ci please test

source/Plugins/Language/Swift/SwiftDictionary.cpp
```diff
@@ -26,95 +26,48 @@ using namespace lldb_private;
...
+  m_summaryProviderName =
+    ConstString("Swift.Dictionary summary provider");
+  m_syntheticChildrenName =
+    ConstString("Swift.Dictionary synthetic children");
+  m_collection_demangledRegex =
+    ConstString("^Swift\\.Dictionary<.+,.+>$");
+
+  // Native storage class
+  m_nativeStorage_mangledPrefix =
+    ConstString(SwiftLanguageRuntime::GetCurrentMangledName("_TtGCs37_HashableTypedNativeDictionaryStorage"));
```
dcci: You can probably use the new mangled name here.

source/Plugins/Language/ObjC/NSArray.cpp
```diff
@@ -35,17 +35,16 @@ using namespace lldb_private::formatters;
 
 namespace lldb_private {
 namespace formatters {
-std::map<ConstString, CXXFunctionSummaryFormat::Callback> &
+NSArray_Additionals::SummaryFormatters &
```
dcci: This looks like it's code in `ObjC/`. Is this Swift specific? If so, maybe we should move it to a `Swift/` specific directory to reduce the amount of conflicts with upstream. Otherwise, if this might be useful for `NSArray` et similia also when running in `ObjC` language mode, this patch (or at least, part of it) should live upstream.

dcci (commented):
I had a quick glance at this patch and the swift bits look OK. I have a more general question about the structure of this patch before going in the details, see comment(s) inline.

source/Plugins/Language/Swift/SwiftDictionary.cpp
lorentey: We apparently still use this mangling for the Objective-C runtime names of NSObject subclasses defined in Swift.

The name is incomplete, so while I cargo-culted GetCurrentMangledName over from the old code, it actually has no effect.

source/Plugins/Language/ObjC/NSArray.cpp
lorentey: These are hooks that Swift inserted so that it can hook into the Objective-C mode's formatter dispatch for Swift's implementations of Foundation collections. ObjC could use these to register its own formatters, too, but it doesn't really need it -- the if ladders it currently has work fine.

The new ObjC formatter (9b0178f172c284a507559d022374cead013fb152) and related NSSet formatter fixes (790e5f6cb1a954b0c778a54568932f6f198d11c6) are entirely Objective-C specific and should definitely live upstream!

lorentey: Ah, wait, I see the hooks are also in the upstream repo. You're right, the Objective-C parts of 66cbe5880522688825c93b91c51470c0aea2149a should also get merged there!

dcci: I'd very much recommend to just submit the patch upstream (I'll be happy to review it!) and then cherry-pick here (at least, this is the flow we've been following for a while).

lorentey:
@swift-ci test

lorentey:
@swift-ci test

lorentey:
@swift-ci test

lorentey:
@dcci I rebased this & restructured the code, removing the parts that touch upstream code. I isolated the NSSet formatter fixes to a separate branch; I'll submit them upstream next week.

It turns out we don't need to hook into Foundation's formatter dispatch; AFAICT, merely registering the Swift formatters with the Objective-C class names works fine, as long as the formatters are prepared to take ObjC pointers as well as Swift values. (The value conversion in `HashedCollectionConfig::SwiftObjectAtAddress` is the messy bit.)

lorentey:
```
FAIL: test_dictionary_nsobject_any_object_dwarf (TestSwiftDictionaryNSObjectAnyObject.TestDictionaryNSObjectAnyObject)
AssertionError: False is not True : 'frame variable -d run -- d2' returns expected result, 
got '([NSObject : AnyObject]) d2 = 0 key/value pairs {}'
```

lldb/packages/Python/lldbsuite/test/lang/swift/variables/dict_nsobj_anyobj/main.swift:
```
  var d1: Dictionary<Int,Int> = [1:1,2:2,3:3,4:4]
  var d2: Dictionary<NSObject,AnyObject> = [1 as NSNumber : 1 as NSNumber,
                                            2 as NSNumber : 2 as NSNumber,
                                            3 as NSNumber : 3 as NSNumber,
                                            4 as NSNumber : 4 as NSNumber]
```

Hm. d1 seems to work fine, but we fail to get the dynamic type of d2's storage reference -- it somehow resolves to its static type, `Swift._RawNativeDictionaryStorage`. This base class is used for the empty singleton instance, so we get on that path, and print d2 as having 0 elements. Weird; I don't know why the dynamic type fails to resolve in this particular case.

Reverting 2476a02c903d23b8a1757abfcb9c3fe48be12773 fixes the issue, but it removes dynamic isa checks for Dictionary variables -- so uninitialized values may cause confusion more often. (This is not a regression; lldb wasn't verifying the storage isa before this PR, either.) I think this is fine for now: the PR already improves storage layout verification elsewhere.

lorentey:
@swift-ci test

lorentey:
@swift-ci test

lorentey:
@swift-ci please test

dcci:
This looks great, thanks (sorry for the delay in replying, I was OOO last week)

PR #814, opened 2018-08-09
[lldb-mi] Disable flakey pexpect tests on linux as well

JDevlieghere:
We previously disabled these two tests on Darwin, but disable them on
Linux as well as they're faleky there too.


JDevlieghere:

PR #815, opened 2018-08-09
Prevent the Swift scratch AST context to be reset while it is being u…

adrian-prantl:
Soliciting early feedback to the approach before going down too far the rabbit hole.

Prevent the Swift scratch AST context to be reset while it is being used.
    
This patch introduces a Reader-Writer-Lock around the scratch
context. Users of a scratch context acquire a read lock while they are
depending on a scratch context. Only if no readers a around is Target
allowed to reset a Swift scratch context and delete the old one.
    
The Swift scratch context may need to be replaced when it gets corrupted,
for example due to incompatible ClangImporter options. This locking
mechanism guarantees that this won't happen while a client is using the
context.
    
In Swift there are three use-cases for ASTContexts with
different requirements and guarantees.
    
- Module ASTContexts are used for the static type system. They
  are created once for each lldb::Module and live forever.
    
- Scratch AST Contexts are used for expressions
  (thus far everything is like in the Clang language support)
    
- Scratch AST Contexts are also used to express the results of
   any dynamic type resolution done by RemoteAST or Archetype
   binding.
    
Because expressions and dynamic type resolution may trigger the
import of another module, the scratch context may become
unusable. When a scratch context is in a fatal error state,
GetScratchSwiftASTContext() will create a fresh global context,
or even separate scratch contexts for each lldb::Module. But it
will only do this if no client holds on to a read lock on \b
m_scratch_typesystem_lock.

<rdar://problem/42399818>

### DIFF
--- a/include/lldb/Core/SwiftASTContextReader.h
+++ b/include/lldb/Core/SwiftASTContextReader.h
@@ -1,0 +1,104 @@
+//===-- SwiftASTContextReader.h ---------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2018 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef liblldb_SwiftASTContextReader_h_
+#define liblldb_SwiftASTContextReader_h_
+
+#include "llvm/Support/RWMutex.h"
+
+namespace lldb_private {
+
+class SwiftASTContext;
+class ExecutionContext;
+class ExecutionContextRef;
+
+/// This is like llvm::sys::SmartRWMutex<false>, but with a try_lock method.
+///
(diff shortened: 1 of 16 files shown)


adrian-prantl:
For those following at home, @jimingham and I discussed this at length in person and convinced ourselves that this *is* a reasonable approach. (But I need to detect that the scratch context has changed in ValueObjectDynamicType).

jimingham:
I haven't gone through the code in detail yet, but I think this approach looks good.

The only common case this won't cover that I can think of is where a stored CompilerType ends up pointing to an invalidated scratch AST Context - which should only happen in ValueObjectDynamic values since they are the only ones that would pull their types from a scratch AST context.  We talked about getting UpdateValueIfNeeded to check this case.

Note, you should also be able to revert 4254d862f02b9397003a09f4bae1d6dbc7822c77 which was a pretty ugly workaround for this problem.

adrian-prantl:
I added the `UpdateValueIfNeeded()` code in the latest update.

adrian-prantl:
@swift-ci test

adrian-prantl:
I added asserts in the Swift dynamic type lookup functions to test that the locking actually works as intended, and fixed a couple of bugs that where discovered through the new assertions.

adrian-prantl:
@swift-ci test

PR #818, opened 2018-08-11
 Perform Archetype binding in a scratch context.

adrian-prantl:
The binding of archetypes is just another form of dynamic type
resolution and as such it can trigger an import of another
module. This operation is only safe in a scratch context.

This patch also changes MetadataPromises use the appropriate scratch context.

Fixes <rdar://problem/42399818>


adrian-prantl:
@swift-ci test and merge

PR #820, opened 2018-08-14
Fix an LLDB Swift expression evaluator bug that prevents any

adrian-prantl:
expressions from being run that use API that is @available in an SDK
with a major version number >10, while debugging in a simulator.

LLDB knows how to correctly parse the deployment target out of the
load commands. This fixed a critical performance problem. To still be
able to run expressions containing API newer than the deployment
target in LLDB, we started marking up LLDB expressions with
availability annotations. The code to compute the SDK version used in
the availability markup had a bug for iOS devices. We fixed
that bug in b34a153ab9d. Unfortunately the bugfix that fixed
devices introduced a regression for simulators that caused the host
macOS version instead of the simulator version to be used when
debugging on a simulator. This makes it impossible to run any
expressions that contain API from an iOS/tvOS version greater than
10.13.

rdar://problem/42884412

### DIFF
--- a/source/Expression/ExpressionSourceCode.cpp
+++ b/source/Expression/ExpressionSourceCode.cpp
@@ -485,10 +485,22 @@ bool ExpressionSourceCode::GetText(
       auto triple = arch_spec.GetTriple();
       if (triple.isOSDarwin()) {
         if (auto process_sp = exe_ctx.GetProcessSP()) {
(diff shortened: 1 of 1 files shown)


adrian-prantl:
@swift-ci test and merge

fredriss (approved):

adrian-prantl:
@swift-ci test and merge

PR #822, opened 2018-08-14
[4.2] [REPL] Set the correct rpath when installing.

shahmishal:
$ORIGIN represents the location of the REPL at runtime,
whenever the binary is located. So, this makes sure
that when we generate the toolchain on one machine and
transfer to another, the repl keeps working, as it needs
to `dlopen` libSwiftCore, which sits in a path relative
to the executable itself.

<rdar://problem/42863249>


shahmishal:
@swift-ci test

fredriss (approved):

PR #833, opened 2018-08-15
Fix a little thinko in generating ___lldb_unnamed_symbol symbols

jimingham:
when we have only an in-memory copy of the binary.

Also added a test for the generation of these symbols in the 
in-memory and regular cases.

<rdar://problem/43160401>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@339833 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #836, opened 2018-08-16
 Add a Swift backwards-compatibility test category.

adrian-prantl:
These tests contains a subset of Swift that is expected to work in
LLDB even when compiled with (slightly) older versions of the Swift
compiler.

<rdar://problem/36031662>


adrian-prantl:
@swift-ci test and merge

PR #841, opened 2018-08-20
[SwiftLanguageRuntime] Map into context before IRGen'ing.

dcci:
IRGen expects the type to be fully realized, and asserts
otherwise. Note that this has been broken for quite a while,
we just didn't crash but printed garbage before because IRGen
was less strict.

<rdar://problem/39050802>


dcci:
@swift-ci test and merge

PR #849, opened 2018-08-22
fix ambiguous reference when you redefine something

marcrasi:
This may not be quite the right way to fix this, so I'm posting this PR to demonstrate what I figured out, and to get advice on how to proceed.

## Example of bug

```
  1> let x = 1; print(x)
1
x: Int = 1
  2> let x = 2; print(x)
error: repl.swift:2:18: error: ambiguous use of 'x'
let x = 2; print(x)
                 ^

repl.swift:2:5: note: found this candidate
let x = 2; print(x)
    ^
```

## My discoveries so far
* `LLDBNameLookup::lookupAdditions` is a function responsible for looking up decls from previous lines (aka "persistent decls").
* When line 2 of the above example executes, the parser finds the line-2-declaration of `x` before calling `lookupAdditions`. `lookupAdditions` proceeds to find the line-1-declaration of `x`, and adds it to the set of results. Now there are 2 declarations and we get the ambiguous use error.
* `lookupAdditions` already contains some logic that looks intended to prevent it from adding the line-1-declaration of `x` to `RV`. Specifically, decls present in `m_staged_decls` take precedence over decls from `m_persistent_decls`. It looks like `m_staged_decls` is intended to contain decls made in the currently-compiling code. Unfortunately, `m_staged_decls` is empty when `lookupAdditions` gets called in the above example.

## My fix

My fix to `lookupAdditions` works by adding the currently-compiling code's decls to `results` before `lookupAdditions` proceeds to add persistent decls. This causes the currently-compiling code's decls to take precedence.

I'm unsure if the decls that my fix adds should have been in `m_staged_decls` in the first place. If so, then a better fix would be to fix the problem that's preventing them from getting into `m_staged_decls`. @jimingham, do you know if decls should show up in `m_staged_decls`? If so, do you have any idea why they're not?

## Some m_staged_decls tracing

`DebuggerContextChange` is the only thing that adds `m_staged_decls`, [by calling debug_client->didGlobalize()](https://github.com/apple/swift/blob/98e95bdbef20ee33b7efba4f1d12cb5ae5ac81e2/lib/Parse/ParseDecl.cpp#L175). But this never actually happens in the above example for at least two reasons:
1. parseDeclVar never instantiates a DebuggerContextChange
2. The DebuggerContextChange wouldn't do anything, because [DebuggerContextChange::inDebuggerContext](https://github.com/apple/swift/blob/98e95bdbef20ee33b7efba4f1d12cb5ae5ac81e2/lib/Parse/ParseDecl.cpp#L150) is returning false.

## Links

Discussed in this forum post: https://forums.swift.org/t/implementing-a-jupyter-kernel-for-swift-using-lldb-questions-about-declaration-behavior-and-code-completion/14513

JIRA: https://bugs.swift.org/browse/SR-8289

### DIFF
--- a/packages/Python/lldbsuite/test/repl/redefinition/TestRedefinition.py
+++ b/packages/Python/lldbsuite/test/repl/redefinition/TestRedefinition.py
@@ -1,0 +1,58 @@
+# TestREPLBreakpoints.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""Test that redefining things in the repl works as desired."""
+
+import lldbsuite.test.lldbrepl as lldbrepl
+
+class TestRedefinition (lldbrepl.REPLTest):
+
+    mydir = lldbrepl.REPLTest.compute_mydir(__file__)
+
+    def doTest(self):
+        # Define some *DefinedInREPL things.
+        self.command('var varDefinedInREPL: Int = 1')
+        self.command('func funcDefinedInREPL() -> Int { return 1 }')
+        self.command('struct StructDefinedInREPL { var field: Int = 1 }')
+
+        # Assert that the REPL can refer to the *DefinedInRepl things.
+        self.command('varDefinedInREPL', patterns=[r'\$R[0-9]+: Int = 1'])
+        self.command('funcDefinedInREPL()', patterns=[r'\$R[0-9]+: Int = 1'])
+        self.command('StructDefinedInREPL().field',
+                     patterns=[r'\$R[0-9]+: Int = 1'])
+
+        # Redefine the *DefinedInREPL things.
+        self.command('var varDefinedInREPL: Int = 2')
+        self.command('func funcDefinedInREPL() -> Int { return 2 }')
+        self.command('struct StructDefinedInREPL { var field: Int = 2 }')
+
+        # Assert that the REPL refers to the redefinitions of the
+        # *DefinedInREPL things.
+        self.command('varDefinedInREPL', patterns=[r'\$R[0-9]+: Int = 2'])
+        self.command('funcDefinedInREPL()', patterns=[r'\$R[0-9]+: Int = 2'])
+        self.command('StructDefinedInREPL().field',
+                     patterns=[r'\$R[0-9]+: Int = 2'])
+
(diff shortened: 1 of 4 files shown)


dcci:
@jimingham I guess you're the right person to review this.

jimingham:
The complexities in the LLDBNameLookup mechanism are really all for uses of the lldb expression parser when debugging some other program (i.e. the "expr" command, not "expr --repl" or other ways of getting to the REPL.)  

In that simple "expr" case, there are lots of names that swift can't find out about from its imported modules in the expression that you are submitting (e.g. local variables, local types, types from foreign modules that the user happens to want to refer to...) and lldb has to provide them.

It also deals differently with variables defined by expressions - in the REPL we aren't wrapping the expression in a "context emulating" containing function - and so shouldn't need to hoist variables out of this context to make them globally available.

But that should never be true in the REPL.  Everything either comes from the modules imported from previous sessions, or from entities made in the previous REPL statements (which are found by this in lookupAdditions:

        m_persistent_vars->GetSwiftPersistentDecls(name_const_str, results);

That and the little bit at the end where the results are added to RV are the only active parts of lookupAdditions for the REPL.  All the Globalize stuff is not relevant, nor are the staged decls...

Given that the two environments are so different, and the REPL so much simpler, I think it would be clearer not to try to merge the two, but to make an LLDBREPLNameLookup and use that for the REPL.  Have it just look in the SwiftPersistentDecls and then add the little bit of logic that you suggest to fix the REPL issue. 

Then you can remove the check for REPLEnabled (and PlaygroundEnabled) and simplify the normal "expr" command LLDBNameLookup as well.

marcrasi:
Thanks for the explanation and suggestion! I tried out your suggestion of having a separate `LLDBREPLNameLookup` and it seems to work pretty well.

There is a function `emitLValueForVariable` that I think makes sense to share between the two NameLookup classes, so I created a base class `LLDBNameLookup` and two subclasses `LLDBREPLNameLookup` and `LLDBExprNameLookup`.

I also changed the signature of `GetSwiftPersistentDecls` a bit to make my new REPL lookup function slightly easier to implement. The change isn't essential, so I could avoid it if you don't like it.

Could you take a look and let me know what you think?

jimingham:
For the most part this looks fine.

I don't mind the addition of the exclusion vector to GetSwiftPersistentDecls.  It does mean that we make a copy of whatever Swift had already found in the case where you pass in the result vector as the exclusion vector (at line 581).  Why not pass both vectors by reference, and then handle the case where they are in fact the same vector by getting the incoming number of exclusion results and just iterating over that range?  That would both avoid the copy and handle the case where the two vectors are the same.

marcrasi:
Yep, I can pass by reference and iterate over only the incoming number. I have modified it to do that.

jimingham (approved):
Looks great!  This should also make it easier to fix other REPL specific lookup issues w/o over-complicating the logic.  Thanks for doing this!

marcrasi:

PR #850, opened 2018-08-22
[TestNSError] Temporarily skip while I investigate.

dcci:
<rdar://problem/43569527>


dcci:
@swift-ci test and merge

PR #851, opened 2018-08-23
[SwiftASTContext] Minor cleanups for GetTypeFromMangledTypename.

dcci:
-> Reduce indentation for slightly easier readability.
-> Use the correct mangling prefix in the error ($S and not _T)
-> Remove else after return as it's not needed.


dcci:
@swift-ci test and merge

PR #855, opened 2018-08-23
 Remove unnecessary code.

adrian-prantl:
The implementation searches for absolute paths to .swiftmodule files
in FileSpecs. This is not how we are supposed to find
.swiftmodules. We are supposed to find them via N_AST symbols, or via
dsymutil inside the ast section of the .dSYM.

rdar://problem/39962565

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1242,18 +1242,6 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
         }
       }
     }
-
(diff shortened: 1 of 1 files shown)


adrian-prantl:
@swift-ci test

jimingham (approved):
That code has been there forever.  It was probably from the early days when a lot more finding of modules was done by hand in lldb, so we shouldn't still need it.  And being more careful about what search paths we add is going to make debugging more deterministic, which is a good thing!

If we do end up finding out that we still needed this in some odd circumstance, we should find a better way to handle that case.  This way is overly broad.

dcci (approved):

PR #856, opened 2018-08-24
Fix crash in Set/Dictionary formatters

lorentey:
This improves error handling in Swift/Dictionary formatters, and makes `id` -> `AnyObject` conversion more robust.

Fixes rdar://problem/43569527

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -294,6 +294,8 @@ public:
   CompilerType GetTypeFromMangledTypename(const char *mangled_typename,
                                           Status &error);
 
+  CompilerType GetAnyObjectType();
+
   // Get a function type that returns nothing and take no parameters
   CompilerType GetVoidFunctionType();
 
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -294,6 +294,7 @@ public:
   CompilerType GetTypeFromMangledTypename(const char *mangled_typename,
                                           Status &error);
 
+  // Retrieve the Swift.AnyObject type.
   CompilerType GetAnyObjectType();
 
   // Get a function type that returns nothing and take no parameters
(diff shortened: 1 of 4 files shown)


lorentey:
@swift-ci please test

lorentey:
Ah, I forgot to re-enable the NSError test. I'll push a new commit to do so once the tests finish.

include/lldb/Symbol/SwiftASTContext.h
```diff
@@ -294,6 +294,8 @@ class SwiftASTContext : public TypeSystem {
   CompilerType GetTypeFromMangledTypename(const char *mangled_typename,
                                           Status &error);
 
+  CompilerType GetAnyObjectType();
```
dcci: Can you add a comment for this member function please?

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -292,7 +292,7 @@ HashedCollectionConfig::SwiftObjectAtAddress(
   if (error.Fail())
     return nullptr;
 
-  CompilerType anyObject_type = ast_ctx->FindQualifiedType("Swift.AnyObject");
+  CompilerType anyObject_type = ast_ctx->GetAnyObjectType();
```
dcci, line 295: I think this is objectively better than looking up the object by name. As a follow-up, I'm going to see whether we do this name-based lookup somewhere else and switch to your method.

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -543,8 +543,10 @@ HashedCollectionConfig::CreateHandler(ValueObject &valobj) const {
   ValueObjectSP valobj_sp = valobj.GetSP();
   if (valobj_sp->GetObjectRuntimeLanguage() != eLanguageTypeSwift &&
       valobj_sp->IsPointerType()) {
-    valobj_sp = SwiftObjectAtAddress(valobj_sp->GetExecutionContextRef(),
-                                     valobj_sp->GetPointerValue());
+    if (auto swiftval_sp = SwiftObjectAtAddress(
```
dcci, line 546: This looks about right.

dcci (approved):
Please re-enable the test, otherwise I guess we're good.

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
lorentey: I think I was the sole perpetrator of such nonsense. (I had a mental note to get rid of it before submitting #813 but apparently left it in..)

lorentey:

PR #858, opened 2018-08-24
[TestNSError] Update and remove expected failure

lorentey:
`TestNSError`'s actual output looks okay to me now. Update expectations to match reality and remove expected failure from this test. Discrepancies seem due to integers being printed as `3` and `4` instead of `Int64(3)` and `Int64(4)`.

For reference, this is the current output of this test case:

```
(lldb) frame variable -d run --ptr-depth=2 
(String, NSError) call1 = {
  0 = " "
  1 = 0x0000000101700880 domain: "lldbrocks" - code: 3133079277 {
    _userInfo = 2 key/value pairs {
      [0] = {
        key = "y"
        value = 0
      }
      [1] = {
        key = "x"
        value = 0
      }
    }
  }
}
(String, NSError) call2 = {
  0 = "x+y"
  1 = 0x0000000101700d70 domain: "lldbrocks" - code: 0 {
    _userInfo = 2 key/value pairs {
      [0] = {
        key = "y"
        value = 4
      }
      [1] = {
        key = "x"
        value = 3
      }
    }
  }
}
```

Note that hash table ordering is unstable, so exact string matches won't work.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/nserror/TestNSError.py
+++ b/packages/Python/lldbsuite/test/lang/swift/nserror/TestNSError.py
@@ -25,8 +25,6 @@ class SwiftNSErrorTest(TestBase):
 
     @decorators.skipUnlessDarwin
     @decorators.swiftTest
-    @decorators.expectedFailureAll(
-        bugnumber="https://bugs.swift.org/browse/SR-782")
     def test_swift_nserror(self):
         """Tests that Swift displays NSError correctly"""
         self.build()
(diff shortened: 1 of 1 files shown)


lorentey:
I'm on the road, so I won't be able to do much local testing for this. Let's see if I manage to pass PR tests on the first attempt.

@swift-ci test

dcci (approved):
This was just a test not properly updated. As you're on the road, I'm taking the freedom to merge on your behalf after approving. Thanks!

PR #860, opened 2018-08-27
[REPL] Test the new optimized layout for Optional<T>

dcci:
where <T> is a struct with an extra-inhabitant-bearing field
anywhere in its definition, not only in the beginning.

Add a test in lldb to make sure the formatter is right for
Optional (and won't bitrot over time).

<rdar://problem/19690160>

### DIFF
--- a/lit/SwiftREPL/Optional.test
+++ b/lit/SwiftREPL/Optional.test
@@ -1,0 +1,22 @@
+// RUN: %lldb --repl < %s | FileCheck %s
+
+enum Patatino {
+  case first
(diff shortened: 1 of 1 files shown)


dcci:
@jckarter hope this makes sense to you.

dcci:
@swift-ci test and merge

jckarter (approved):
Makes sense, thanks!

dcci:
Thanks for your review, Joe.

dcci:

PR #866, opened 2018-08-30
Avoid an llvm::unreachable in the Swift compiler

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #869, opened 2018-08-31
File completion bugfix

fredriss:
If you tried to complete somwthing like ~/., lldb would come up with a lot
of non-existent filenames by concatenating every exisitng file in the directory
with an initial '.'.

This was due to a workaround for an llvm::fs::path::filename behavior that
was not applied selectively enough.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@341268 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 733a51f0a31cc32a3a73027df376777f6868c951)


fredriss:
@swift-ci test and merge

PR #874, opened 2018-09-05
Refactor the SwiftLanguageRuntime::GetDynamicTypeAndAddress.* family …

adrian-prantl:
…of functions.

This pulls out common functionionality such as performing archetype
binding and creating the scratch into the outmost function and
simplifies the code by converting to early exists where this makes
sense.

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -366,46 +366,58 @@ protected:
                                 const CompilerType &dynamic_type,
                                 bool is_indirect_enum_case);
 
-  virtual bool GetDynamicTypeAndAddress_Class(
-      ValueObject &in_value, lldb::DynamicValueType use_dynamic,
-      TypeAndOrName &class_type_or_name, Address &address);
+  bool GetDynamicTypeAndAddress_Class(ValueObject &in_value,
+                                      SwiftASTContext &scratch_ctx,
+                                      lldb::DynamicValueType use_dynamic,
+                                      TypeAndOrName &class_type_or_name,
+                                      Address &address);
 
-  virtual bool GetDynamicTypeAndAddress_Protocol(
-      ValueObject &in_value, lldb::DynamicValueType use_dynamic,
-      TypeAndOrName &class_type_or_name, Address &address);
+  bool GetDynamicTypeAndAddress_Protocol(ValueObject &in_value,
+                                         SwiftASTContext &scratch_ctx,
+                                         lldb::DynamicValueType use_dynamic,
+                                         TypeAndOrName &class_type_or_name,
+                                         Address &address);
 
-  virtual bool GetDynamicTypeAndAddress_ErrorType(
-      ValueObject &in_value, lldb::DynamicValueType use_dynamic,
-      TypeAndOrName &class_type_or_name, Address &address);
+  bool GetDynamicTypeAndAddress_ErrorType(ValueObject &in_value,
+                                          lldb::DynamicValueType use_dynamic,
+                                          TypeAndOrName &class_type_or_name,
+                                          Address &address);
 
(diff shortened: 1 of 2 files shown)


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

include/lldb/Target/SwiftLanguageRuntime.h
```diff
@@ -366,46 +366,58 @@ class SwiftLanguageRuntime : public LanguageRuntime {
                                 const CompilerType &dynamic_type,
                                 bool is_indirect_enum_case);
 
-  virtual bool GetDynamicTypeAndAddress_Class(
```
dcci, line 369: Thanks for removing the `virtual` here!

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2479,59 +2418,89 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress(
...
-  if (SwiftASTContext *swift_ast = llvm::dyn_cast_or_null<SwiftASTContext>(
-          in_value.GetCompilerType().GetTypeSystem())) {
-    if (swift_ast->HasFatalErrors() || !swift_ast->GetClangImporter()) {
-      return false;
+  // Dynamic type resolution in RemoteAST might pull in other Swift modules, so
+  // use the scratch context where such operations are legal and safe.
+  assert(IsScratchContextLocked(in_value.GetTargetSP()) &&
+         "Swift scratch context not locked ahead of dynamic type resolution");
+  auto scratch_ctx = in_value.GetScratchSwiftASTContext();
+  if (!scratch_ctx)
```
dcci, line 2429: Can we actually get here? We just asserted that we locked a scratchASTContext so this check seems redundant?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2479,59 +2418,89 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress(
...
-        success = GetDynamicTypeAndAddress_Struct(in_value, use_dynamic,
-                                                  class_type_or_name, address);
-      else if (type_info.AllSet(eTypeIsBuiltIn | eTypeIsPointer |
-                                eTypeHasValue))
-        success = GetDynamicTypeAndAddress_Class(in_value, use_dynamic,
-                                                 class_type_or_name, address);
-    }
+    // Perform archetype binding in the scratch context.
+    auto *frame = in_value.GetExecutionContextRef().GetFrameSP().get();
+    if (!frame)
```
dcci, line 2483: We always have a frame here, I think.

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: We can get there if target==nullptr (not sure if that is possible, but perhaps when evaluating expressions without creating a target) or if the valueobject doesn't have a valid module.

Either of those cases seem really unlikely, but it I can't prove they are impossible.

dcci: Fair, better safe than sorry (here I think it's better being defensive).

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: Why? Couldn't in_value be an out-of-date value object whose frame doesn't exist any more?

dcci: My understanding is that we never get to dynamic type resolution if our frame is empty (there are several layers of `if (!frame) return false`, IIRC, but I can't (as well!) prove all the paths converging here have the check.

dcci (approved):
LGTM

PR #875, opened 2018-09-06
Re-instate a bit of code that was commented out in r188246 which

jasonmolenda:
Re-instate a bit of code that was commented out in r188246 which
reads an ObjectFileMachO's string table in one chunk.  Originally
this was commented out because binaries in the system's shared cache
all share a mega-string table and so reading the entire mega-strtab
for each binary was a performance problem.

In the reinstated code, I add a check that the binary we're reading
from memory is not in the shared cache (there isn't a constant in
<mach-o/loader.h> for this bit yet; we hardcode the value in one
other place in ObjectFileMachO alread).  For binaries that we're
reading out of memory that are NOT in the shared cache, reading 
the string table in one chunk is a big performance improvement.

Also have debugserver send up the flags value for binaries in its
response to the jGetLoadedDynamicLibrariesInfos request.

NFC.

<rdar://problem/33604496> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@341511 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit a8d7125bb8eb189cd6a0edc0c1fb43f3f3d511c5)
(cherry picked from commit a8e4762ee4dcc92ba23918ade5d96bafc187cf12)

### DIFF
--- a/source/Plugins/ObjectFile/Mach-O/ObjectFileMachO.cpp
+++ b/source/Plugins/ObjectFile/Mach-O/ObjectFileMachO.cpp
@@ -2457,14 +2457,6 @@ size_t ObjectFileMachO::ParseSymtab() {
             if (nlist_data_sp)
               nlist_data.SetData(nlist_data_sp, 0,
                                  nlist_data_sp->GetByteSize());
-            // Load strings individually from memory when loading from memory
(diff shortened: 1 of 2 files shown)


jasonmolenda:
@swift-ci test and merge

PR #878, opened 2018-09-06
Print column info in backtraces et al. if available

adrian-prantl:
This patch allows LLDB to print column info in backtraces et al. if
available, which is useful when the backtrace contains a frame like
the following:

  f(can_crash(0), can_crash(1));

Differential Revision: https://reviews.llvm.org/D51661

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@341506 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 91d709c7af26b9fa394ae7a20e95e721044b8703)

### DIFF
--- a/include/lldb/Core/FormatEntity.h
+++ b/include/lldb/Core/FormatEntity.h
@@ -104,6 +104,7 @@ public:
       FunctionIsOptimized,
       LineEntryFile,
       LineEntryLineNumber,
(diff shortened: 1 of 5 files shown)


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

PR #880, opened 2018-09-07
Fix a typo in mac SIP workaround

adrian-prantl:
presumably the or subexpression was meant to be evaluated first. The way
it is now means that we apply the workaround for any python in `/usr`,
which matches pretty much every unix system.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@341167 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 8bcb5e7f073ebc8f0e93d3666c83672fafe43243)


adrian-prantl:
@swift-ci test and merge

PR #886, opened 2018-09-10
Make SwiftASTContext::EitherComparator::operator() const.

allevato:
Our build of swift-lldb within Google is transitioning to libc++,
which is stricter than libstdc++ about the const-ness of comparators
used with std collections.

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -57,7 +57,7 @@ public:
 
 private:
   struct EitherComparator {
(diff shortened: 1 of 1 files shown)


dcci (approved):

dcci:

dcci:
FWIW, I build all the time with libc++ here, but I haven't seen this. It doesn't hurt being explicit though.

allevato:
Which version of C++? This coincided with some C++17 migrations on our end; I don't know if the issue existed in older versions of libc++.

Either way, thanks for the quick review!

dcci:
I always build with C++11.

dcci:
Or well, at least I never built with C++17.

PR #888, opened 2018-09-10
[SwiftLanguageRuntime] Fix the error return register on AArch64.

dcci:
No test as this fixes TestExpressionErrorBreakpoint there.

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -3881,7 +3881,7 @@ llvm::Optional<Value> SwiftLanguageRuntime::GetErrorReturnLocationAfterReturn(
       error_reg_name = "r6";
       break;
     case llvm::Triple::ArchType::aarch64:
-      error_reg_name = "x19";
+      error_reg_name = "x21";
       break;
     case llvm::Triple::ArchType::x86_64:
       error_reg_name = "r12";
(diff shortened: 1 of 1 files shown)


dcci:
I didn't run the test suite myself, just tried a similar test program to the one in the test. Close enough I guess.

dcci:
@swift-ci test and merge

dcci:
For future reference (mainly for myself), the values are documented in LLVM (e.g.

```
 $ grep  SwiftError lib/Target/AArch64/AArch64CallingConvention.td
```

thanks to @aschwaighofer for pointing out)

jrose-apple:
Is there any way to get these values from the various target definitions, rather than copying them here?

dcci:
@jrose-apple yes, there should be one (this hardcoding makes me uneasy too). 
How does swiftc access them?

jrose-apple:
swiftc just lets LLVM figure it out; it doesn't hardcode anything. I think you'll have to ask LLVM for the target's calling convention info yourself, something that's out of my area of expertise.

dcci:
@aschwaighofer / @mikeash / @jckarter any ideas on how we make this more robust in face of changes? I do understand that having ABI-stable binaries make this less important, but still I think this information should be entirely opaque to lldb.

mikeash:
In theory, I like the idea of consolidating the information. In practice, I think it's reasonable for some duplicated information like this to exist in the debugger.

Would it be possible to have lldb consume the information from the .td file? This document implies it should be possible to make an arbitrary backend for the .td parser, and maybe lldb could have one that emits the information it needs. https://llvm.org/docs/TableGen/index.html

PR #890, opened 2018-09-12
[SwiftLanguageRuntime] Catch up with changes for swift_willThrow.

dcci:
The compiler as size optimizations stores the error object in the
error return register. This commits update(s) lldb to look in the
right place (before, it was the first argument to the function).

<rdar://problem/44303855>


dcci:
The fallout was my fault, I should've known better at this point.

dcci:
@swift-ci test and merge

dcci:
Ah, if anybody is wondering. No test because this fixes a (currently) broken test.

PR #892, opened 2018-09-13
[REPL] Test the repl directly here.

dcci:
Jim is investigating what's wrong with the SBAPI in the meanwhile, so that we can re-enable the python test. For now, let's try to not regress functionality (and it's always good to test the REPL directly anyway).

<rdar://problem/44428726>


dcci:
@swift-ci test and merge

PR #894, opened 2018-09-14
[DWARFExpression] Read literars as unsigned values.

JDevlieghere:
After landing r341457, we started seeing a failure on the swift-lldb
bots. The change was correct and pretty straightforward, a DW_OP_constu
was replaced with DW_OP_lit23, the value remaining identical.

  0x000000f4: DW_TAG_variable
		DW_AT_location    (0x00000000
		  [0x0000000100000a51,  0x0000000100000d47): DW_OP_lit23, DW_OP_stack_value)
		DW_AT_name        ("number")

However, this broke LLDB.

  (Int) number = <extracting data from value failed>

The value was read correctly, but apparently the value's type was different.
When reading a constu it was reading a uint64 (m_type = e_ulonglong) while for
the literal, it got a signed int (m_type = e_sint). This change makes sure we
read the value as an unsigned.

Differential revision: https://reviews.llvm.org/D51730

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@342142 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit a87f4cb64d8130367e2c07b524acb44ac87adbbe)

### DIFF
--- a/packages/Python/lldbsuite/test/lang/c/local_variables/Makefile
+++ b/packages/Python/lldbsuite/test/lang/c/local_variables/Makefile
@@ -1,0 +1,7 @@
+LEVEL = ../../../make
+
+C_SOURCES := main.c
+
+CFLAGS_EXTRAS += -O1
+
+include $(LEVEL)/Makefile.rules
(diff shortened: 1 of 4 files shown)


JDevlieghere:

PR #897, opened 2018-09-18
 Don't run Swift tests in the -gmodules configuration

adrian-prantl:
it only makes sense for Clang tests.

rdar://problem/42977904


adrian-prantl:
@swift-ci test and merge

dcci:
@adrian-prantl Are the changes in the xcodeproject intended?

jimingham:
Maybe somebody forgot to sort the xcode project?  Does git/svn have pre-commit hooks that we could get to do this.  I've forgotten a couple of times too and it makes for huge unnecessary diffs.

Jim

PR #899, opened 2018-09-18
[IRInterpreter] Fall back to JIT with 128-bit values.

dcci:
They're not that common, and falling back is definitely
better than throwing an error instead of the result. If we
feel motivated, we might end up implementing support for these,
but it's unclear whether it's worth the effort/complexity.

Fixes PR38925.

<rdar://problem/44436068>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@342262 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/packages/Python/lldbsuite/test/expression_command/rdar44436068/Makefile
+++ b/packages/Python/lldbsuite/test/expression_command/rdar44436068/Makefile
@@ -1,0 +1,3 @@
+LEVEL = ../../make
+C_SOURCES := main.c
+include $(LEVEL)/Makefile.rules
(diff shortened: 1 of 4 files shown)


dcci:
@swift-ci test and merge

jrose-apple:
This should be going to swift-5.0-branch rather than stable, no?

dcci:
I think both of them. We should set up auto mergers unless they're already up, though.

jrose-apple:
Yes, but the direction of the merger is swift-5.0-branch -> stable.

dcci:
For some reason I thought we had merges in both directions. Anyway, I'm going to cherry pick this to 5.0 as well :)

fredriss:
Stable will be merged in swift-5.0-branch the next time Swift re-branches from master. No need to put it there

PR #902, opened 2018-09-18
Factor out Swift frontend variables into Makefile.rules

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #903, opened 2018-09-18
[Swift] Remove all the hardcoded occurrences of mangling prefixes.

dcci:
<rdar://problem/44467875>

### DIFF
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -15,9 +15,10 @@
 #include "lldb/DataFormatters/StringPrinter.h"
 #include "lldb/Symbol/ClangASTContext.h"
 #include "lldb/Target/Process.h"
+#include "lldb/Target/SwiftLanguageRuntime.h"
 #include "lldb/Utility/DataBufferHeap.h"
 #include "lldb/Utility/Status.h"
-#include "lldb/Target/SwiftLanguageRuntime.h"
+#include "swift/Demangling/ManglingMacros.h"
 #include "llvm/ADT/Optional.h"
 #include "llvm/ADT/StringRef.h"
 
@@ -486,8 +487,7 @@ bool lldb_private::formatters::swift::NSContiguousString_SummaryProvider(
   if (guts_sp)
     return StringGuts_SummaryProvider(*guts_sp, stream, options);
 
-  static ConstString g_StringGutsType(
-      SwiftLanguageRuntime::GetCurrentMangledName("$Ss11_StringGutsVD"));
+  static ConstString g_StringGutsType(MANGLING_PREFIX_STR "s11_StringGutsVD");
   lldb::addr_t guts_location = valobj.GetValueAsUnsigned(LLDB_INVALID_ADDRESS);
   if (guts_location == LLDB_INVALID_ADDRESS)
     return false;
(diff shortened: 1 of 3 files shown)


dcci:
@swift-ci test and merge

dcci:
I wasn't entirely sure whether the correct spelling was `prefixes` or `prefices`, but I'm under the impression nobody really cares as long as it's understandable.

jimingham:
There's some contention about `prefices`.  The consensus is that prefices isn't the right pluralization of prefix, since "prefix" isn't actually a latin word.  Instead it is "derived from praefixum" and it's plural is praefices...  Since it isn't a direct latin word, using the latinate pluralization is considered wrong.  OTOH, the OED does cite `prefices` as a sometimes used alternate form of `prefixes`.

But the patch looks fine.  I wonder if in the last diff we're providing too much information.  We don't actually know that the only check IsSwiftMangledName does (or will do in the future) is the prefix, so by saying the prefix is the problem, we're speaking past what we know.  Maybe it would be better for that one to just say it isn't a valid swift name.

dcci:
I agree the message could be improved. In general, this is fundamentally something people shouldn't care about (unless they're developing lldb proper).

The error is returned by `GetTypeFromMangledTypename()` which is an internal function of lldb. It should be the caller responsibility to filter out invalid names, even though being defensive here doesn't hurt.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3862,9 +3863,10 @@ SwiftASTContext::GetTypeFromMangledTypename(const char *mangled_typename,
 
   if (!mangled_typename ||
       !SwiftLanguageRuntime::IsSwiftMangledName(mangled_typename)) {
-    error.SetErrorStringWithFormat("typename '%s' is not a valid Swift mangled "
-                                   "typename, it should begin with $S",
-                                   mangled_typename);
+    error.SetErrorStringWithFormat(
+        "typename '%s' is not a valid Swift mangled "
+        "typename, it should begin with " MANGLING_PREFIX_STR,
+        mangled_typename);
```
adrian-prantl, line 3869: This error message is way too specific. It should just say ... is not a valid Swift mangled name. No consumer will read this error and think "Oh, I need to change the prefix!" :-)

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -486,8 +487,7 @@ bool lldb_private::formatters::swift::NSContiguousString_SummaryProvider(
   if (guts_sp)
     return StringGuts_SummaryProvider(*guts_sp, stream, options);
 
-  static ConstString g_StringGutsType(
-      SwiftLanguageRuntime::GetCurrentMangledName("$Ss11_StringGutsVD"));
+  static ConstString g_StringGutsType(MANGLING_PREFIX_STR "s11_StringGutsVD");
```
adrian-prantl, line 490: This looks dangerous. This should be code that tree-matches the demangle tree instead.

dcci: I mean, sure, but it's worse than any other formatter in-tree? I don't think it's a problem specific to this very instance.

source/Symbol/SwiftASTContext.cpp
dcci: Agreed, see previous comment :)

PR #904, opened 2018-09-18
[Swift] Link against the swift reflection library.

zisko:
Link against swift reflection, special cased for linux.


dcci:
This is fine to me, @adrian-prantl ?

zisko:
@swift-ci test

source/Target/CMakeLists.txt
```diff
@@ -1,3 +1,19 @@
+include_directories(../Plugins/Process/Utility)
+
+set(REFLECTION_LIB "swiftReflection")
+
+if(LLVM_HOST_TRIPLE MATCHES "^((\\?!linux).)*$")
```
adrian-prantl: I think it would be better to use MATCHES ".*-apple-.*" here, because Swift also builds on Windows and other platforms.

adrian-prantl (changes requested):

source/Target/CMakeLists.txt
zisko: ok! I'll change that now!

zisko:
Failed on linux:
15:37:30 /usr/bin/ld: cannot find -lswiftReflection

this didn't happen locally... going to circle back

source/Target/CMakeLists.txt
zisko: changed to if(CMAKE_SYSTEM_NAME MATCHES "Darwin")

zisko:
@swift-ci test

source/Target/CMakeLists.txt
```diff
@@ -1,3 +1,23 @@
...
+if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
+    # libswiftReflection is placed in a subdirectory which depends
+    # on the host we're currently running on.
+    # FIXME: There must be a better way to get at libswiftReflection.a
+    string(REPLACE "-" ";" TRIPLE_LIST "${LLVM_HOST_TRIPLE}")
+    list(GET TRIPLE_LIST 0 ARCH)
+    list(GET TRIPLE_LIST 2 HOST_OS_VERSION)
+    string(REGEX MATCHALL "[a-z]+" HOST_OS "${HOST_OS_VERSION}")
+    set(REFLECTION_LIB "${REFLECTION_LIB}-${HOST_OS}-${ARCH}")
+    message("Building for non-linux host, setting reflection lib to ${REFLECTION_LIB}")
```
adrian-prantl, line 18: We can probably remove that message, too. Either we print it unconditionally, or, just not at all.

adrian-prantl (approved):

source/Target/CMakeLists.txt
zisko: Ah ok, I'll see how testing goes and change that...

zisko:
squashing and merging

adrian-prantl:
I think this just broke the build:

https://ci.swift.org/job/oss-lldb-incremental-osx-cmake/1739/
https://ci.swift.org/job/oss-lldb-incremental-osx-cmake/1739/consoleFull#11681430863122a513-f36a-4c87-8ed7-cbc36a1ec144

adrian-prantl:
@zisko ^

dcci:
@zisko this broke the bots. Can you please revert the commit?

dcci:
@zisko I ended up reverting this in #915 to unblock the bots. You can see the error in the logs, please let me know if you need anything else to reproduce this issue etc (FWIW, I think a `swiftci please test` should do it, but in case in doesn't, let me know and we can take a look together).

PR #911, opened 2018-09-19
Rename various ObjC-visible classes

mikeash:
Swift is renaming some classes that are visible to Objective-C to avoid conflicts between older versions and ABI-stable versions. Change those names here as well.

This goes with this Swift PR: apple/swift#19295

rdar://problem/35768222


dcci (approved):
We discussed this already via e-mail, but, just in case, explicitly approving here.

mikeash:
https://github.com/apple/swift/pull/19295

@swift-ci please test

mikeash:
apple/swift#19295

@swift-ci please test

mikeash:
apple/swift#19295

@swift-ci please test

PR #912, opened 2018-09-19
Emit a precise error message when lldb can't load a swift AST

vedantk:
Emit the same diagnostics the swift compiler would emit, and make sure
to actually unique these diagnostics (previously they would be emitted
each time lldb tried to create a context).

(cherry picked from commit 2bde84223ec637fb27575ef48f7207eb3d95585f)


vedantk:
@swift-ci test and merge

vedantk:
This is the wrong branch.

PR #917, opened 2018-09-20
Revert "Emit a precise error message when lldb can't load a swift AST"

dcci:
This reverts commit 2bde84223ec637fb27575ef48f7207eb3d95585f
because it broke the bots.


dcci:
@swift-ci test and merge

PR #921, opened 2018-09-20
Retry: Emit a precise error message when lldb can't load a swift AST

vedantk:
Emit the same diagnostics the swift compiler would emit, and make sure
to actually unique these diagnostics (previously they would be emitted
each time lldb tried to create a context).

Absent a good way to attach to & attempt to evaluate expressions in a
program compiled with an old version of swiftc, check this in without a
test.


vedantk:
@swift-ci test and merge

PR #923, opened 2018-09-20
[SwiftLanguageRuntime] Don't throw away a valid type already resolved.

dcci:
The logic here was backwards, so we ended up not displaying the value
correctly. This is another step towards a saner and more reliable
dynamic type resolution in the debugger. More to come!

<rdar://problem/36256619>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/protocols/class_protocol/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/protocols/class_protocol/Makefile
@@ -1,5 +1,3 @@
 LEVEL = ../../../../make
-
 SWIFT_SOURCES := main.swift
-
 include $(LEVEL)/Makefile.rules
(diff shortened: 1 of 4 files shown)


dcci:
@swift-ci test and merge

dcci:
In the example here:

```
  let myWhy : Int
   init(input : T)
   {
     myWhy = input.why  // Break here and print input
     myWhy = input.why //%self.expect('expr -d run -- input', substrs=['a.ShouldBeWhy', 'isa = a.ShouldBeWhy', 'before_why = 4277009102', 'why = 10'])
                       //%self.expect('expr -d run -- input', substrs=['a.ShouldBeWhy', 'isa = a.ShouldBeWhy', 'before_why = 4277009102', 'why = 10'])
   }
 }
```

The result of `p input` was a crash in expression evaluation and `frame var input` returned 
`(T) 0x0fffffffff`

dcci:
We still don't get class constrained protocol quite right, but that needs compiler modifications I'm going to tackle next.

jimingham (approved):
Can you add a comment before the "if (!instance_type)" explaining why the failure to get instance_type does NOT mean that the contents of class_type_or_name is not useful?  Viewed naively, returning true after a failure seems like the wrong behavior.  Other than that this is fine.

dcci:
@jimingham addressed your comments.

dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

jimingham:
Excellent, thanks!

PR #924, opened 2018-09-20
[lldbtest] Fix self.filecheck check file lookup

vedantk:
The 'test_filename' property in TestBase changes over time, so
attempting to find a check file relative to the directory containing
'test_filename' is flaky.

Use the absolute path of the check file as that's always correct (and
simpler). This relies on the test driver changing into the test
directory, which it seems we can safely assume.

As a drive-by, make self.filecheck respect the trace (-t) option.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@342699 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 3eeca50886b4bca5ffbaf14d6f282bc0c0ef339d)


vedantk:
@swift-ci test and merge

PR #925, opened 2018-09-21
Fix a use-after-free in ~SwiftASTContext(). 

adrian-prantl:
swift::ASTContext holds references to bits of the CompilerInvocation,
the SourceMgr, and the DiagnosticsEngine.  By ensuring they are
declared as members before the ASTContext, they get deallocated
*after* the ASTContext.

N3797, 12.4.8.

rdar://problem/44582316


adrian-prantl:
@swift-ci test and merge

PR #927, opened 2018-09-21
Update Swift Set/Dictionary formatters in preparation of upcoming changes

lorentey:
https://github.com/apple/swift/pull/19213 will change the internal layout of Set/Dictionary. Prepare for this by adding lldb support for the new storage layouts.

Bonus change: fix the new synthetic children frontend so that enumerating all children takes linear time. (It was quadratic with the old formatters.)

### DIFF
--- a/source/Plugins/Language/Swift/SwiftDictionary.cpp
+++ b/source/Plugins/Language/Swift/SwiftDictionary.cpp
@@ -44,19 +44,22 @@ DictionaryConfig::DictionaryConfig()
   // Note: We need have use the old _Tt names here because those are
   // still used to name classes in the ObjC runtime.
 
+  m_nativeStorageRoot_demangled =
+    ConstString("Swift._RawDictionaryStorage");
+
   // Native storage class
   m_nativeStorage_mangledRegex_ObjC =
-    ConstString("^_TtGCs37_HashableTypedNativeDictionaryStorage.*");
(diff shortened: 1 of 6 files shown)


lorentey:
@swift-ci test

lorentey:
This PR keeps support for old layouts. This can be removed in a followup PR later.

source/Plugins/Language/Swift/SwiftDictionary.cpp
```diff
@@ -65,6 +68,18 @@ DictionaryConfig::DictionaryConfig()
     = ConstString("Swift._SwiftDeferredNSDictionary<");
   m_deferredBridgedStorage_demangledRegex
     = ConstString("^Swift\\._SwiftDeferredNSDictionary<.+,.+>$");
+
+  // Legacy Swift 4.2 storage class names
+  m_legacy_nativeStorage_mangledRegex_ObjC =
```
dcci, line 73: Feel free to remove the old layout. `swiftc` and `swift-lldb` are revlocked so keeping around legacy is pointless for now.

dcci (approved):

source/Plugins/Language/Swift/SwiftDictionary.cpp
lorentey: Cool, I will do that in a followup next week. (I kept support for the old stuff in case we need to temporarily take out the Set/Dictionary PR.)

PR #929, opened 2018-09-21
Remove pointless divergences from llvm.org

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #932, opened 2018-09-24
[Swift] Add a test for generalized accessors basic functionality.

dcci:
<rdar://problem/43578347>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/generalized_accessors/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/generalized_accessors/Makefile
@@ -1,0 +1,3 @@
+LEVEL = ../../../make
+SWIFT_SOURCES := main.swift
+include $(LEVEL)/Makefile.rules
--- a/packages/Python/lldbsuite/test/lang/swift/generalized_accessors/TestGeneralizedAccessors.py
+++ b/packages/Python/lldbsuite/test/lang/swift/generalized_accessors/TestGeneralizedAccessors.py
@@ -1,0 +1,70 @@
+import lldb
+from lldbsuite.test.lldbtest import *
+import lldbsuite.test.decorators as decorators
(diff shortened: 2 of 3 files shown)


dcci:
@swift-ci test and merge

jrose-apple:
Sorry, how does this test generalized accessors?

dcci:
@jrose-apple They're tested through Dictionary subscript _modify. I might consider adding something more direct as a follow-up.

jrose-apple:
But you don't stop inside the accessor or anything. This is just "run some code, then break, then run some more code". It doesn't matter that the code before or after is a generalized accessor.

dcci:
cc: @airspeedswift (who actually crafted the test case for us in the debugger and explained to me how this works).

My understanding is that accessors modify how `double()` behave.
Before:
`subscript get` returns a copy of the array, the function modifies it, and `subscript set` replaces the old value with the new one (triggering CoW).

After:
`subscript _modify` moves the array into the optional, and so the array should be unique and will be modified without triggering CoW. Then after the yield, it’s moved back.

So, this is just testing that lldb behaves correctly during the yield. Am I missing something here?

jrose-apple:
The "during the yield" part is the part I'm suspicious of. You stopped on a line, not at a particular expression, so the access won't have started yet.

dcci:
I see your point. What about something modeled after what's in the test suite?

```
private var backingX = 0
var x: Int {
  _read {
    yield backingX
    backingX += 1
  }
  _modify {
    var temp = backingX
    yield &temp
    if temp >= backingX {
      backingX = temp
    }   
  }
}

print(x)
x += 10
print (x)
x -= 20
```

At that point I can stop on the `yield` line or anywhere in modify and make sure the value printed is correct.

jrose-apple:
That seems reasonable to me. You should also be able to print *other* parts of the struct besides the backing.

PR #937, opened 2018-09-26
[Swift] Make `import Dispatch` a lit test.

dcci:
So that we don't need to write a bunch a boilerplate, and a Makefile,
and do a fragile dance to search lldb (which is failing on the bots).


dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

PR #939, opened 2018-09-26
[Swift] Remove a test that wasn't testing what was supposed to.

dcci:
After a discussion with Jordan, I realized we should test something else.
This is not valuable, so, let's throw it away.

<rdar://problem/43578347>


dcci:
@swift-ci test and merge

PR #941, opened 2018-09-26
Re-enable TestSwiftDedupMacros.py

vedantk:
This test may sporadically segfault. It's likely that the underlying
memory corruption issue has been addressed (see the fix in
rdar://problem/44582316), but due to the difficulty of reproducing the
crash, we are not sure. If a crash is observed, try to collect a
crashlog before disabling this test.

rdar://44201206


vedantk:
CC @adrian-prantl

vedantk:
@swift-ci test and merge

adrian-prantl (approved):

PR #943, opened 2018-09-27
XFAIL failing test while investigating

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #956, opened 2018-10-03
[Swift] Stop emitting bogus warnings when we run a process again.

dcci:
We should only emit an error, if there's a fatal one.

<rdar://problem/44766464>

### DIFF
--- a/lit/Swift/Inputs/ContextError.swift
+++ b/lit/Swift/Inputs/ContextError.swift
@@ -1,0 +1,12 @@
+func use<T>(_ t : T) {}
+
+func string_tuple<T, U>(_ t : (T, U)) {
+  let (_, y) = t // here
+  use(y)
+}
+
+
+let s = "patatino"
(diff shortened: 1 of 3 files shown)


dcci:
@swift-ci test and merge

lit/Swift/Inputs/ContextError.swift
```diff
@@ -0,0 +1,12 @@
...
+func string_tuple<T, U>(_ t : (T, U)) {
+  let (_, y) = t // here
+  use(y)
+}
+
+
+let s = "patatino"
+string_tuple((s, s))
+
+// CHECK-NOT: warning
```
adrian-prantl, line 12: Can you add one positive check, too? Anything that run prints or perhaps the output of `fr v s` would be fine.

dcci: Probably, good idea.

PR #964, opened 2018-10-09
[Swift] Support evaluating of `$0`/`$1` etc.. inside a closure.

dcci:
<rdar://problem/20719448>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/TestClosureShortcuts.py
+++ b/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/TestClosureShortcuts.py
@@ -1,0 +1,16 @@
+# TestAccelerateSIMD.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+import lldbsuite.test.lldbinline as lldbinline
+from lldbsuite.test.decorators import *
+
+lldbinline.MakeInlineTest(__file__, globals(),
+                          decorators=[swiftTest,skipUnlessDarwin])
--- a/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/TestClosureShortcuts.py
+++ b/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/TestClosureShortcuts.py
@@ -1,4 +1,4 @@
-# TestAccelerateSIMD.py
+# TestClosureShortcuts.py
 #
 # This source file is part of the Swift.org open source project
 #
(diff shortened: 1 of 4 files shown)


source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -1184,13 +1184,8 @@ static swift::ASTContext *SetupASTContext(
...
     swift_ast_context->GetLanguageOptions().EnableObjCInterop = false;
 
-  if (repl || playground) {
-    swift_ast_context->GetLanguageOptions().Playground = true;
-    swift_ast_context->GetIRGenOptions().Playground = true;
-  } else {
-    swift_ast_context->GetLanguageOptions().Playground = true;
-    swift_ast_context->GetIRGenOptions().Playground = false;
-  }
+  swift_ast_context->GetLanguageOptions().Playground = repl || playground;
```
dcci, line 1187: This was a typo that prevented the feature to work correctly.

dcci:
Compiler bits here: 
https://github.com/apple/swift/pull/19808

packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/TestClosureShortcuts.py
```diff
@@ -0,0 +1,16 @@
+# TestAccelerateSIMD.py
```
jimingham: That's not the right name...

jimingham (commented):
With this patch in place, can I make $0 variables in random contexts.  What happens if I do:

(lldb) expression let $0 = 5

not in a closure context?  

And what kind of error do you get if you do:

(lldb) expression $0

in a non-closure context.  Currently this says:

error: anonymous closure argument not contained in a closure

it would be nice if we didn't lose that error and just let people make these variables.  It would be even worse if having made one  accidentally it collided with a real $0 variable...

If the behaviors in those cases are reasonable, then this seems great.  

If the answer is you can make $0 variables and they cause trouble but why would you ever do that...  Then I guess that's okay, it's more important to get this working.  But I'm a little sad.

packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/TestClosureShortcuts.py
dcci: Thanks!

dcci:
Thanks for your comments, Jim.

To reply to your questions:

```
(lldb) expression $0
error: use of undeclared identifier '$0'
```

(which seems reasonable to me). what do you think? I'm afraid there's not very much we can do here because `lldb` and `swift` have a fundamental namespace clash here. Variable names starting with `$` I think are really unlikely that maybe it's worth giving this one a shot. I'm not particularly happy with this solution either, but considering the number of reports we get about this being broken, it's probably OK to penalize other cases to get this one working.

dcci:
https://github.com/apple/swift/pull/19808

@swift-ci Please test

jimingham:
We also checked by hand that if you do:

(lldb) expr let $0 = 100

which will succeed with this patch, and then stop in a frame that has a real $0 in it, we find the closure automatic variable $0 and not the one we defined in the expression. It might be a good idea to add a test for that so we make sure the lookup rules don't change and make that fail.

Other than that, this looks good to me.

dcci:
https://github.com/apple/swift/pull/19965

@swift-ci Please test

PR #965, opened 2018-10-10
Cherry pick echo comment commands and test multiline expr

weliveindetail:
Cherry-picking these two commits to `stable` after they landed in the upstream. https://reviews.llvm.org/D52788 (Add EchoCommentCommands)
https://reviews.llvm.org/D52270 (TestMultilineExpr).


weliveindetail:
@swift-ci test and merge

PR #967, opened 2018-10-12
V for frame var

jimingham:
This adds "var" and "v" as short-hand aliases for "frame var" and "vo" as a short-hand alias for "frame var -O".


jimingham:

PR #969, opened 2018-10-12
Change the default handling for SIGPIPE to pass/,no-stop/no-notify.

jimingham:
Most of the time SIGPIPE is just annoying, and so we should
pass it on silently it by default.

<rdar://problem/39359145>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@344418 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #971, opened 2018-10-14
Improve libswiftReflection Linking on Linux

Kaiede:
Using CMAKE_SYSTEM_PROCESSOR is a good first step, but for platforms where it maps to something different (say, armv7l to armv7), then you need more complete logic. Ideally, we would want Swift itself to tell us what to link against, but this change is still an improvement, although it is blatantly reusing similar logic from Swift's cmake scripts.


dcci:
This is definitely better.

dcci:
@swift-ci test and merge

dcci (approved):

PR #972, opened 2018-10-14
Ensure Consistent Typing in GetWordCount()

Kaiede:
Inside SwiftHashedContainer::GetWordCount, it uses std::max with potentially confliciting types. The literal constant should be coerced into the correct type that will work on both 32-bit and 64-bit.


dcci:
I don't see any problem with this, but Karoy wrote this formatter, so he should sign-off.

dcci:
@swift-ci please test

lorentey (approved):
👍 Thanks for noticing this!

PR #973, opened 2018-10-15
[ExpressionParser] Perform archetype binding in the materializer.

dcci:
(if needed). We immediately after IRGen the type and ask its size.
IRGen needs a fully realized type, otherwise it asserts.

Fixes (part of) SR-8938.

<rdar://problem/45216847>


dcci:
@swift-ci test and merge

PR #975, opened 2018-10-15
[SwiftLanguageRuntime] Remove unneded and potentially wrong code.

dcci:
We shouldn't change the address of the class and replace it with
the ISA pointer. RemoteAST, if anything, should do that for us,
but we don't need it anyway.


dcci:
@jimingham we discussed about this code at some point, but until today, I never got around to audit it and make sure it was really not needed :)

dcci:
@swift-ci test and merge

PR #984, opened 2018-10-18
Swift location

compnerd:
Enable building lldb in-tree with LLVM and swift in standalone mode.  Adding the
LLDB_PATH_TO_SWIFT_SOURCE allows us to alter the module cache path
appropriately.


compnerd:
CC: @llvm-beanz @dcci

dcci (approved):

dcci:
standalone mode is not really officially supported, so I don't have strong opinions on it. If this patch fixes your issues, I guess it's fine.

compnerd:
I think that the unified mode really does make sense and will be needed anyways.  I'll focus my work there instead.  #989 works towards that.

PR #988, opened 2018-10-20
[Swift] Fix several bugs in existential dynamic type resolution.

dcci:
lldb was performing their own dynamic type resolution and payload
projection. At lesat the following cases were broken (and fixed
with my patch), in a generic context:

-> Error as AnyObject (class-backed existentials)
-> enums conforming to the Error protocol (and, probably, Error
   existentials in general)
-> Tuples of Errors.

The patch switches this codepath to RemoteAST, removing a bit
of code, and fixing several corner cases of existential resolution.
There are maybe other cases that are still broken, but this is
a decent enough improvement that I figure I could just go ahead
and commit.

Fixes:
<rdar://problem/41228727>
<rdar://problem/37479968>
<rdar://problem/41229005>

Adding several testcases, while here, to make sure this doesn't
regress in future.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/generic_error_tuple/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/generic_error_tuple/main.swift
@@ -1,0 +1,29 @@
+class MyErr : Error {
+  var x : Int
+
+  init(_ x : Int) {
+    self.x = x
+  }
+}
+
+class MyOtherErr : MyErr {}
+
+func h<U, V>(_ tuple : (U, V)) -> (U, V) {
+  return tuple //%self.expect("frame var -d run-target -- tuple", 
+               //%             substrs=['(a.MyErr, a.MyErr) tuple',
+               //%                      '0 =', '(x = 23)',
+               //%                      'a.MyErr = {', 'x = 42'])
+               //%self.expect("expr -d run-target -- tuple",
+               //%             substrs=['(a.MyErr, a.MyErr) $R',
+               //%                      '0 =', '(x = 23)',
+               //%                      'a.MyErr = {', 'x = 42'])
+}
+
+func main() -> Int {
+  let foo : MyErr = MyErr(23)
+  let goo : MyErr = MyOtherErr(42)
+  h((foo, goo))
+  return 0
+}
+
+let _ = main()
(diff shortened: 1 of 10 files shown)


dcci:
apple/swift#19965

@swift-ci Please test

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2001,57 +2001,28 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Promise(
...
-        address.SetLoadAddress(load_addr, &m_process->GetTarget());
-        return true;
-      }
-    }
+    lldb::addr_t existential_address = in_value.GetPointerValue();
+    if (!existential_address || existential_address == LLDB_INVALID_ADDRESS)
+      return false;
+    auto &target = m_process->GetTarget();
+    assert(IsScratchContextLocked(target) &&
+           "Swift scratch context not locked ahead");
```
dcci, line 2011: @adrian-prantl Adrian, I'll appreciate if you can take a look and let me know whether the scratchASTContext acquisition is correct (I think so, but another pair of eyes is always welcome).

dcci:
apple/swift#19965

@swift-ci Please test

packages/Python/lldbsuite/test/lang/swift/generic_error_tuple/main.swift
```diff
@@ -0,0 +1,29 @@
...
+
+  init(_ x : Int) {
+    self.x = x
+  }
+}
+
+class MyOtherErr : MyErr {}
+
+func h<U, V>(_ tuple : (U, V)) -> (U, V) {
+  return tuple //%self.expect("frame var -d run-target -- tuple",
```
dcci, line 12: I just noticed that, for some reason, RemoteAST thinks the dynamic type of this is the base class `MyErr` and not the derived class `MyOtherErr`. It's not a bug in lldb, I think, but I'm going to open a rdar to investigate.

dcci:
apple/swift#19965

@swift-ci Please test

dcci:
apple/swift#19965

@swift-ci Please test

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: LGTM.

PR #989, opened 2018-10-21
build: add dependency on generated target

compnerd:
This repairs the build for the unified layout.  The lldb target depends
on generated headers.  When building in non-standalone mode, ensure that
we have a dependency present to ensure the ordering.

### DIFF
--- a/source/Commands/CMakeLists.txt
+++ b/source/Commands/CMakeLists.txt
@@ -47,3 +47,6 @@ add_lldb_library(lldbCommands
   LINK_COMPONENTS
     Support
   )
+if(NOT LLDB_BUILT_STANDALONE)
+  add_dependencies(lldbCommands swift-syntax-generated-headers)
+endif()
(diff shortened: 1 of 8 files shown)


compnerd:
CC: @adrian-prantl @dcci

dcci:
This should go against `stable`, not `upstream-with-swift`.

compnerd:
@dcci, sure, happy to adjust that.  I'm just unfamiliar with the branching model that swift-lldb uses.  Is it `stable` -> `upstream-with-swift` instead of `upstream-with-swift` -> `stable`?

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

dcci:
As this adds `swift` specific logic to cmake files which are also upstream, might I ask you to add markers? If we decide to move this code upstream, it would be useful to highlight swift-specific parts. Otherwise, LGTM.

compnerd:
@dcci - sure, is there a specific marker that lldb is using for the cmake changes?

compnerd:
Interestingly enough the dependency issue seems to have gone away?  I'm gonna discard this for now.

PR #996, opened 2018-10-23
[ValueObject] Stop assuming types are non-zero sized.

dcci:
Some backends might violate this assumption, e.g. swift.

<rdar://problem/40962410>


dcci:
@swift-ci test and merge

PR #997, opened 2018-10-23
Drop support for non-standard Swift resource directories.

adrian-prantl:
Attempting to import anything from  a Swift resource directory that is
not the one used by LLDB's  embedded Swift compiler is doomed to fail;
especially so if the binary being debugged was compiled on  a different
machine.

Note that the resource dir does not encode the target platform, so
this should not affect cross-debugging.

Part of rdar://problem/45282507

[Apologies for the large patch, I had to reorder static functions.]


adrian-prantl:
@swift-ci test

adrian-prantl:
@swift-ci test

PR #998, opened 2018-10-24
[DO NOT MERGE][Swift] Update String formatters for the UTF-8 storage layout

lorentey:
We're experimenting with switching `String`'s internal representation to UTF-8. This PR updates lldb's data formatters to support the new string representation.

This PR should not be merged until/unless the new representation lands on Swift's master branch.

### DIFF
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -35,95 +35,13 @@ using namespace llvm;
 
 bool lldb_private::formatters::swift::Character_SummaryProvider(
     ValueObject &valobj, Stream &stream, const TypeSummaryOptions &options) {
-  static ConstString g_Small("Small");
-  static ConstString g_Large("Large");
+  static ConstString g__str("_str");
 
-  static ConstString g__representation("_representation");
-
-  static ConstString g__storage("_storage");
-  static ConstString g_storage("storage");
-  static ConstString g_Some("Some");
-
-  ProcessSP process_sp(valobj.GetProcessSP());
-  if (!process_sp)
+  ValueObjectSP str_sp = valobj.GetChildMemberWithName(g__str, true);
+  if (!str_sp)
     return false;
 
-  ValueObjectSP representation_sp =
-      valobj.GetChildMemberWithName(g__representation, true);
-
-  if (!representation_sp)
-    return false;
-
-  ConstString value(representation_sp->GetValueAsCString());
-
-  if (value == g_Large) {
-    ValueObjectSP largeBuffer_sp(representation_sp->GetChildAtNamePath(
-        {g_Large, g__storage, g_storage, g_Some}));
-    if (!largeBuffer_sp)
-      return false;
-
-    lldb::addr_t buffer_ptr =
-        largeBuffer_sp->GetValueAsUnsigned(LLDB_INVALID_ADDRESS);
-    if (LLDB_INVALID_ADDRESS == buffer_ptr || 0 == buffer_ptr)
-      return false;
-
-    buffer_ptr += 2 * process_sp->GetAddressByteSize();
-    Status error;
-    buffer_ptr = process_sp->ReadPointerFromMemory(buffer_ptr, error);
-    if (LLDB_INVALID_ADDRESS == buffer_ptr || 0 == buffer_ptr)
-      return false;
-
-    StringPrinter::ReadStringAndDumpToStreamOptions options;
-    options.SetLocation(buffer_ptr)
-        .SetEscapeNonPrintables(true)
-        .SetNeedsZeroTermination(true)
-        .SetPrefixToken(0)
-        .SetProcessSP(valobj.GetProcessSP())
-        .SetQuote('"')
-        .SetStream(&stream)
-        .SetLanguage(lldb::eLanguageTypeSwift);
-
-    return StringPrinter::ReadStringAndDumpToStream<
-        StringPrinter::StringElementType::UTF16>(options);
-  } else if (value == g_Small) {
-    const uint64_t invalidRepr = 0xFFFFFFFFFFFFFFFF;
-
-    ValueObjectSP smallBuffer_sp(
-        representation_sp->GetChildAtNamePath({g_Small}));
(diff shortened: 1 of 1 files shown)


milseman:
@dcci can you review this change for use in conjunction with UTF-8 backed String?

dcci:
@milseman @lorentey sorry this slipped through the cracks. I'm going to take a closer look at it today.

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -180,67 +98,109 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
   } else {
-    object = valobj.GetChildAtNamePath({g_object});
-    otherBits = valobj.GetChildAtNamePath({g_otherBits, g_value});
+    return false; // Unsupported arch?
   }
 
-  if (!object || !otherBits)
-    return false;
-
-  // Taken from StringObject.swift:
```
dcci, line 205: I found out this comment describing the layout very useful. Can we have something equivalent in the new world order?

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -256,149 +216,68 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
-      }
-
-      uint64_t nsstringAddr =
-          decodedVal->GetValueAsUnsigned(LLDB_INVALID_ADDRESS);
-      return readStringAsNSString(nsstringAddr);
-    }
+  } else if ((discriminator & 0xF8) == 0x18) { // 00011xxx: Exotic
+    // Not currently generated
+    return false;
+  } else { // Invalid discriminator
```
dcci, line 278: `else` after `return` is not really needed.

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -256,149 +216,68 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
-                                     payloadAddr, isUTF16)) {
-        uint64_t count = otherBits->GetValueAsUnsigned(0);
-        return readStringFromAddress(payloadAddr, count, isUTF16);
-      }
-
-      uint64_t nsstringAddr =
-          decodedVal->GetValueAsUnsigned(LLDB_INVALID_ADDRESS);
-      return readStringAsNSString(nsstringAddr);
-    }
+  } else if ((discriminator & 0xF8) == 0x18) { // 00011xxx: Exotic
```
dcci, line 275: Can you elaborate what "exotic" means here?

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -256,149 +216,68 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
-        process->GetTarget().GetScratchClangASTContext()->GetBasicType(
-            lldb::eBasicTypeObjCID);
+  uint8_t discriminator = raw1 >> 56;
 
-    bool success = false;
-    ValueObjectSP nsstring;
+  if ((discriminator & 0xB0) == 0xA0) { // 1x10xxxx: Small string
+    uint64_t count = (raw1 >> 56) & 0x0F;
+    uint64_t maxCount = (ptrSize == 8 ? 15 : 10);
+    if (count > maxCount) return false;
```
dcci, line 228: can you please `clang-format` (this is true for the patch in general).

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -180,67 +98,109 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
+      raw0 = count | (pointerBits << 32);
+      raw1 = flags | (discriminator << 56);
+    } else {
+      raw0 = count | (flags << 48);
+      raw1 = pointerBits | (discriminator << 56);
+    }
   } else {
-    object = valobj.GetChildAtNamePath({g_object});
-    otherBits = valobj.GetChildAtNamePath({g_otherBits, g_value});
+    return false; // Unsupported arch?
```
dcci, line 200: This makes me a little nervous. Can we make is an `lldbassert` ?

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -180,67 +98,109 @@ bool lldb_private::formatters::swift::StringGuts_SummaryProvider(
...
+    } else if (variantCase.startswith("native")) {
+      static ConstString g_native("native");
+      payload_sp = variant_sp->GetChildMemberWithName(g_native, true);
+    } else if (variantCase.startswith("bridged")) {
+      static ConstString g_bridged("bridged");
+      auto anyobject_sp = variant_sp->GetChildMemberWithName(g_bridged, true);
+      if (!anyobject_sp) return false;
+      payload_sp = anyobject_sp->GetChildAtIndex(0, true); // "instance"
+    } else {
+      return false; // Unknown variant
```
dcci, line 185: Can we make this an `lldbassert` to make sure we catch new variants added in the future?

dcci (commented):
I like the fact the new format (and hence the formatter(s)) are simple. Some comments inline.

source/Plugins/Language/Swift/SwiftFormatters.cpp
milseman: Right, the new diagram is at https://github.com/milseman/swift/blob/17f7eca94b4b57bd25ef7f7dc4f6dbf32321fc5b/stdlib/public/core/StringObject.swift#L27

When we finalize, I can either copy-paste it in here, or else link to the formal ABI documentation. We'll have a series of diagrams describing the layout of each component, so it might just make sense to link to that in the end.

source/Plugins/Language/Swift/SwiftFormatters.cpp
milseman: Would this look like?:
```c++
    } else {
      lldbassert("Uknown variant")
      return false;
    }
```

milseman:
This has been subsumed by https://github.com/apple/swift-lldb/pull/1022

PR #1003, opened 2018-10-25
Warning Cleanups

compnerd:
This is intended to cleanup most of the warnings from the LLDB build.  The ones in particular that are left are related to C11 and four character codes.

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -46,7 +46,7 @@ class TypeInfo;
 }
 namespace serialization {
 struct ValidationInfo;
-struct ExtendedValidationInfo;
+class ExtendedValidationInfo;
 }
 }
 
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -34,8 +34,8 @@ class MemoryReader;
 class RemoteAddress;
 }
 
-template <typename T> class External;
-template <unsigned PointerSize> class RuntimeTarget;
+template <typename T> struct External;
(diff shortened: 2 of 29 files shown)


compnerd:
CC: @dcci @adrian-prantl @fredriss

jimingham (changes requested):
It seems like this patch has a bunch of changes against files that exist on llvm.org, and in code that is not swift specific - so should apply there.  Any change that touches code shared between llvm.org and github should be made on llvm.org, and then the llvm.org -> github/upstream-with-swift automerger will pick it up.

Can you either ensure that the code in shared files is in code only present on github (there shouldn't be much of that anymore) or push it to llvm.org first and then merge it over.  We really want to minimize the divergence between upstream-with-swift on github and llvm.org.

compnerd:
Sure, I can upstream the changes to the files that exist there already and let it come through overnight through the automerger.

dcci (approved):
Yes, please upstream the changes first, then back port here and update the remaining files.
Other than that, it looks good.

compnerd:
@jimingham, @dcci - I just had a look, the change that can be upstreamed is close to zero: there is one switch statement that can be upstreamed (if I change the code around it since the enumeration is different there).  The other changes are all local only.  They are due to the local changes that have been made to lldb and the extensions due to swift.

dcci:
OK, well, I guess you're almost fine then. Upstream that little bit and then apply all the changes here.
I think we should keep the divergence as little as possible.

dcci:
For the future, I think this is not the flow that we use.
We generally merge to stable, not `upstream-with-swift.` Also, we first do the llvm.org upstream commit and then cherry pick here. You may do the other way around now, but something to keep in mind for the future.

PR #1004, opened 2018-10-25
Support nwere versions of the Segger J-Link jtag board software.

jasonmolenda:
Support newer versions of the Segger J-Link jtag board software.
Add support in ProcessGDBRemote::GetGDBServerRegisterInfo
for recognizing a generic "arm" architecture that will be used if
nothing better is available so that we don't ignore the register
definitions if we didn't already have an architecture set.
Also in ProcessGDBRemote::DoConnectRemote don't set the target
arch unless we have a valid architecture to set it to.

Platform::ConnectProcess will try to get the current target's
architecture, or the default architecture, when creating the
target for the connection to be attempted.  If lldb was started
with a target binary, we want to create this target with that
architecture in case the remote gdb stub doesn't supply a
qHostInfo arch.

Add logging to Target::MergeArchitecture.

<rdar://problem/34916465>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@345106 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit e0009a48e5f467b39dc3d9acc2f1ba9a78878dec)

### DIFF
--- a/packages/Python/lldbsuite/test/functionalities/gdb_remote_client/TestArmRegisterDefinition.py
+++ b/packages/Python/lldbsuite/test/functionalities/gdb_remote_client/TestArmRegisterDefinition.py
@@ -1,0 +1,130 @@
+from __future__ import print_function
+import lldb
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test.decorators import *
+from gdbclientutils import *
+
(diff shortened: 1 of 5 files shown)


jasonmolenda:
@swift-ci test and merge

PR #1008, opened 2018-10-29
Replace boolean parameter with enum value according r342633 (Fix building with clang-cl on Windows)

gmittert:
Building on Windows with Clang-cl fails without this patch from upstream. I filed this bug: https://bugs.swift.org/browse/SR-9107, but this might be an easier solution.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@342671 91177308-0d34-0410-b5e6-96231b3b80d8


dcci (approved):

dcci:
@swift-ci test and merge

PR #1012, opened 2018-10-30
Target: compute the swiftReflection target more generically

compnerd:
We did not compute the swiftReflection target sufficiently generically.  This
manifested as a problem when cross-compiling lldb for Windows.  We would try to
link against swiftReflection rather than swiftReflection-windows-x86_64.  Yank
the architecture mapping function from the corelibs SwiftSupport and use that to
generically map for non-Darwin targets.


compnerd:
@swift-ci please test

compnerd:
CC: @gottesmm @dcci @adrian-prantl

gottesmm:
This makes the code a lot better/etc. Gets a +1 from me on the cmake.

dcci (approved):

PR #1015, opened 2018-10-30
Utility: fix cross-compilation from Linux to Windows

compnerd:
Only attempt to link against Backtrace if it is found.  Without this,
trying to cross-compile to Windows would try to link against
"Backtrace_LIBRARY-NOTFOUND.lib".

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@345569 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 53d62c8a5f480ec61cc372c7c0275502ef6123b6)


compnerd:
CC: @dcci @adrian-prantl

compnerd:
@swift-ci please test

gottesmm:
Can lldb build without this? What is the downstream effect of this? Are you sure this is safe?

compnerd:
@gottesmm - no, it breaks when cross-compiling.  This is totally safe, its a backport from upstream :-)

gottesmm:
I explained offline to compnerd that my real question here was since he was not touching users if the users of backtrace were properly guarded. He said it was... so assuming that is true, no concerns from me.

compnerd:
Right, those are already guarded correctly.  This is a fix for a linker option snafu.

dcci (approved):

PR #1017, opened 2018-10-30
Target: use the secondary host build of swiftReflection

compnerd:
Adjust for the new host build of the swiftReflection library for the tools.

### DIFF
--- a/source/Target/CMakeLists.txt
+++ b/source/Target/CMakeLists.txt
@@ -1,52 +1,5 @@
 include_directories(../Plugins/Process/Utility)
 
-set(REFLECTION_LIB "swiftReflection")
-
-# Duplicated from SwiftSupport.cmake in
-# swift-corelibs-{libdispatch,Foundation,XCTest}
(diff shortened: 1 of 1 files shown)


compnerd:
https://github.com/apple/swift/pull/20156

@swift-ci please test

gottesmm:
This cmake looks good to me. @dcci your thoughts?

dcci:
If this doesn't break anything, I'm very happy to see it going in :)

dcci (approved):

dcci:
Please keep an eye on the bots, though ... linking swiftReflection has been source of issues in the past (sometimes not caught by PR testing).

gottesmm:
@dcci It shouldn't make a difference. What we are doing here is compiling swiftReflection for host separately from the target one. So there may be a broken build due to automergers/etc, if lldb merges before swift downstream, but once that resolves, this should be much better.

dcci:
I agree. Let me merge this one :)

PR #1021, opened 2018-11-01
[Swift] Defer initialization of variables checks to the compiler.

dcci:
<rdar://problem/45588366>


dcci:
@swift-ci test and merge

jimingham:
There was a little code between type-checking and this second hand-verification pass, so theoretically this extra check could serve the purpose of verifying that the intervening AST manipulations didn't introduce an error.  But those manipulations don't do very much, so it's hard to see this as being necessary.  Anyway, if we end up needing to do another verifier pass at this point, that's something we should depute to the compiler rather than trying to roll it by hand.

So this seems good to me.

dcci:
thank you @jimingham !

PR #1026, opened 2018-11-02
Add ad-hoc codesigning to all of the swift test makefiles that build

jasonmolenda:
Add ad-hoc codesigning to all of the swift test makefiles that build
binaries/dylibs manually.  Change the -static-stdlib option in
Makefile.rules so it is only added when linking; don't include it
when compiling a .swift file.  ad-hoc codesigning is needed for some
ios etc builds.


jasonmolenda:
@swift-ci test and merge

jasonmolenda:
@swift-ci test and merge

jasonmolenda:
@swift-ci test and merge

PR #1039, opened 2018-11-06
Adjust for SVN r346157

compnerd:
Go plugin has been removed.  Update the code for that.


compnerd:

dcci:
sure.

dcci:
Note that we have an auto merger that will pick up Jonas commit. So this is not really needed.

JDevlieghere:
Thanks! I only resolved the merge conflicts, I should've double checked if there were references left.

PR #1032, opened 2018-11-05
Cherry-pick upstream patches

adrian-prantl:
rdar://problem/21814005

PR #1040, opened 2018-11-06
Adjust for SVN r346165

compnerd:


compnerd:
CC: @dcci @adrian-prantl 

@adrian-prantl - I'm not sure if there is an execution context here that needs to be plumbed through.

adrian-prantl:
Thanks for doing the work, but can we abandon this in favor of https://github.com/apple/swift-lldb/pull/1032 ? That PR does thread through the exe_ctx.

adrian-prantl:
The automerger should bubble it up to upstream-with-swift eventually.

adrian-prantl (changes requested):
https://github.com/apple/swift-lldb/pull/1032 does this, too.

compnerd:
Sure, I'm fine with #1032 instead.

PR #1045, opened 2018-11-06
Fix deadlock rdar://45600292

fredriss:
This was already worked around in rdar://problem/38461035, but it was lost in
a follow-up merge resolution. This reinstates the workaround and adds a test
(from https://bugs.swift.org/browse/SR-7114) to make sure we don't regress this.

### DIFF
--- a/lit/SwiftREPL/Deadlock.test
+++ b/lit/SwiftREPL/Deadlock.test
@@ -1,0 +1,15 @@
+// RUN: %lldb --repl < %s | FileCheck %s
+
+// From https://bugs.swift.org/browse/SR-7114
+// This sequence was deadlocking.
+
+let a = 9007199254740991.0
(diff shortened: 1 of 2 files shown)


fredriss:
@swift-ci test and merge

lit/SwiftREPL/Deadlock.test
```diff
@@ -0,0 +1,15 @@
+// RUN: %lldb --repl < %s | FileCheck %s
+
+// From https://bugs.swift.org/browse/SR-7114
+// This sequence was deadlocking.
+
+let a = 9007199254740991.0
+// CHECK: a: Double = 9007199254740991
```
dcci, line 7: You might want to use `{{a}}` instead of `a`, otherwise this test will always pass.

PR #1050, opened 2018-11-07
Fix the lldb-gtest-build target in the xcode project file so

jasonmolenda:
Fix the lldb-gtest-build target in the xcode project file so
we can build  & run the unit tests again.


jasonmolenda:
@swift-ci test and merge

PR #1055, opened 2018-11-08
[Swift][DataFormatters] Add formatters for GLKit types.

fredriss:
GLKit types are completely opaque, but reusing the support we have for
Accelearate types, it was easy to add a summary formatters for them.

rdar://44685031

### DIFF
--- a/lit/SwiftREPL/GLKIT.test
+++ b/lit/SwiftREPL/GLKIT.test
@@ -1,0 +1,46 @@
+// Test formatters for Accelerate/simd.
+// REQUIRES: darwin
+
+// RUN: %lldb --repl < %s | FileCheck %s
+
+import GLKit
+
+GLKVector3(v: (1, 2, 3))
+// CHECK: $R{{.*}}: GLKVector3 = (1.000000e+00, 2.000000e+00, 3.000000e+00)
+
+GLKVector4(v: (1, 2, 3, 4))
+// CHECK: $R{{.*}}: GLKVector4 = (1.000000e+00, 2.000000e+00, 3.000000e+00, 4.000000e+00)
+
+GLKVector2(v: (4, 5))
+// CHECK: $R{{.*}}: GLKVector2 = (4.000000e+00, 5.000000e+00)
+
+GLKQuaternion(q: (4, 3, 2, 1))
+// CHECK: $R{{.*}}: GLKQuaternion = (4.000000e+00, 3.000000e+00, 2.000000e+00, 1.000000e+00)
+
+GLKMatrix3Identity
+// CHECK: $R{{.*}}: GLKMatrix3 =
+// CHECK: [ [1.000000e+00, 0.000000e+00, 0.000000e+00],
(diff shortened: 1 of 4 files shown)


fredriss:
@swift-ci test

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -1115,27 +1149,60 @@ bool lldb_private::formatters::swift::AccelerateSIMD_SummaryProvider(
...
+    return false;
+
+  SIMDElementFormatter formatter(SIMDElementKind::Float32);
+
+  unsigned num_elements =
+      is_quaternion ? 4 : llvm::hexDigitValue(type_name.back());
+  DataExtractor data;
+  uint64_t len = valobj.GetData(data, error);
+  const uint8_t *buffer = data.GetDataStart();
+  if (!is_matrix) {
```
dcci, line 1188: `is_matrix` & `!is_matrix` are complementary condition, so I guess you can just do.
```
if (!is_matrix)
  [....]
straight_line_code
```

(in fact, the last `return false` I think is never reached).

dcci (approved):
Pretty straightforward, just one minor.

lit/SwiftREPL/GLKIT.test
```diff
@@ -0,0 +1,46 @@
...
+
+GLKMatrix4Identity
+// CHECK: $R{{.*}}: GLKMatrix4 =
+// CHECK: [ [1.000000e+00, 0.000000e+00, 0.000000e+00, 0.000000e+00],
+// CHECK:   [0.000000e+00, 1.000000e+00, 0.000000e+00, 0.000000e+00],
+// CHECK:   [0.000000e+00, 0.000000e+00, 1.000000e+00, 0.000000e+00],
+// CHECK:   [0.000000e+00, 0.000000e+00, 0.000000e+00, 1.000000e+00] ]
+
+let mat = GLKMatrix3Make(10, 20, 30, 40, 50, 60, 70, 80, 90)
+// CHECK: mat: GLKMatrix3 =
```
dcci: Also, this should probably be `{{mat}}`

source/Plugins/Language/Swift/SwiftFormatters.cpp
fredriss: yep. Too much copy/pasting

fredriss:
@swift-ci test

vedantk (approved):

jimingham:
ReadVector and PrintMatrix should be static, shouldn't they?  I know they weren't before, but...

vedantk:
They are, it's all in an anonymous namespace.

jimingham:
Ah, okay.  The llvm coding standards rule for anonymous namespaces is:

make anonymous namespaces as small as possible, and only use them for class declarations

precisely because it makes it hard to see the visibility of large swaths of code contained in the namespace.  

Maybe we should declare these functions in the anonymous namespace and then have the implementations outside?

PR #1058, opened 2018-11-08
[MemoryReader] Only dump a reasonable number of bytes.

dcci:
Otherwise the log gets too verbose, not really readable.
Discussed with (and proposed by) Adrian.

<rdar://problem/45592295>

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1162,7 +1162,7 @@ SwiftLanguageRuntime::GetMemoryReader() {
         return false;
       }
 
-      if (log) {
+      if (log && log->GetVerbose()) {
         StreamString stream;
         for (uint64_t i = 0; i < size; i++) {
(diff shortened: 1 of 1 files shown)


dcci:
@swift-ci test and merge

jimingham:
Can you change this so you check whether the log is Verbose, and emit the full memory if it is verbose?

jimingham:
You can just call GetVerbose on the Log object you get back.

adrian-prantl:
Could we hide the entire MemoryReader logging behind the verbose flag? Or am I biased just because *I* don't know how to make sense of the dumped bytes?  :-)

dcci:
I think it's useful to get errors from the memory reader because it means you're reading something you shouldn't have, so I'm less happy about making the whole thing guarded by verbose. What Jim suggests is fine, I think.

dcci:
@jimingham updated the review, please take a look.

dcci:

jimingham (approved):
That's fine.

PR #1059, opened 2018-11-08
Replace uses of $CURRENT_ARCH in the xcode project file with

jasonmolenda:
Replace uses of $CURRENT_ARCH in the xcode project file with
"x86_64" - newer versions of Xcode do not define this.


jasonmolenda:
@swift-ci test and merge

PR #1060, opened 2018-11-08
Don't use $CURRENT_ARCH in the xcode project file; hardcode

jasonmolenda:
Don't use $CURRENT_ARCH in the xcode project file; hardcode
x86_64.


jasonmolenda:
@swift-ci test and merge

PR #1064, opened 2018-11-09
Hardcode x86_64 in all the places where the

jasonmolenda:
Hardcode x86_64 in all the places where the
now-obsolete CURRENT_ARCH setting was used.


jasonmolenda:
@swift-ci test and merge

PR #1065, opened 2018-11-09
Revert "Refactor SwiftASTContext::GetResourceDir() to make it testable"

dcci:
This reverts commit 331fafaf164cadfce2d388d2e44c8f529cc9b374.


dcci:
This commit seems to cause an infinite loop in the lldb testsuite. Reverting to unblock swift.

adrian-prantl:
Is it because m_once_flag is uninitialized?

adrian-prantl:
Do you have any instructions for reproducing?

dcci:
`ninja check-lldb` on a Linux machine. We reproduced on real hw, but I don't see why this shouldn't repro on a docker instance.

PR #1070, opened 2018-11-09
Work with a gdb-remote target that doesn't handle the

jasonmolenda:
Work with a gdb-remote target that doesn't handle the
qWatchpointSupportInfo packet correctly.  

In GDBRemoteCommunicationClient::GetWatchpointSupportInfo,
if the response to qWatchpointSupportInfo does not
include the 'num' field, then we did not get an answer
we understood, mark this target as not supporting that
packet.

In Target.cpp, rename the very confusingly named
CheckIfWatchpointsExhausted to CheckIfWatchpointsSupported,
and check the error status returned by 
Process::GetWatchpointSupportInfo.  If we cannot determine
what the number of supported watchpoints are, assume that
they will work.  We'll handle the failure
later when we try to create/enable the watchpoint if the
Z2 packet isn't supported.

Add a gdb_remote_client test case.

<rdar://problem/42621432> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@346561 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 256337a57d0015bc8179b85e9946a7af3d539190)

### DIFF
--- a/packages/Python/lldbsuite/test/functionalities/gdb_remote_client/TestNoWatchpointSupportInfo.py
+++ b/packages/Python/lldbsuite/test/functionalities/gdb_remote_client/TestNoWatchpointSupportInfo.py
@@ -1,0 +1,64 @@
+from __future__ import print_function
+import lldb
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test.decorators import *
+from gdbclientutils import *
+
+class TestNoWatchpointSupportInfo(GDBRemoteTestBase):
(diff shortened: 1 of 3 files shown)


jasonmolenda:
@swift-ci test and merge

jasonmolenda:
@swift-ci test and merge

jasonmolenda:
@swift-ci test and merge

PR #1071, opened 2018-11-10
[Swift] Ask the runtime about the padding for tuples.

dcci:
<rdar://problem/45791023>

### DIFF
--- a/source/Plugins/Language/Swift/SwiftHashedContainer.cpp
+++ b/source/Plugins/Language/Swift/SwiftHashedContainer.cpp
@@ -24,6 +24,9 @@
 #include "Plugins/Language/ObjC/NSDictionary.h"
 
 #include "swift/AST/ASTContext.h"
+#include "swift/AST/Types.h"
+#include "swift/Remote/RemoteAddress.h"
+#include "swift/RemoteAST/RemoteAST.h"
 #include "llvm/ADT/StringRef.h"
 
 #include <algorithm>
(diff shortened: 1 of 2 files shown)


dcci:
Don't merge, it's still a WIP because I need to write a test.

dcci:
@swift-ci test

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -587,8 +590,33 @@ NativeHashedStorageHandler::GetElementAtIndex(size_t idx) {
...
+
+  uint64_t offset = m_key_stride_padded;
+  if (llvm::isa<::swift::TupleType>(swift_type)) {
+    auto &remote_ast = runtime->GetRemoteASTContext(*scratch_ctx);
+
+    ::swift::remote::RemoteAddress optmeta(nullptr);
+    ::swift::remoteAST::Result<uint64_t> result =
+        remote_ast.getOffsetOfMember(swift_type, optmeta, "1");
+    if (result)
+      offset = result.getValue();
```
lorentey: I think this code should be moved to the constructor so that `m_key_stride_padded` gets initialized correctly. (Also, `m_value_offset` would be a better name for it.) The tuple type is constructed there anyway, so it’s easier to inspect.

The `full_buffer_sp` initialization above should just use the size of the element type, so perhaps we should explicitly cache that too.

dcci: Back to this after the break. I'm going to send an update soon.

dcci:
-> Moved to the constructor as suggested by Karoy.
-> Added test.

dcci:
@swift-ci please test

dcci:
@swift-ci please test macOS

dcci:
@swift-ci please test macOS

dcci:
@lorentey can you please take a look?

dcci:
@swift-ci please test macOS

lorentey (approved):
👍LGTM!

dcci:
Thanks for the review, @lorentey !

PR #1074, opened 2018-11-12
Replace std::string with StringRef

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #1077, opened 2018-11-13
Since ABI's now hold a process WP, they should be handed

jimingham:
out one per process rather than keeping a single global instance.

Differential Revision: https://reviews.llvm.org/D54460

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@346775 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:
@swift-ci please test

PR #1078, opened 2018-11-13
 Fix a bug in the parsing of the LC_BUILD_VERSION Mach-O load command.

adrian-prantl:
LC_BUILD_VERSION records are of variable length. The original code
would use uninitialized memory when the size of a record was exactly 24.

rdar://problem/46032185

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@346812 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 2e9aff681eee5e9ab5155e635b5c1bdc0cdd38cf)


adrian-prantl:
@swift-ci test

PR #1085, opened 2018-11-15
Add setting to require hardware breakpoints.

JDevlieghere:
When debugging read-only memory we cannot use software breakpoint. We
already have support for hardware breakpoints and users can specify them
with `-H`. However, there's no option to force LLDB to use hardware
breakpoints internally, for example while stepping.

This patch adds a setting target.require-hardware-breakpoint that forces
LLDB to always use hardware breakpoints. Because hardware breakpoints
are a limited resource and can fail to resolve, this patch also extends
error handling in thread plans, where breakpoints are used for stepping.

Differential revision: https://reviews.llvm.org/D54221

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@346920 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 92ece34c9136b93a9e000fc53df58c3b8090de7b)

### DIFF
--- a/include/lldb/API/SBBreakpoint.h
+++ b/include/lldb/API/SBBreakpoint.h
@@ -129,6 +129,8 @@ public:
   static uint32_t
   GetNumBreakpointLocationsFromEvent(const lldb::SBEvent &event_sp);
 
+  bool IsHardware() const;
+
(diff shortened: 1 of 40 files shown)


JDevlieghere:

PR #1086, opened 2018-11-15
Fix a use-after-free of the ABI plugin.

jimingham:
This was introduced in r346775.  Previously the ABI shared_ptr
was declared as a function local static meaning it would live
forever.  After the change, someone has to create a strong
reference to it or it will go away.  In this code, we were
calling ABI::FindPlugin(...).get(), so it was being immediately
destroyed and we were holding onto a dangling pointer.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@346932 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/source/Symbol/Variable.cpp
+++ b/source/Symbol/Variable.cpp
@@ -156,14 +156,14 @@ void Variable::Dump(Stream *s, bool show_context) const {
                                 .GetBaseAddress()
                                 .GetFileAddress();
     }
-    ABI *abi = nullptr;
+    ABISP abi;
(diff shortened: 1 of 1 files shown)


jimingham:
I checked and this was the only code that did this bad thing.

jimingham:
@swift-ci please test

jasonmolenda (approved):
Yeah, looks good, I can't find any other instances of this bug either.

jimingham:
@swift-ci please test

PR #1087, opened 2018-11-16
Remove PlaygroundQuickLook test.

fredriss:
Historically, this test was verifying some reflection support that was
used for Playgrounds support. Today, through API changes, this was just
testing that we could create an print a specific enum.

With the switch to Swift 5.0 (https://github.com/apple/swift/pull/20374),
this enum is now deprecated and the test fails.

As the coverage provided here seems pretty insignificant, let's just remove
the test.


fredriss:
@swift-ci test and merge

PR #1094, opened 2018-11-18
Fix a lifetime issue in FoundationValueTypes.cpp

fredriss:
m_synth_up was keeping a reference to the locally created ObjCRuntimeSyntheticProvider.
Fix this by making the lifetime of ObjCRuntimeSyntheticProvider the same as the
lifetime of the frontend.

rdar://problem/43799990

### DIFF
--- a/source/Plugins/Language/Swift/FoundationValueTypes.cpp
+++ b/source/Plugins/Language/Swift/FoundationValueTypes.cpp
@@ -13,6 +13,8 @@
 #include "FoundationValueTypes.h"
 #include "ObjCRuntimeSyntheticProvider.h"
 
+#include "llvm/ADT/STLExtras.h"
+
 #include "lldb/Core/ValueObject.h"
 #include "lldb/DataFormatters/FormattersHelpers.h"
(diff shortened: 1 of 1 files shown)


fredriss:
@swift-ci test and merge

source/Plugins/Language/Swift/FoundationValueTypes.cpp
```diff
@@ -282,7 +284,8 @@ class URLComponentsSyntheticChildrenFrontEnd
     : public SyntheticChildrenFrontEnd {
 public:
   URLComponentsSyntheticChildrenFrontEnd(lldb::ValueObjectSP valobj_sp)
-      : SyntheticChildrenFrontEnd(*valobj_sp), m_synth_up(nullptr)
+      : SyntheticChildrenFrontEnd(*valobj_sp), m_synth_frontend_up(),
+        m_synth_backend_up()
```
adrian-prantl, line 288: Not really part of this patch, but: I suppose these should be named `_sp` (smartpointer) instead of `_up`?

PR #1099, opened 2018-11-26
[CMake] Use NO_CMAKE_FIND_ROOT_PATH when finding clang, llvm, and swift

bulbazord:
When you cross-compile or specify a custom sysroot, it is possible that
the CMake toolchain files will set CMAKE_FIND_ROOT_PATH and set
CMAKE_FIND_ROOT_PATH_MODE_* VARIABLES to ONLY, so packages will only be
searched in a rerooted setting. This is an issue when we search for
clang, llvm, and swift, because we are usually passing them explicitly with
LLDB_PATH_TO_{CLANG,LLVM,SWIFT}_BUILD and do not want these paths to be
rerooted.


bulbazord:
Also if I should submit this to another branch as well, let me know and I can create another PR.

compnerd:
LGTM

compnerd:
CC: @dcci @adrian-prantl

dcci:
@xiaobai this should go to `stable` (the automerger will auto propagate to `upstream-with-swift`).
I'm reluctant to accept this patch without CI testing, so my kind request is that you open a PR against stable.

bulbazord:
@dcci: Will do. :)

PR #1100, opened 2018-11-27
[CMake] Use NO_CMAKE_FIND_ROOT_PATH when finding clang, llvm, and swift

bulbazord:
When you cross-compile or specify a custom sysroot, it is possible that
the CMake toolchain files will set CMAKE_FIND_ROOT_PATH and set
CMAKE_FIND_ROOT_PATH_MODE_* VARIABLES to ONLY, so packages will only be
searched in a rerooted setting. This is an issue when we search for
clang, llvm, and swift, because we are usually passing them explicitly with
LLDB_PATH_TO_{CLANG,LLVM,SWIFT}_BUILD and do not want these paths to be
rerooted.


bulbazord:
cc @dcci @compnerd @adrian-prantl

dcci:
@swift-ci please test

bulbazord:
@dcci: Is this good to go?

dcci (approved):
LGTM.

dcci:
feel free to merge this.

PR #1102, opened 2018-11-27
 Fix the layout calculation of generic tuple types.

adrian-prantl:
Performing dynamic type resolution on a tuple type can actually change
the tuple's layout, so when computing the offsets of a tuple's
elements, we need to use the static (but archetype-bound) type instead
of the dynamic type.
    
This patch requires a matching commit in Swift compiler.
    
rdar://problem/45462765

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -6905,6 +6905,25 @@ CompilerType SwiftASTContext::GetChildCompilerTypeAtIndex(
   } break;
 
   case swift::TypeKind::Tuple: {
+    // Dynamic type resolution may actually change(!) the layout of a tuple, so
+    // we need to get the offset from the static (but archetype-bound) version.
+    auto static_value = valobj->GetStaticValue();
+    auto static_type = static_value->GetCompilerType();
+    auto static_swift_type = GetCanonicalSwiftType(static_type);
+    if (swift::isa<swift::TupleType>(static_swift_type))
+      swift_can_type = static_swift_type;
+    if (swift_can_type->hasTypeParameter()) {
+      if (!exe_ctx)
+        return {};
+      auto *exe_scope = exe_ctx->GetBestExecutionContextScope();
+      auto *frame = exe_scope->CalculateStackFrame().get();
+      auto *runtime = exe_scope->CalculateProcess()->GetSwiftLanguageRuntime();
+      if (!frame || !runtime)
+        return {};
+      auto bound = runtime->DoArchetypeBindingForType(*frame, static_type);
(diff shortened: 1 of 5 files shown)


adrian-prantl:
@swift-ci test with https://github.com/apple/swift/pull/20814

adrian-prantl:
@swift-ci test

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1474,15 +1479,14 @@ SwiftLanguageRuntime::GetMemberVariableOffset(CompilerType instance_type,
     // Bind generic parameters if necessary.
     if (instance && swift_type->hasTypeParameter())
       if (auto *frame = instance->GetExecutionContextRef().GetFrameSP().get())
-        if (auto bound = DoArchetypeBindingForType(*frame, instance_type)) {
+        if (auto bound = DoArchetypeBindingForType(
+                *frame, instance->GetStaticValue()->GetCompilerType())) {
```
dcci: Can we consider adding a comment here explaining why we do the archetype binding on the static type?

dcci (approved):
This looks awesome, I'm glad we got rid of another hack :)

adrian-prantl:
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/20814
@swift-ci test

source/Symbol/SwiftASTContext.cpp
```diff
@@ -6905,6 +6905,25 @@ CompilerType SwiftASTContext::GetChildCompilerTypeAtIndex(
   } break;
 
   case swift::TypeKind::Tuple: {
+    // Dynamic type resolution may actually change(!) the layout of a Tuple, so
+    // we need to get the offset from the static (but archetype-bound) version.
+    auto static_value = valobj->GetStaticValue();
+    auto static_type = static_value->GetCompilerType();
+    auto static_swift_type = GetCanonicalSwiftType(static_type);
+    if (static_swift_type->getKind() == swift::TypeKind::Tuple)
```
slavapestov: Instead can you write `if (swift::isa<TupleType>(static_swift_type))`?

adrian-prantl: Sure.. why is that preferred, because it's shorter?

adrian-prantl: To answer my own question: I guess it would also work for subclasses of TupleType.

adrian-prantl:
test with https://github.com/apple/swift/pull/20814
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/20814
@swift-ci test

PR #1103, opened 2018-11-28
 Fix the layout calculation of generic tuple types.

adrian-prantl:
Performing dynamic type resolution on a tuple type can actually change
the tuple's layout, so when computing the offsets of a tuple's
elements, we need to use the static (but archetype-bound) type instead
of the dynamic type.

This patch requires a matching commit in the Swift compiler.

rdar://problem/45462765


adrian-prantl:
test with https://github.com/apple/swift/pull/20815
@swift-ci test

PR #1104, opened 2018-11-28
[SwiftASTContext] Lift PerformAutoImport() so that it can be used els…

dcci:
…ewhere.

<rdar://problem/30398933>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -8051,3 +8051,189 @@ void lldb_private::printASTValidationInfo(
   for (StringRef ExtraOpt : ext_ast_info.getExtraClangImporterOptions())
     LLDB_LOG(log, "  -- {0}", ExtraOpt);
 }
+
+static void DescribeFileUnit(Stream &s, swift::FileUnit *file_unit) {
+  s.PutCString("kind = ");
+
+  switch (file_unit->getKind()) {
+  default: { s.PutCString("<unknown>"); }
+  case swift::FileUnitKind::Source: {
+    s.PutCString("Source, ");
+    if (swift::SourceFile *source_file =
+            llvm::dyn_cast<swift::SourceFile>(file_unit)) {
+      s.Printf("filename = '%s', ", source_file->getFilename().str().c_str());
+      s.PutCString("source file kind = ");
+      switch (source_file->Kind) {
+      case swift::SourceFileKind::Library:
+        s.PutCString("Library");
+      case swift::SourceFileKind::Main:
+        s.PutCString("Main");
+      case swift::SourceFileKind::REPL:
+        s.PutCString("REPL");
+      case swift::SourceFileKind::SIL:
+        s.PutCString("SIL");
+      }
+    }
+  } break;
+  case swift::FileUnitKind::Builtin: {
+    s.PutCString("Builtin");
+  } break;
+  case swift::FileUnitKind::SerializedAST:
+  case swift::FileUnitKind::ClangModule: {
+    if (file_unit->getKind() == swift::FileUnitKind::SerializedAST)
+      s.PutCString("Serialized Swift AST, ");
+    else
+      s.PutCString("Clang module, ");
+    swift::LoadedFile *loaded_file = llvm::cast<swift::LoadedFile>(file_unit);
+    s.Printf("filename = '%s'", loaded_file->getFilename().str().c_str());
+  } break;
+  };
+}
+
+// Gets the full module name from the module passed in.
+
+static void GetNameFromModule(swift::ModuleDecl *module, std::string &result) {
+  result.clear();
+  if (module) {
+    const char *name = module->getName().get();
+    if (!name)
+      return;
+    result.append(name);
+    const clang::Module *clang_module = module->findUnderlyingClangModule();
+
+    // At present, there doesn't seem to be any way to get the full module path
+    // from the Swift side.
+    if (!clang_module)
+      return;
+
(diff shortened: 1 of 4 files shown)


dcci:
@swift-ci please test

dcci:
@adrian-prantl may I ask you to take a look?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -8032,3 +8032,173 @@ void lldb_private::printASTValidationInfo(
...
+        if (!load_one_module(module_name))
+          return false;
+      }
+    }
+  } else {
+    llvm::SmallVector<swift::ModuleDecl::ImportedModule, 2> parsed_imports;
+    source_file.getImportedModules(parsed_imports,
+                                   swift::ModuleDecl::ImportFilter::All);
+    auto *persistent_expression_state =
+        sc.target_sp->GetSwiftPersistentExpressionState(exe_scope);
```
adrian-prantl, line 8211: This is the only bit that seems to be specific to the expression evaluator.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -8032,3 +8032,173 @@ void lldb_private::printASTValidationInfo(
...
+                    "module: %s.\n",
+                    module_name.c_str());
+      if (!load_one_module(module_const_str))
+        return false;
+      // How do we tell we are in REPL or playground mode?
+      persistent_expression_state->AddHandLoadedModule(module_const_str);
+    }
+    // Finally get the hand-loaded modules from the
+    // SwiftPersistentExpressionState and load them into this context:
+    if (!persistent_expression_state->RunOverHandLoadedModules(load_one_module))
```
adrian-prantl, line 8234: Looking at the implementation of RunOverHandLoadedModules this is just a really fancy `for`-loop. I wonder where else we need `m_hand_loaded_modules` if at all.

adrian-prantl (commented):
This is generally fine, I just wonder if the bit with the persistent expression state should be left in the expression parser.

source/Symbol/SwiftASTContext.cpp
jimingham: PerformAutoImport is used in two different ways, depending on the value of user_imports.

When user_imports is false, then we just look in the current ExecutionContext's CU for modules that that CU imports, and hand import them.  That's because sometimes dependent modules get loaded lazily and we need to force them to be present.

When user_imports is true, then we peer into the swift::SourceFile that is the result of expression parsing and find any import statements that might be there, and import the modules references.  That is to support:

(lldb) expr import Foundation

The first use seems general, but the second use seems very expression parser specific.  I wonder if we oughtn't to split this into two functions, one of which does the general auto-importing and the other handles SourceFile import statements?

source/Symbol/SwiftASTContext.cpp
jimingham: We are currently using m_hand_loaded_modules in two ways.  

The first is in PerformAutoImport where we dig through the parsed AST and find all the import statements and then use RunOverHandLoadedModules in the same function to import them.  

We also use it to force loading the Swift module so that you can use the swift expression parser even if you haven't loaded Swift yet (e.g. in the REPL), for instance in ParseAndImport.  That was really what the ivar was for, so various different pieces of code could say "I need you to load this for me!" wherever you were in the expression parser and it would get done.  

OTOY, that only got used to force Swift to be loaded.  Since I don't think you're ever going to get very far in the swift expression parser w/o loading the Swift module, if it's cheap to "import" the Swift module when it's already present, maybe PerformAutoImport could just always do that?

dcci:
@swift-ci please test linux

dcci:
Looks like this codepath was covered after all :)

```
20:27:17 Error output:
20:27:17 error: in auto-import:
20:27:17 Failed to load linked library dl of module Foundation - errors:
20:27:17 Failed to find "libdl.so" in paths:
20:27:17 ,  /home/buildnode/jenkins/workspace/swift-lldb-PR-Linux/branch-stable/Ninja-ReleaseAssert/swift-linux-x86_64/lib/swift/linux
20:27:17 Failed to find "libdl.so" on RPATH, error: dlopen failed for unknown reasons.
20:27:17 
```

dcci:
@swift-ci please test

dcci:
@adrian-prantl updated to make it a static function as you suggested. May I ask you to take another look?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -8051,3 +8051,189 @@ void lldb_private::printASTValidationInfo(
...
+      s.PutCString("Serialized Swift AST, ");
+    else
+      s.PutCString("Clang module, ");
+    swift::LoadedFile *loaded_file = llvm::cast<swift::LoadedFile>(file_unit);
+    s.Printf("filename = '%s'", loaded_file->getFilename().str().c_str());
+  } break;
+  };
+}
+
+// Gets the full module name from the module passed in.
```
adrian-prantl, line 8093: `///` ?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -8051,3 +8051,189 @@ void lldb_private::printASTValidationInfo(
...
+
+static void GetNameFromModule(swift::ModuleDecl *module, std::string &result) {
```
adrian-prantl, line 8095: GetModuleName?

source/Symbol/SwiftASTContext.cpp
adrian-prantl: The bits with the persistent expression state might be better if they remained in the expression parser, since they won't be needed by anyone else.

PR #1105, opened 2018-11-28
[CMake] Add missing dep for unified builds

davezarzycki:
The target of the symlink needs to exist too.


davezarzycki:
@swift-ci please test platform

davezarzycki:
I don't think many people are doing unified builds, so I'm going to merge this. We can always revert it.

PR #1106, opened 2018-11-28
[ExpressionParser] Inline RunOverHandLoadedModules.

dcci:
It has only one use, and if we want to make LoadOneModule a
standalone function instead of a lambda this gets in the way.
The function was quite too general for what we needed to be,
so as an added bonus this makes the code slightly simpler.

<rdar://problem/30398933>

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftPersistentExpressionState.h
+++ b/source/Plugins/ExpressionParser/Swift/SwiftPersistentExpressionState.h
@@ -37,6 +37,8 @@ namespace lldb_private {
 /// 0-based counter for naming result variables.
 //----------------------------------------------------------------------
 class SwiftPersistentExpressionState : public PersistentExpressionState {
+  typedef std::set<lldb_private::ConstString> HandLoadedModuleSet;
+
 public:
   class SwiftDeclMap {
   public:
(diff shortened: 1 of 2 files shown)


dcci:

source/Plugins/ExpressionParser/Swift/SwiftPersistentExpressionState.h
```diff
@@ -111,15 +113,8 @@ class SwiftPersistentExpressionState : public PersistentExpressionState {
...
-
-  bool RunOverHandLoadedModules(HandLoadedModuleCallback callback) {
-    for (ConstString name : m_hand_loaded_modules) {
-      if (!callback(name))
-        return false;
-    }
-    return true;
-  }
+  /// This returns the list of hand-loaded modules.
+  HandLoadedModuleSet GetHandLoadedModules() { return m_hand_loaded_modules; }
```
adrian-prantl, line 117: Was RunOverHandLoadedModules an override function? If not, this LGTM.

dcci:
I haven't seen it marked as override.
This is what I see on my machine

```
dtdebugger:lldb davide$ grep -R 'RunOverHandLoadedModules' source/ include/
source//Plugins/ExpressionParser/Clang/ClangPersistentVariables.h:  bool RunOverHandLoadedModules(HandLoadedModuleCallback callback) {
```

dcci:
So, there's another `RunOverHandLoadedModules` in the clang expression parser, which is, also, unused. I'm going to remove that on llvm.org at some point.

jimingham (commented):
This seems fine to me.  The method wasn't overloaded, though in fact there's an identical method in ClangPersistentVariables.h.  That one - and the associated AddHandLoadedModule - don't seem to get used at all.  I bet that's just detritus from the big CompilerType refactoring we did when the Swift support was submitted to github.  Can we remove those?

jimingham:
That's actually not on llvm.org.  It was just some code we didn't remove from the source when we refactored the pre-CompilerSystem version of the Swift support into the CompilerSystem world.  I'll put up a PR to remove it here once I've ensured it still builds.

dcci:
I can handle this, while I'm in the area :)

jimingham:
'kay, thanks!

PR #1107, opened 2018-11-29
[SwiftASTContext] Hoist LoadOneModule.

dcci:
This way we can call it from other context(s).

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -29,6 +29,8 @@
 #include "llvm/ADT/Optional.h"
 #include "llvm/Support/Threading.h"
 
+#include "swift/AST/Module.h"
+
 #include <map>
 #include <set>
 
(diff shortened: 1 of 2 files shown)


dcci:

include/lldb/Symbol/SwiftASTContext.h
```diff
@@ -769,6 +771,14 @@ class SwiftASTContext : public TypeSystem {
 
   void SetCachedType(const ConstString &mangled, const lldb::TypeSP &type_sp);
 
+  static bool
```
jimingham, line 774: Is there any reason why these have to be static methods?  They have to have a swift_ast_context to work, and they are in SwiftASTContext.h.  Why not make them instance methods?

adrian-prantl: A static method has the advantage of explicitly declaring what it modifies and makes it a bit easier to read the code if you are unfamiliar with it. That said; if one argument is `this` and it's not const, then this argument doesn't hold up :-)

dcci: You are correct, I'll constify `SwiftASTContext` here.

jimingham: Not sure I understand this argument.  This function is loading a module into the AST context.   Conceptually this is not a const operation even if the compiler can't tell that.

PR #1108, opened 2018-11-29
Disable accelerate_simd test for now.

stephentyrone:


stephentyrone:
@swift-ci please test

vedantk (approved):

vedantk:
You might need to disable https://github.com/apple/swift-lldb/blob/stable/lit/SwiftREPL/SIMD.test as well.

PR #1109, opened 2018-11-29
[SwiftASTContext] Split PerformAutoImport in two functions.

dcci:
PerformAutoImport was, in fact, doing both auto import and
user import. The former is general and needed in several places
in the debugger. The latter is expression parser specific.
This splits the function in two, improving the API.

<rdar://problem/30398933>


dcci:

PR #1110, opened 2018-11-29
[ExpressionParser] Remove a legacy broken call that doesn't compile.

dcci:


dcci:
@swift-ci test and merge

jimingham:
Greg did all the completion work for the REPL so I have no state on why this unused branch was added.  It was there since this was checked into GitHub, and there's no sign the other branch was ever used.  So I'm fine with deleting it.

PR #1111, opened 2018-11-30
One other test to disable to land simd changes.

stephentyrone:
Also need to disable lit/SwiftREPL/SIMD.test for the short-term.


stephentyrone:
@swift-ci please test

dcci (approved):

PR #1112, opened 2018-11-30
Disable simd tests 5.0

stephentyrone:
Disable simd tests so we can land simd changes in Swift for 5.0


stephentyrone:
@swift-ci please test

stephentyrone:
@fredriss, this is a cherry-pick disabling two simd tests while we land the corresponding simd change in Swift. Vedant and Davide reviewed these for stable.

PR #1115, opened 2018-11-30
Thread-safety: Lock the global LLVMContext before constructing an IRGenModule

adrian-prantl:
The DWARFParser may call into multiple SwiftASTContexts at the same
time but only once per SwiftASTContext.

<rdar://problem/43877611>

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -801,6 +801,8 @@ protected:
 
   std::vector<lldb::DataBufferSP> &GetASTVectorForModule(const Module *module);
 
+  /// Data members.
+  /// @{
   std::unique_ptr<swift::CompilerInvocation> m_compiler_invocation_ap;
   std::unique_ptr<swift::SourceManager> m_source_manager_ap;
   std::unique_ptr<swift::DiagnosticEngine> m_diagnostic_engine_ap;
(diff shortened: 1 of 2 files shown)


adrian-prantl:
@swift-ci test

adrian-prantl:
Hmm, no. This crashes TestSwiftConditionalBreakpoint.py ; also there exists IRExecutionUnit::GetLLVMGlobalContextMutex().

Back to the drawing board.

adrian-prantl:
Alternative approach that locks the global context before creating IRGenModule.

adrian-prantl:
@swift-ci test

include/lldb/Symbol/SwiftASTContext.h
```diff
@@ -859,7 +859,7 @@ class SwiftASTContext : public TypeSystem {
 
   typedef ThreadSafeDenseMap<const char *, lldb::TypeSP> SwiftTypeMap;
   SwiftTypeMap m_swift_type_map;
-
+  /// @{
```
JDevlieghere: Should this be `@}`?

JDevlieghere (commented):
I was confused for a second with the title being outdated. Other than that this looks sensible.

adrian-prantl:
@swift-ci test

JDevlieghere:
Thanks :-)

PR #1116, opened 2018-12-03
Add time cast to SymbolFileDWARFDebugMap

JDevlieghere:
After TimePoint's precision was increased in LLVM we started seeing
failures because the modification times didn't match. This adds a time
cast to ensure that we're comparing TimePoints with the same level of
precision.

Based on r347660 on llvm.org


JDevlieghere:

dcci (approved):

PR #1117, opened 2018-12-03
[Target] AutoImport modules when getting the scratch ASTContext.

dcci:
This is easy, after a fair amount of yak shaving. The patch
    fixes the case where we need to print something from foundation,
    but we didn't import foundation already (i.e. this is the first `expr`).
    Don't imprort if the ASTContext was already poisoned, it would be
    a disaster.
    
    Finally fixes <rdar://problem/30398933>

### DIFF
--- a/source/Target/Target.cpp
+++ b/source/Target/Target.cpp
@@ -2494,6 +2494,15 @@ SwiftASTContextReader Target::GetScratchSwiftASTContext(
         llvm::cast_or_null<SwiftASTContext>(GetScratchTypeSystemForLanguage(
             &error, eLanguageTypeSwift, create_on_demand));
   }
+
+  StackFrameWP frame = exe_scope.CalculateStackFrame();
+  StackFrameSP frame_sp = StackFrameSP(frame.lock());
+  if (frame_sp && frame_sp.get() && swift_ast_ctx &&
+      !swift_ast_ctx->HasFatalErrors()) {
+    SymbolContext sc = exe_scope.CalculateStackFrame()->GetSymbolContext(true);
+    swift_ast_ctx->PerformAutoImport(*swift_ast_ctx, sc, frame, nullptr, error);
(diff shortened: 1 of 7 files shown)


dcci:
@swift-ci test and merge

source/Target/Target.cpp
```diff
@@ -2494,6 +2494,15 @@ SwiftASTContextReader Target::GetScratchSwiftASTContext(
         llvm::cast_or_null<SwiftASTContext>(GetScratchTypeSystemForLanguage(
             &error, eLanguageTypeSwift, create_on_demand));
   }
+
+  StackFrameWP frame = exe_scope.CalculateStackFrame();
+  StackFrameSP frame_sp = StackFrameSP(frame.lock());
```
adrian-prantl, line 2499: Does this work when you do:
```
lldb a.out
(lldb) p global_var
```

without running the program first?

dcci: Checked with Adrian this doesn't regress what's already there.

source/Target/Target.cpp
```diff
@@ -2494,6 +2494,15 @@ SwiftASTContextReader Target::GetScratchSwiftASTContext(
...
+  if (frame_sp && frame_sp.get() && swift_ast_ctx &&
+      !swift_ast_ctx->HasFatalErrors()) {
+    SymbolContext sc = exe_scope.CalculateStackFrame()->GetSymbolContext(true);
```
fredriss, line 2502: Is there a reason why this is not frame_sp->GetSymbolContext(true)?

dcci: Yes, me not knowing the shortcut. I'll fix this in follow up.

dcci: https://github.com/dcci/swift-lldb/pull/new/cleanme,

PR #1118, opened 2018-12-04
[Target] Avoid an unneeded function call. Pointed out by Fred.

dcci:


dcci:
@swift-ci test and merge

jimingham:
Did you add the code above that as well?  It looks a little weird.  ExecutionContextScope::CalculateStackFrame() returns a StackFrameSP.  It's odd to set that to a StackFrameWP and then lock it.  If the StackFrame isn't good any more, the SP will be empty, so this seems unnecessary.

dcci:
`PerformAutoImport` wants a `wp`. Is there a better way of expressing that?

jimingham:
Ah, I missed you were using it as a weak pointer as well as as a shared pointer in that code...

The only reason PerformAutoImport takes a weak pointer is that this call stack used to only be used by the SwiftExpressionParser which held onto the current stack frame as a weak pointer because it is a longer lived object (reasonable because it is a longer lived object.)  

But the ivar (m_stack_frame_wp) gets passed to ParseAndImport which locks the pointer (and fails if it doesn't resolve), then it passes it as a weak pointer AGAIN to PerformAutoImport which again locks it.  That seems weird and complicates other uses of PerformAutoImport.  We don't in general pass weak pointers in lldb API's and that's just awkward.

I'd change the ParseAndImport and PerformAutoImport take shared pointers, and have the SwiftExpressionParser lock m_stack_frame_wp before passing it to ParseAndImport.

jasonmolenda:
You can pass a StackFrameSP to something that takes a StackFrameWP, right?

PR #1120, opened 2018-12-05
[CMake] Fix side-effect from LLDB_VERSION change in r346668 for frame…

weliveindetail:
…work-enabled builds

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@348360 91177308-0d34-0410-b5e6-96231b3b80d8


weliveindetail:
Conflict resolutions can be push directly to upstream-with-swift.

PR #1135, opened 2018-12-12
[Swift] Rewrite IsSwiftThunk so that it doesn't need runtime instanti…

dcci:
…ation.

It's a property of a mangled name, so there's no need to do the
heavy lifting. Makes this call much much cheaper, as it showed
up in several profile runs.

### DIFF
--- a/source/API/SBFrame.cpp
+++ b/source/API/SBFrame.cpp
@@ -1411,23 +1411,19 @@ bool SBFrame::IsSwiftThunk() const {
   StackFrame *frame = nullptr;
   Target *target = exe_ctx.GetTargetPtr();
   Process *process = exe_ctx.GetProcessPtr();
-  if (target && process) {
-    Process::StopLocker stop_locker;
-    if (stop_locker.TryLock(&process->GetRunLock())) {
-      frame = exe_ctx.GetFramePtr();
-      if (frame) {
-        SwiftLanguageRuntime *runtime = process->GetSwiftLanguageRuntime();
-        if (!runtime)
-          return false;
-        SymbolContext sc;
-        sc = frame->GetSymbolContext(eSymbolContextSymbol);
-        if (!sc.symbol)
-          return false;
-        return runtime->IsSymbolARuntimeThunk(*sc.symbol);
-      }
-    }
-  }
-  return false;
+  if (!target || !process)
+    return false;
+  Process::StopLocker stop_locker;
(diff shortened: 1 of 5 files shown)


dcci:
@swift-ci test and merge

source/API/SBFrame.cpp
```diff
@@ -1411,23 +1411,19 @@ bool SBFrame::IsSwiftThunk() const {
   StackFrame *frame = nullptr;
```
JDevlieghere, line 1411: Can we move this down and to the initializing in the same statement?

source/API/SBFrame.cpp
```diff
@@ -1411,23 +1411,19 @@ bool SBFrame::IsSwiftThunk() const {
...
-  return false;
+  if (!target || !process)
+    return false;
+  Process::StopLocker stop_locker;
+  if (!stop_locker.TryLock(&process->GetRunLock()))
+    return false;
+  frame = exe_ctx.GetFramePtr();
+  if (!frame)
+    return false;
+  SymbolContext sc;
```
JDevlieghere, line 1422: Same here, any reason you've split this up?

JDevlieghere (approved):
One small nit, otherwise this LGTM

adrian-prantl:
Could we cherry-pick that to swift-5.0-branch?

source/Target/LanguageRuntime.cpp
```diff
@@ -344,24 +344,8 @@ LanguageRuntime::GuessLanguageForSymbolByName(Target &target,
...
-  for (enum LanguageType language : languages_to_try) {
-    language_runtime = process_sp->GetLanguageRuntime(language);
-    if (language_runtime) {
-        if (language_runtime->IsSymbolARuntimeThunk(symbol))
-          return true;
-    }
-  }
-  return false;
+bool LanguageRuntime::IsSymbolAnyRuntimeThunk(Symbol &symbol) {
+  return SwiftLanguageRuntime::IsSymbolARuntimeThunk(symbol);
```
adrian-prantl, line 348: Since `LanguageRuntime::IsSymbolAnyRuntimeThunk()` is a swift-lldb internal addition, the LHS really looks like premature generalization to me. Either we add no-op support in llvm.org or we just don't bother, like in this patch. I'm leaning towards the latter.

dcci:
@adrian-prantl I'll cherry-pick, sure.

PR #1143, opened 2018-12-14
Support unknown references to Objective-C classes.

adrian-prantl:
On Darwin the Swift runtime stores references to Objective-C
objects as a pointer to a struct that has the actual object
pointer at offset zero. The least significant bit of the
reference pointer indicates whether the reference refers to an
Objective-C or Swift object.

This is a property of the Swift runtime(!). In the future it
may be necessary to check for the version of the Swift runtime
(or indirectly by looking at the version of the remote
operating system) to determine how to interpret references.

This patch teaches the SwiftLanguageRuntime about this and handles
unwoned references accordingly whengetting the data of any
ValueObjectChildren (which classes always are).

<rdar://problem/42300829>

### DIFF
--- a/include/lldb/Target/LanguageRuntime.h
+++ b/include/lldb/Target/LanguageRuntime.h
@@ -105,6 +105,10 @@ public:
   virtual TypeAndOrName FixUpDynamicType(const TypeAndOrName &type_and_or_name,
                                          ValueObject &static_value) = 0;
 
+  /// This allows a language runtime to adjust references depending on the type.
+  /// \return true on success.
+  virtual bool FixupReference(lldb::addr_t &addr, CompilerType type) {}
+
   virtual void SetExceptionBreakpoints() {}
 
   virtual void ClearExceptionBreakpoints() {}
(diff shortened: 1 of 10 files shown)


adrian-prantl:
@swift-ci test

jrose-apple:
This only happens for `unowned` references, not all references.

jimingham:
> This only happens for `unowned` references, not all references.

Adrian does check for "unowned" in the Swift implementation of FixupReference (line 2637).  It wouldn't be appropriate for the generic code in ValueObjectChild to do this check.

adrian-prantl:
I clarified the comment and commit message.

adrian-prantl:
@swift-ci test and merge

jimingham (approved):
LGTM - this seems like a reasonable addition to the ValueObject system.  Thanks for the thorough test coverage.

We will need to put the generic part of this change back on llvm.org.  We already have some API's there (like pretty much all of the REPL support) that have no actual implementations on llvm.org, so this kind of change has precedent.

dcci (approved):
LGTM.

PR #1151, opened 2018-12-18
Call DeleteCurrentProcess before we replace the old process.

jimingham:
We need to ensure that Finalize gets called before we start
to destroy the old Process or the weak_ptr->shared_ptr link
from Threads to Target gets broken before the threads are 
destroyed.

<rdar://problem/43586979>

Differential Revision: https://reviews.llvm.org/D55631

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@349435 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #1152, opened 2018-12-18
Playground console: Don't misreport or suppress line numbers for warnings / errors

alexbbrown:
Replace specific line number adjustment with generic adjustment 1, when wrapping playground text for evaluation.

This fix prevents console errors with (in-editor) line numbers less than 7 from being suppressed totally, and correctly reports line numbers for console errors with (in-editor) line numbers equal or larger than 7.

<rdar://problem/46408318>


alexbbrown:
@swift-ci please test

dcci:
@alexbbrown This commit without testing makes me a little nervous. On one hand I like the fact it removes some hardcoded magic, OTOH, I'm a little afraid this will break again. Is there an easy way to trigger this behaviour from lldb?

alexbbrown:
Would you like to see a test case?  That's a great idea.  Please decline
the PR and I'll look see what can be added

dcci:
Thanks.

dcci (dismissed):
Marking as "request changes" while Alex tries to craft a testcase.

alexbbrown:
I have upstreamed the existing playground REPL tests from Apple's repo, and will built on those to test this PR.

https://github.com/apple/swift-lldb/pull/1158

alexbbrown:
Force pushed.  This change now depends upon https://github.com/apple/swift-lldb/pull/1158

alexbbrown:
@swift-ci please test

alexbbrown:
'before the patch' but with the new test:

./dotest.py --filecheck cat --log-success  testcases/lang/swift/playgrounds-repl/invalid_input/ 

AssertionError: ':15:3: error: left side of mutating operator' not found in "error: error: <REPL>:9:3: error: left side of mutating operator isn't mutable: 'r' is a 'let' constant\nr *= 2\n~ ^\n\n" : Error line number does not match


note: --filecheck cat is not correct usage, but I don't have time to work out what is.

Afterwards: 

passes!

alexbbrown:
@swift-ci please test

alexbbrown:
@swift-ci please test

PR #1153, opened 2018-12-18
[SwiftLanguageRuntime] Fix address projection of opaque existentials.

dcci:
<rdar://problem/44936803>


dcci:
@jimingham FYI, this will be gone once your fix for the expression parser gets in.

dcci:
@swift-ci test and merge

PR #1157, opened 2018-12-18
Un-conditionalize use of libcompression.  debugserver only builds

jasonmolenda:
Un-conditionalize use of libcompression.  debugserver only builds
on Darwin systems and libcompression has been in the OS for over
three years. 

Remove use of / linking to zlib.  We'll always have libcompression
available now.

Create a scratch buffer via compression_encode_scratch_buffer_size()
and use it in calls to compression_encode_buffer() to avoid 
compression_encode_buffer having to malloc & free a scratch buffer
on each call.  

Tested by forcing compression to be enabled on macos native (normally
only enabled on iOS et al devices), running the testsuite.

<rdar://problem/41601084>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@349553 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit d04d395a740ed6fa00292de75c1ba7c4f59ee478)


jasonmolenda:
@swift-ci test and merge

PR #1159, opened 2018-12-19
Force libcompression calls to be enabled when building on Darwin

jasonmolenda:
Force libcompression calls to be enabled when building on Darwin
systems.  It has been available in the OS over over three years
now.  If lldb doesn't link against -lcompression, it should be an
error.

Allocate a scratch buffer for libcompression to use when decoding
packets, instead of it having to allocate & free one on every call.

Fix a typeo with the size of the buffer that compression_decode_buffer()
is expanding into.

<rdar://problem/41601084>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@349563 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 44c2e3041882f2e78f637bc601ab79e50a532c5b)


jasonmolenda:
@swift-ci test and merge

PR #1163, opened 2018-12-19
Playground console: Don't misreport or suppress line numbers for warnings / errors

alexbbrown:
Replace specific line number adjustment with generic adjustment 1, when wrapping playground text for evaluation.

This fix prevents console errors with (in-editor) line numbers less than 7 from being suppressed totally, and correctly reports line numbers for console errors with (in-editor) line numbers equal or larger than 7.

rdar://problem/46408318


alexbbrown:
@swift-ci please test

jimingham (changes requested):
So if I'm reading this code correctly, that magic "7" was to handle skipping over the playground_prefix when it happens to be playground_logger_declarations.  That used to be 7, though it is 8 now...  That only partially makes sense, since if my guess is right somebody bobbled the thing, because it should really count the newlines in playground_prefix.  Presumably this is working with your change because the first_expression always has pound_file and pound_line set?  So I'm not sure we would ever actually find our way to this code IRL...

Anyway, if my understanding is correct, then maybe it would be better to set first_body_line in the else branch like:

StringRef str_ref(playground_prefix);
first_body_line = str_ref.count('\n') + 1

I am pretty sure that's the intent.

alexbbrown:
hi Jim

You are correct, that IS the intent, and it never worked correctly because
nobody maintained the delicate balance with the number of new lines in the
define block.

However, since that was implemented playgrounds started sending pound
directives directly in the source text, and not via the dedicated variable
mentioned in that function.

That had the unfortunate effect of applying the line number offset on top
of the modified (correct) alignment provided by the unmanaged pound line in
the source text, making it incorrect (or causing it to suppress errors
entirely, to the users presumed confusion).

The correct way to fix this would be to have playgrounds code in Xcode use
the pound line API, but that requires modifying two code bases with
distinct release cycles in synchrony.

My fix in this PR neatly changes output from incorrect to correct, and adds
a test to defend that in future.

I agree that it's not intuitive unless you trace down a whole bunch of
things going on elsewhere in the code.

Let me know what you want to do.

jimingham:
Grr...  That's confusing.  

So one solution to that problem would be to search the incoming text for #sourceLocation and if that is present, then set the beginning line to 1, and otherwise count the newlines in the prefix.  That would be correct and understandable on the lldb side.

If that seems too annoying to do, then can you put in a comment explaining why you had to do this?  Your test will fail when this gets changed somehow (for which thanks!!!), but then pity the poor person who has to figure out what is going on if we're not still around...

alexbbrown:
@jimingham Please take a look at this fix (which also passes testing).

jimingham (approved):
That looks fine.  Thanks for cleaning this up.

alexbbrown:
@swift-ci please test

CodaFi:
@swift-ci please test

alexbbrown:
Looks like there's still a bug in the non-error path in this code.  *testing*.

PR #1164, opened 2018-12-19
[Swift] Strip spare bits from pointers before printing them.

dcci:
SR-7565
<rdar://problem/39869332>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/enum_tagged/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/enum_tagged/main.swift
@@ -1,0 +1,43 @@
+struct A {
+    private class X {
+        let xx: Int
+
+        init(xx: Int) {
+            self.xx = xx
+        }
+    }
+    private class Y {
+        let yy: Int
+
+        init(yy: Int) {
+            self.yy = yy
+        }
+    }
+
+    private enum Variant {
+        case x(X)
+        case y(Y)
+        case z
+    }
+
+    private let variant: Variant
+
+    init(x: Int) {
+        variant = .x(X(xx: x))
+    }
+
+    init(y: Int) {
+        variant = .y(Y(yy: y))
+    }
+
+    init(z: Void) {
+        variant = .z
+    }
+}
+
+let x = A(x: 42)
+let y = A(y: 39)
+print("!")  //%self.expect("frame var -d run-target -- x", substrs=['(xx = 42)'])
+            //%self.expect("expr -d run-target -- x", substrs=['(xx = 42)'])
+            //%self.expect("frame var -d run-target -- y", substrs=['(yy = 39)'])
+            //%self.expect("expr -d run-target -- y", substrs=['(yy = 39)'])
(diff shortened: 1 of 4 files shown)


dcci:
@swift-ci please test

dcci:
@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/enum_tagged/main.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    init(z: Void) {
+        variant = .z
+    }
+}
+
+let x = A(x: 42)
+let y = A(y: 39)
+print("!")  //%self.expect("frame var -d run-target -- x", substrs=['(xx = 42)'])
```
adrian-prantl, line 40: I recently learned about `lldbutil.check_variable()`, which seems to be slightly less fragile / string-based than self.expect. Is it possible to use it inside of an inline test?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2639,6 +2639,26 @@ bool SwiftLanguageRuntime::FixupReference(lldb::addr_t &addr,
     }
   }
   default:
+    // Adjust the pointer to strip away the spare bits.
```
adrian-prantl, line 2642: Can you add a comment whether this is runtime or compiler-specific?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2639,6 +2639,26 @@ bool SwiftLanguageRuntime::FixupReference(lldb::addr_t &addr,
...
+      addr &= ~SWIFT_ABI_ARM_SWIFT_SPARE_BITS_MASK;
+      break;
+    case llvm::Triple::ArchType::x86:
+      addr &= ~SWIFT_ABI_I386_SWIFT_SPARE_BITS_MASK;
+      break;
+    case llvm::Triple::ArchType::x86_64:
+      addr &= ~SWIFT_ABI_X86_64_SWIFT_SPARE_BITS_MASK;
+      break;
+    default:
+      llvm_unreachable("unsupported arch");
```
adrian-prantl, line 2659: This is not unreachable. Swift supports powerpc and whatever arch z/os uses.

adrian-prantl: I would just leave the value as is on other arches.

dcci: OK.

packages/Python/lldbsuite/test/lang/swift/enum_tagged/main.swift
jimingham: The commands in lldbinline.py are run by "exec(command_string") in a method of the test class defined in lldbinline.py.  That file imports lldbutil.  IIUC how python exec works, the command string should see whatever the surrounding context sees.  So you should be able to use anything from lldbutil in the  test commands.

PR #1165, opened 2018-12-19
[Swift] Strip spare bits from pointers before printing them.

dcci:
SR-7565
<rdar://problem/39869332>


dcci:
@swift-ci test and merge

PR #1166, opened 2018-12-19
Fix a bug in PerformAutoImport that triggered when Swift was built wi…

adrian-prantl:
…th debug info.


adrian-prantl:
@swift-ci test and merge

PR #1173, opened 2018-12-21
"help finish" tells you it is an alias.  "help fin" doesn't.

jimingham:
They both run the same command, and people get used to typing the shortest
string they can, so we should support alias info on shortened strings as well.

<rdar://problem/46859207>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@349874 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

dcci:
This was failing for an unrelated reason.

dcci:
@swift-ci test and merge

PR #1176, opened 2018-12-21
unxfail test

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
Ah.. so this only works on macOS?

dcci:
hmm, this particular one should work everywhere.

PR #1177, opened 2018-12-24
Fix file_idx type to uint16_t from 11 bits.

frozenrainyoo:
file_idx type is NOT uint16_t, provide 11 bits.

if file_idx is 2691
file_idx = 2691 (0b101010000011) <-- 12bits

file_idx provides 11 bits.
so, file_idx variable changed from 2691(0b101010000011) to 643(0b01010000011).

https://bugs.llvm.org/show_bug.cgi?id=39816

### DIFF
--- a/include/lldb/Symbol/LineTable.h
+++ b/include/lldb/Symbol/LineTable.h
@@ -315,21 +315,22 @@ protected:
                      ///number information.
     uint16_t column; ///< The column number of the source line, or zero if there
                      ///is no column information.
-    uint16_t file_idx : 11, ///< The file index into CompileUnit's file table,
-                            ///or zero if there is no file information.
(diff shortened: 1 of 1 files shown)


adrian-prantl:
Thanks for the patch. This should go into lldb on llvm.org first and then should be cherry-picked here if need be.

adrian-prantl (changes requested):
We need to land this upstream first and then cherry-pick it to swift-5.0-branch. The stable branch will get it automatically the next time we rebranch from upstream-with-swift.

adrian-prantl:
I created https://reviews.llvm.org/D56218 for the upstream patch.

frozenrainyoo:
Thank you.
I will check this bug in next version.
we have being used custom lldb. T T

PR #1178, opened 2019-01-02
DWARF: Fix a bug in array size computation

adrian-prantl:
Summary:
r346165 introduced a bug, where we would fail to parse the size of an
array if that size happened to match an existing die offset.

The logic was:
if (DWARFDIE count = die.GetReferencedDie(DW_AT_count))
  num_elements = compute_vla_size(count);
else
  num_elements = die.GetUsigned(DW_AT_count); // a fixed-size array

The problem with this logic was that GetReferencedDie did not take the
form class of the attribute into account, and would happily return a die
reference for any form, if its value happened to match some die.

As this behavior is inconsistent with how llvm's DWARFFormValue class
operates, I chose to fix the problem by making our version of this class
match the llvm behavior. For this to work, I had to add an explicit form
class check to the .apple_XXX tables parsing code, because they do
(incorrectly?) use data forms as die references.

Reviewers: aprantl, clayborg

Subscribers: JDevlieghere, lldb-commits

Differential Revision: https://reviews.llvm.org/D55991

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@350086 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 1aead698a042d234b9213bdaf9e91b848fcda219)

### DIFF
--- a/lit/SymbolFile/DWARF/array-sizes.s
+++ b/lit/SymbolFile/DWARF/array-sizes.s
@@ -1,0 +1,147 @@
+# This tests a bug where we would incorrectly parse the size of an array if that
+# size happened to match an existing DIE offset. This happened because we
+# misinterpreted that value as a reference to a DIE specifying the VLA size even
+# though the form was a data form (as it should be).
+
+# REQUIRES: lld
+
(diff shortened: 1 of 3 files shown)


adrian-prantl:
@swift-ci test and merge

PR #1183, opened 2019-01-07
[SwiftREPL] Runtime checks for exclusive access should trap.

dcci:
<rdar://problem/33831489>


dcci:
@swift-ci test and merge

PR #1185, opened 2019-01-08
build: normalise paths before use

compnerd:
Ensure that we normalise the paths to the CMake sytle paths before use.  If we
do not do this, we will fail to configure on Windows.


compnerd:
@swift-ci please test

compnerd:

JDevlieghere:
Any reason this shouldn't live upstream?

compnerd:
@JDevlieghere, yeah my reasoning for that is that the entire codepath that this is touching doesn't live upstream :-)  None of these variables are in the upstream build.

JDevlieghere:
@compnerd Aight, I assumed this was the result of Stefan's change this morning and vaguely remembered him touching these paths, hence the question. Thanks for confirming, this LGTM.

compnerd:
@swift-ci please test macOS platform

PR #1186, opened 2019-01-08
 [Swift] Add support for looking up Clang types using module debug info.

adrian-prantl:
Previously all Clang fallback types were assumed to be of type
(id). By reading the -gmodules breadcrumbs, we can do much better.


adrian-prantl:
test with https://github.com/apple/swift/pull/21709
@swift-ci test

PR #1187, opened 2019-01-08
[DNM] Add LLDB formatters for types that live primarily in FoundationBase.

millenomi:
This supports the change at https://github.com/apple/swift-corelibs-foundation/pull/1799.

### DIFF
--- a/source/Plugins/Language/Swift/SwiftLanguage.cpp
+++ b/source/Plugins/Language/Swift/SwiftLanguage.cpp
@@ -563,6 +563,8 @@ static void LoadSwiftFormatters(lldb::TypeCategoryImplSP swift_category_sp) {
                    ConstString("CoreGraphics.CGFloat"), summary_flags);
   AddStringSummary(swift_category_sp, "${var.native}",
                    ConstString("Foundation.CGFloat"), summary_flags);
+  AddStringSummary(swift_category_sp, "${var.native}",
+                   ConstString("FoundationBase.CGFloat"), summary_flags);
 #endif // LLDB_DISABLE_PYTHON
 }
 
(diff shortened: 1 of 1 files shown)


source/Plugins/Language/Swift/SwiftLanguage.cpp
```diff
@@ -563,6 +563,8 @@ static void LoadSwiftFormatters(lldb::TypeCategoryImplSP swift_category_sp) {
                    ConstString("CoreGraphics.CGFloat"), summary_flags);
   AddStringSummary(swift_category_sp, "${var.native}",
                    ConstString("Foundation.CGFloat"), summary_flags);
+  AddStringSummary(swift_category_sp, "${var.native}",
```
dcci, line 566: So, is `CGFloat` living exclusively in `FoundationBase`? If so, the other formatter can be removed.

millenomi: Only in swift-corelibs-foundation. Does this cover Darwin too?

millenomi: Also, I'd like to be able to debug things built with swift-corelibs-foundation and Swift 4.x :)

millenomi:
This is blocked by https://bugs.swift.org/browse/SR-9635.

source/Plugins/Language/Swift/SwiftLanguage.cpp
dcci: This covers everything (Darwin and Linux). If you really want to match both, use a regex instead of duplicating the code (see other examples, or ping me if you have issues).

Also, compiler and debugger are revlocked until swift grows a stable module format so the backcompatibility argument doesn't hold.

Thanks!

dcci:
The problem you have is that swift and lldb are out of sync.

dcci (changes requested):
Mark as "requested changes" so that it gets out of my review queue.

PR #1189, opened 2019-01-10
[Swift] Disable delayed parsing in the expression evaluator.

dcci:
<rdar://problem/38396444>


dcci:
@swift-ci test and merge

PR #1190, opened 2019-01-10
 Support zero-sized types in extensions.

adrian-prantl:
In order to correctly determine the size of a generic type it needs to
be mapped into the context first and archetypes need to be bound.

Note: It would be even better to change
CompilerType::GetBit/ByteSize() to return an Optional<uint64_t>, but
that needs happen upstream.

<rdar://problem/38231646>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -5090,8 +5090,32 @@ bool SwiftASTContext::IsSelfArchetypeType(const CompilerType &compiler_type) {
   return false;
 }
 
-bool SwiftASTContext::IsPossibleZeroSizeType(
-  const CompilerType &compiler_type) {
+static CompilerType BindAllArchetypes(CompilerType type,
+                                      ExecutionContextScope *exe_scope) {
+  if (!exe_scope)
+    return type;
+  auto *frame = exe_scope->CalculateStackFrame().get();
+  auto *runtime = exe_scope->CalculateProcess()->GetSwiftLanguageRuntime();
+  if (!frame || !runtime)
+    return type;
+  ExecutionContext exe_ctx;
+  exe_scope->CalculateExecutionContext(exe_ctx);
+  return runtime->DoArchetypeBindingForType(*frame, type);
+}
+
+bool SwiftASTContext::IsPossibleZeroSizeType(const CompilerType &compiler_type,
+                                             ExecutionContextScope *exe_scope) {
+  if (compiler_type && exe_scope) {
+    swift::CanType swift_can_type(GetCanonicalSwiftType(compiler_type));
+    if (swift_can_type->hasTypeParameter()) {
+      ExecutionContext exe_ctx;
+      exe_scope->CalculateExecutionContext(exe_ctx);
+      auto swift_scratch_ctx_lock = SwiftASTContextLock(&exe_ctx);
+      return IsPossibleZeroSizeType(BindAllArchetypes(compiler_type, exe_scope),
+                                    nullptr);
+    }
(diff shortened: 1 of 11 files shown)


adrian-prantl:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -5897,27 +5920,37 @@ SwiftASTContext::GetSwiftFixedTypeInfo(void *type) {
...
-      return fixed_type_info->getFixedSize().getValue() * 8;
+    ExecutionContext exe_ctx;
+    exe_scope->CalculateExecutionContext(exe_ctx);
+    auto swift_scratch_ctx_lock = SwiftASTContextLock(&exe_ctx);
+    CompilerType bound_type = BindAllArchetypes({this, type}, exe_scope);
+    // Note thay the bound type may be in a different AST context.
+    return bound_type.GetBitSize(exe_scope);
   }
+
+  const swift::TypeKind type_kind = swift_can_type->getKind();
```
slavapestov, line 5939: This part makes no sense. LValues, UnboundGenerics and GenericFunctions are not real types -- nothing in memory has that type. And Functions are not a single pointer in size.

adrian-prantl: That's part of the original code I didn't touch. I'll see if I can just remove those cases.

adrian-prantl:
@swift-ci test

source/Symbol/SwiftASTContext.cpp
```diff
@@ -5090,8 +5090,32 @@ bool SwiftASTContext::IsSelfArchetypeType(const CompilerType &compiler_type) {
   return false;
 }
 
-bool SwiftASTContext::IsPossibleZeroSizeType(
-  const CompilerType &compiler_type) {
+static CompilerType BindAllArchetypes(CompilerType type,
```
dcci, line 5093: Maybe this can be made public in `SwiftASTContext`? 
Seems something could be of more general use.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2042,6 +2042,7 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Promise(
     address.SetLoadAddress(val_ptr_addr, &m_process->GetTarget());
     return true;
   } break;
+  case swift::MetadataKind::Function:
```
dcci, line 2045: Is this needed?

source/Symbol/SwiftASTContext.cpp
adrian-prantl: Removing those cases breaks:
    lldb-Suite :: lang/swift/variables/func/TestFunctionVariables.py
    lldb-Suite :: lang/swift/variables/inout/TestInOutVariables.py

It looks like at least LLDB itself is still generating LValue types.

dcci: We should really fix this. Can you file a rdar? (I'll be happy to take a stab at it).

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: No.

source/Symbol/SwiftASTContext.cpp
slavapestov: How do they break?

slavapestov: Looks like I already filed rdar://41608567, and it was closed. I guess there were two overloads of this function. Sigh...

PR #1193, opened 2019-01-11
Add XFAILed testcase for generic self bound to a private type.

adrian-prantl:
This is currently broken because of rdar://problem/38231646


adrian-prantl:
@swift-ci test and merge

PR #1195, opened 2019-01-11
 Ensure result variables of generic types are bound to the runtime type.

adrian-prantl:
Rather than mapping type in and out of contexts a follow-up commit
should make sure that CompilerType::MapIntoContext isn't called where
it isn't needed.

Fixes <rdar://problem/47157089>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/generic_self/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/generic_self/main.swift
@@ -2,45 +2,50 @@
 //
 // This source file is part of the Swift.org open source project
 //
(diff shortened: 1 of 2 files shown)


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

dcci (approved):
This is good for 5.0, as it's the minimal change to fix (for stable we might want to avoid the map-in/map-out dance).

dcci:
(as we already discussed).

PR #1204, opened 2019-01-15
Fix a use-after-free found by ASAN.

adrian-prantl:
rdar://problem/47270935


adrian-prantl:
@swift-ci test and merge

PR #1207, opened 2019-01-15
[SwiftArray] Don't assume pointer size is 8.

dcci:
Fixes yet another failure visible only on 32-bits architectures.
(TestSwiftieFormatting/TestNestedArray)

<rdar://problem/47272296>
<rdar://problem/47271440>


dcci:
@lorentey this probably regressed when we rewrote the formatter.

dcci:
@swift-ci test and merge

PR #1209, opened 2019-01-16
[DWARFASTParserSwift] Update after API change upstream.

dcci:
GetByteSize() now returns an optional to flag error conditions,
  rather than `0` as it did before, because, turns out, `0` is
  a valid size in swift (and somewhere else, presumably).

### DIFF
--- a/source/Expression/Materializer.cpp
+++ b/source/Expression/Materializer.cpp
@@ -538,15 +538,18 @@ public:
         Status extract_error;
         valobj_sp->GetData(data, extract_error);
         if (!extract_error.Success()) {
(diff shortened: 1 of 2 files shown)


dcci:
@adrian-prantl please take a look.

adrian-prantl:
Thanks, but I'm already working on a patch to update Swift — I just haven't gotten all testcases to pass yet. Is this urgent?

dcci:
@adrian-prantl this is blocking `master-next`. Do you have an ETA for the patch? I can hold this off for a bit, but I would like to start running the test suite to see what are the failures etc..

dcci:
Closing this as Adrian is working on an alternative.

PR #1210, opened 2019-01-16
Retire IsPossibleZeroSizeType()

adrian-prantl:
TypeSystem::GetSizeinBits() now supports 0 as a non-error value, which
makes this function redundant.

<rdar://problem/47178964>


adrian-prantl:
@swift-ci test

dcci (approved):
Thanks!

PR #1215, opened 2019-01-17
Add support for keeping around serialized modules in SwiftASTContext.

bgogul:
This change is needed for some language features to work across cells in notebook environments such as colab and Jupyter.  

e.g., Differentiation in Swift requires that the body of a SIL function be present for generating the proper adjoint code. Suppose that the following function is defined in one lldb expression (or equivalently in one colab/jupyter cell):
```
1> func square(_ x: Float) -> Float { return x * x } 
```
If `gradient` of square is invoked in a different lldb expression, it will not work unless the body of `square` is available:
```
2> gradient(at: 4, in: square)
```

This PR adds an option to serialize an lldb expression so that the body of `square` is available in other cells.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
@@ -91,6 +91,7 @@ SwiftExpressionParser::SwiftExpressionParser(
 
   // TODO This code is copied from ClangExpressionParser.cpp.
   // Factor this out into common code.
+
   lldb::TargetSP target_sp;
   if (exe_scope) {
     target_sp = exe_scope->CalculateTarget();
@@ -1961,7 +1962,7 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
   }
 
   runSILDiagnosticPasses(*sil_module);
-  
+
   // SWIFT_ENABLE_TENSORFLOW
   // FIXME: When partitioning joins the mandatory pass pipeline, we should be able to
   // stop running the optimization passes and drop the explicit call of the partitioning
@@ -1974,6 +1975,7 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
   // runs at -O0.  We need a proper deabstraction pass to do that though.
   runSILTFPartitionPass(*sil_module);
   // SWIFT_ENABLE_TENSORFLOW
+
 
   if (log) {
     std::string s;
--- a/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
@@ -71,6 +71,7 @@
 #include "swift/SIL/SILFunction.h"
 #include "swift/SIL/SILModule.h"
 #include "swift/SILOptimizer/PassManager/Passes.h"
+#include "swift/Serialization/SerializationOptions.h"
 #include "swift/Serialization/SerializedModuleLoader.h"
 #include "swift/Subsystems.h"
 
@@ -90,7 +91,6 @@ SwiftExpressionParser::SwiftExpressionParser(
 
   // TODO This code is copied from ClangExpressionParser.cpp.
   // Factor this out into common code.
-
   lldb::TargetSP target_sp;
   if (exe_scope) {
     target_sp = exe_scope->CalculateTarget();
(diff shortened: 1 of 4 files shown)


source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -90,7 +91,6 @@ SwiftExpressionParser::SwiftExpressionParser(
 
   // TODO This code is copied from ClangExpressionParser.cpp.
   // Factor this out into common code.
-
```
marcrasi: undo this

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -1940,8 +1940,28 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
     log->PutCString(s.c_str());
   }
 
-  runSILDiagnosticPasses(*sil_module);
+  // Serialize the file if modules directory is set.
+  if (auto expr_module_dir = swift_ast_ctx->GetReplExprModulesDir()) {
```
marcrasi, line 1952: Could this, and the chunk of code on L2043 that initializes `module` happen at the same time, to keep it all in the same place? It would serialize it to a file and then immediately load it again.

Also, would it be possible to put it all around the end of `ParseAndImport` so that `ParseAndImport` returns a `module` that has been all correctly processed and no one gets confused that `parsed_expr->module` is not the "right" module to use in some places any more?

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -2020,7 +2039,37 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
...
+
+    lldb::StackFrameSP this_frame_sp(m_stack_frame_wp.lock());
+
+    if (this_frame_sp) {
+      lldb::ProcessSP process_sp(this_frame_sp->CalculateProcess());
+      Status error;
+      std::string module_name = parsed_expr->module.getName().str();
+      if (process_sp && !module_name.empty()) {
+        ConstString module_const_str(module_name);
+        module = swift_ast_ctx->FindAndLoadModule(module_const_str,
```
marcrasi, line 2071: Is there a fundamental reason why it has to be stored into a file, or is it that all the functions we need like `FindAndLoadModule` only operate on files?

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -1955,7 +1975,6 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
   runSILTFPartitionPass(*sil_module);
   // SWIFT_ENABLE_TENSORFLOW
 
-
```
marcrasi: undo

marcrasi (approved):
Looks good to try out for now, though I have some inline questions about simplification possibilities that we should think about eventually.

A lot of the new chunks don't have `SWIFT_ENABLE_TENSORFLOW`, could you add that before merging?

Was there some thread in the Swift forums where you asked about this? If so, could you post a link to this PR for more context about it?

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
bgogul: Done.

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
bgogul: Unfortunately, no. The module needs to be serialized just after SILGen. I added a brief explanation as to why that is necessary. See above.

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
bgogul: Yes. [SerializedSILLoader::lookupSILFunction](https://github.com/apple/swift/blob/9fa91e97aa98990c78cf2595960bb40cf0772abc/lib/Serialization/SerializedSILLoader.cpp#L48) only searches for a function in serialized SIL modules in a given AST context. The serialized SIL modules are stored in [LoadedSILSections](https://github.com/apple/swift/blob/9fa91e97aa98990c78cf2595960bb40cf0772abc/lib/Serialization/SerializedSILLoader.cpp#L36).  SerializedSILLoader::lookupSILFunction is how Differentiation and other passes (like inlining) find the implementation of SILFunctions.

bgogul (commented):
Thanks for the comments, Marc. I have addressed them and will merge it in now. 

> A lot of the new chunks don't have SWIFT_ENABLE_TENSORFLOW, could you add that before merging?

Done

> Was there some thread in the Swift forums where you asked about this? If so, could you post a link to this PR for more context about it?

Yes there was a  [thread](https://forums.swift.org/t/generating-serialized-sil-for-inlinable-transparent-functions-in-repl-expressions/16237/6), but we did not settle on a final design.

PR #1218, opened 2019-01-17
[test] Fix the definition of getCompilerVersion().

dcci:
Fixes a bunch of failures on upstream-with-swift.

### DIFF
--- a/packages/Python/lldbsuite/test/lldbtest.py
+++ b/packages/Python/lldbsuite/test/lldbtest.py
@@ -1295,14 +1295,12 @@ class Base(unittest2.TestCase):
         """Returns the compiler binary the test suite is running with."""
         return self.getCompiler().split()[0]
 
-    def getCompilerVersion(self):
+    def getCompilerVersion(self, compiler=None):
         """ Returns a string that represents the compiler version.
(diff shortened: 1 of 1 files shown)


dcci:
For the curious ones, it's the incarnation of the issue described here:
https://stackoverflow.com/questions/10202938/how-do-i-use-method-overloading-in-python/26554662

Also, `self` as default argument for methods doesn't really work so we have to use the `None` trick.

packages/Python/lldbsuite/test/lldbtest.py
```diff
@@ -1295,14 +1295,12 @@ def getCompilerBinary(self):
...
-    def getCompilerVersion(self):
+    def getCompilerVersion(self, compiler=None):
         """ Returns a string that represents the compiler version.
             Supports: llvm, clang.
         """
-        compiler = self.getCompilerBinary()
-        self.getCompilerVersion(compiler)
-
-    def getCompilerVersion(self, compiler):
+        if compiler is None:
```
JDevlieghere, line 1302: More Pythonic would be `if not compiler:` but I think this is fine.

PR #1219, opened 2019-01-17
[CMake] Fix RPATH for repl

JDevlieghere:
Due to r350391 we could no longer construct the correct RPATH for the
repl. In order to construct the correct path we need to get the path to
the LLDB framework by reading the LIBRARY_OUTPUT_DIRECTORY property from
liblldb. We also need the LLDB_FRAMEWORK_RESOURCE_DIR which was
previously removed, and re-introduced in this patch.


JDevlieghere:
@weliveindetail I'm going to merge this to unblock the REPL tests, but feel free to revert this if you don't think it's the right solution.

PR #1223, opened 2019-01-17
[lit] Add swift support to toolchain.py

JDevlieghere:
Zachary introduced some changes to substitutions in lit in r347216. This re-adds support for Swift.


dcci (commented):
LGTM

PR #1224, opened 2019-01-17
[lit] Add `-frontend` when expanding `%target-swift-frontend`.

dcci:


JDevlieghere:
Good catch, thank you!

PR #1231, opened 2019-01-18
[Swift] Fix native string formatter for 32-bits.

dcci:
<rdar://problem/47162433>


dcci:
@swift-ci test and merge

PR #1232, opened 2019-01-19
 Add a missing return statement to the default implementation of Fixup...

adrian-prantl:
…Reference.

This is supposed to be a successful no-op everywhere but in Swift.

<rdar://problem/47402084>


adrian-prantl:
@swift-ci test and merge

jasonmolenda:
Yep, looks good.

jasonmolenda (approved):
Yep, looks good.

PR #1234, opened 2019-01-23
Add initialization of the Swift runtime to Windows

lanza:
Swift builds and runs on Windows now and the LanguageRuntime and
Language subclasses are necessary to properly debug executables.


lanza:
@compnerd It still fails with this patch, but at least it fails at a different spot.

compnerd:
Sounds like an improvement ...

compnerd (approved):

compnerd:

PR #1236, opened 2019-01-24
[SwiftLanguageRuntime] Remove an assertion that was too aggressive.

dcci:


dcci:
@swift-ci test and merge

PR #1237, opened 2019-01-24
Treat a uuid of all 0's coming from debugserver as an invalid UUID.

jimingham:
That is what it means.

<rdar://problem/47226381>


jimingham:

jasonmolenda:
Yeah, looks good to me.  Tests to be added when upstreamed.

PR #1239, opened 2019-01-24
[SwiftLanguageRuntime] Try to be more defensive.

dcci:
<rdar://problem/47348115>


dcci:
@swift-ci test and merge

fredriss:
Please put this on 5.0 too, it's obviously safe and can avoid a crash.

jimingham:
Could you also change "M" to a variable name consistent with lldb conventions?  Local variables are of the form "foo_bar" with a lower case first letter and we discourage one letter variables.

jasonmolenda:
Target::GetExecutableModule returns a ModuleSP so the convention is that this local is called module_sp.

PR #1240, opened 2019-01-24
[CMake] Don't clobber linker flags already set

bulbazord:
Setting `CMAKE_EXE_LINKER_FLAGS` directly here clobbers any linker flags set ahead of time (e.g. in a cmake cache). Instead of overwriting `CMAKE_EXE_LINKER_FLAGS`, let's just append to the end of it.


bulbazord:
@swift-ci please test

compnerd (approved):
err, yes, we should not be overwriting these variables

compnerd:

dcci (approved):
lgtm.

PR #1243, opened 2019-01-25
[dotest.py] Canonicalize most Swift Makefiles that build dynamic libraries

adrian-prantl:
Canonicalize most Swift Makefiles that build dynamic libraries
towards always using the rules in Makefiles.rules for building dylibs
rather than rolling our own potentially slightly different Swift
compiler and linker incovations.

<rdar://problem/47498714>


adrian-prantl:
@swift-ci test

packages/Python/lldbsuite/test/lang/swift/clangimporter/bridging_header_headermap/Makefile
```diff
@@ -1,12 +1,12 @@
 LEVEL = ../../../../make
-SRCDIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 SWIFT_OBJC_INTEROP := 1
-USESWIFTDRIVER := 1
-# This Makefile overwrites the default rules.
+SWIFT_SOURCES := main.swift
 MAKE_DSYM := NO
-EXE :=
+LD_EXTRAS := -ldylib -L.
+SWIFTFLAGS_EXTRAS = -import-objc-header $(SRCDIR)/bridging-header.h \
```
compnerd, line 6: Please add an `-enable-objc-interop` to the flags to ensure that this works on all targets

packages/Python/lldbsuite/test/lang/swift/clangimporter/bridging_header_headermap/Makefile
```diff
@@ -16,35 +16,11 @@ all: a.out
...
-endif
-	$(DSYMUTIL) a.out
-
-libDylib: dylib.swift
-	$(SWIFTC) -g -Onone $^ -emit-library -module-name Dylib -emit-module \
-	  -Xlinker -install_name -Xlinker @executable_path/$@.dylib $(SWIFTFLAGS)
-ifneq "$(CODESIGN)" ""
-	$(CODESIGN) -s - "$@"
-endif
+libdylib.dylib: dylib.swift
```
compnerd, line 19: Why hard code the `dylib` extension?  The bridging header stuff actually does work on other targets

packages/Python/lldbsuite/test/lang/swift/clangimporter/bridging_header_headermap/Makefile
adrian-prantl: This test is marked as `@skipUnlessDarwin`. Which is probably why the old version didn't bother with it either. I think porting the test to Linux or other systems would be great, but I'd like to keep this PR as targeted as possible. It's already pretty big :-)

packages/Python/lldbsuite/test/lang/swift/clangimporter/bridging_header_headermap/Makefile
adrian-prantl: This test uses header maps, so it's kind of Darwin-specific. Headermaps *work* on other platforms, but I've never seen them generated outside of Xcode. Thankfully.

compnerd: Just FYI, buck does generate headermaps as well (and they are used pretty extensively elsewhere as well)

adrian-prantl: I was happier before I knew that.

PR #1244, opened 2019-01-25
[CMake] Minor cleanups to LLDBStandalone.cmake

bulbazord:
I plan on cherry-picking some changes that I and others made upstream to swift-lldb, specifically replacing the use of llvm-config with the LLVM CMake package. I noticed that this file has diverged a decent amount from upstream, and so to make the cherry-picking easier I cleaned up some things in this file, e.g. removing unused variables and inlining variables used once.o

cc @dcci @compnerd @weliveindetail


compnerd:
I think that we should also use the LLVM CMake package like clang here.  But, that can be a subsequent change.  LGTM.

compnerd (approved):

compnerd:
@swift-ci please test

bulbazord:
The changes I plan to cherry-pick will include the use of the LLVM CMake package instead of llvm-config. :)

Additionally, I would like to use variables from the Swift CMake package but that currently needs work. I started improving it with https://github.com/apple/swift/pull/21388, but I need to fix up a few things there.

compnerd:
@swift-ci please test

PR #1246, opened 2019-01-25
Simplify Swift-specific logging code (NFC)

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

jimingham:
From the naming it looks like somebody intended the verbose log to only be set if you said "log enable -v", but then the "| LLDB_LOG_OPTION_VERBOSE got lost.  Do we know that we still want this particular message to always come out in the logs?

PR #1247, opened 2019-01-25
Remove a warning in DynamicLoaderDarwin::UpdateImageLoadAddress

jasonmolenda:
Remove a warning in DynamicLoaderDarwin::UpdateImageLoadAddress
when the binary loaded in memory has a section that we cannot find
in the on-disk version.  I added this warning out of an overabundance
of caution originally, but I've never seen an instance of it being
hit in the past few years, and there are some changes for the shared
cache on darwin systems where a segment is added when the shared
cache is constructed so we're now hitting this warning.  I've decided
to remove it altogether.

<rdar://problem/46889346> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@352158 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 6e6d86341cf7e8b0b3c86316b724b3c668b95391)


jasonmolenda:
@swift-ci test and merge

PR #1249, opened 2019-01-25
[Dictionary] Fix displaying of dictionary with > 12 elements on 32-bits.

dcci:
<rdar://problem/47271307>

### DIFF
--- a/source/Plugins/Language/Swift/SwiftHashedContainer.cpp
+++ b/source/Plugins/Language/Swift/SwiftHashedContainer.cpp
@@ -493,8 +493,6 @@ NativeHashedStorageHandler::NativeHashedStorageHandler(
   if (!scale_sp)
     return;
   auto scale = scale_sp->GetValueAsUnsigned(0);
-  if (scale > m_ptr_size)
-    return;
   m_scale = scale;
 
   auto keys_ivar = value_type ? g__rawKeys : g__rawElements;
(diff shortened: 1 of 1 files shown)


dcci:
This fixes TestSwiftStdlibDictionary.py on 32-bits.

dcci:
@swift-ci test and merge

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -493,8 +493,6 @@ NativeHashedStorageHandler::NativeHashedStorageHandler(
   if (!scale_sp)
     return;
   auto scale = scale_sp->GetValueAsUnsigned(0);
-  if (scale > m_ptr_size)
-    return;
```
lorentey, line 497: Hm, I don't think we found out why this was failing. Scales larger than 31/63 would wreak havoc in `GetBucketCount`:

```c++
  // Line 88
  size_t GetBucketCount() { return 1 << m_scale; }
```

(Related: this should probably be changed to read `1ULL << m_scale`, as well.)

dcci: This is `m_ptr_size` in /bytes/, i.e. 4. Should we have a check for this (but in bits?)

dcci: i.e. `(scale > m_ptr_size * 8)`

lorentey: Argh, of course. 🤭 That's simultaneously a relief and a huge face palm.

This wasn't a super important condition -- it was just a sanity check against trying to interpret corrupt data. (In a valid Set/Dictionary, the scale should be between 0 and 32/64, depending on pointer size in bits.) I think we can safely omit it; subsequent code will fail when it tries to access memory anyway.

dcci: Great :)

PR #1250, opened 2019-01-26
UnXFAIL test; this has been fixed in the compiler.

adrian-prantl:
rdar://problem/38231646

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/private_generic_self/TestSwiftPrivateGenericSelf.py
+++ b/packages/Python/lldbsuite/test/lang/swift/private_generic_self/TestSwiftPrivateGenericSelf.py
@@ -12,5 +12,4 @@
 import lldbsuite.test.lldbinline as lldbinline
 from lldbsuite.test.decorators import *
 
-lldbinline.MakeInlineTest(__file__, globals(), decorators=[swiftTest,
-    expectedFailureAll(bugnumber="rdar://problem/38231646")])
+lldbinline.MakeInlineTest(__file__, globals(), decorators=[swiftTest])
--- a/packages/Python/lldbsuite/test/lang/swift/private_generic_self/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/private_generic_self/main.swift
@@ -18,7 +18,7 @@ struct MyStruct<S : MyKey> : MyProtocol {
 }
 extension MyProtocol {
   func decode() {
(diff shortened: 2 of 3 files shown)


adrian-prantl:
test with https://github.com/apple/swift/pull/22136
@swift-ci test

adrian-prantl:
@swift-ci test

adrian-prantl:
`11:33:33 AssertionError: False is not True : expected typename: '(a.MyStruct<a.Outer.CodingKeys>)' - actual typename: 'τ_0_0'`

@DougGregor Should we expect this to work after https://github.com/apple/swift/pull/22136 is merged?

DougGregor:
RemoteAST was doing the right thing in my tests, and the failure looks like something involving the use of archetypes. Didn’t you have another change in flight that also affected this example?

adrian-prantl:
I'll take a closer look at where this is failing.

adrian-prantl:
@swift-ci test and merge

DougGregor (approved):
Great! Based on my other improvements to RemoteAST (in https://github.com/apple/swift/pull/22150), "frame var" should start working for local types and types in extensions as well. The RemoteAST tests I added in that PR could have LLDB-side counterparts to make sure everything is plumbed through.

adrian-prantl:
The problem was that while `swift -g` implies `-enable-anonymous-context-mangled-names`, `swift -frontend`'s -g does not and LLDB's testsuite Makefiles are calling the Swift frontend directly. I added the option to the frontend flags in the master Makefile fragment.

I need to check whether xcbuild and swiftpm are doing the right thing, since they might also be calling the swift frontend directly.

DougGregor:
I was resisting wiring `-g` up to `-enable-anonymous-context-mangled-names` in the frontend on principle, but we could do that if this is going to turn into a game of build-system whack-a-mole.

adrian-prantl:
Since `-enable-anonymous-context-mangled-names` changes what metadata is being produced it violates the promise that `-g` should have no effect on the stripped binary. Perhaps we should make it on by default and only provide a negative variant to turn it off if desired? Or we could tie it to `-Onone` instead of `-g`, which is what we do for other IRGen changes that benefit debugging.

DougGregor:
We didn't want to turn it on by default for fear of bloating code size. Tying it to `-Onone` rather than `-g` would be acceptable, I guess, although it's very much intended for debugging: Swift's runtime won't look at this information (intentionally).

adrian-prantl:
I created https://github.com/apple/swift/pull/22213 to key `-enable-anonymous-context-mangled-names` off the optimization level. Ideally we should figure out a way to extract the extra metadata into a section in the `.dSYM` bundle, but I don't know enough about the metadata format to understand how feasible this is at this point.

PR #1252, opened 2019-01-28
[CMake] Replace use of llvm-config with LLVM's CMake package and find NATIVE llvm-tblgen

bulbazord:
This PR is a few cherry-picks from upstream. It is one major patch for replacing llvm-config with LLVM's CMake package as well as a few follow up patches.

Additionally, there is one other patch for finding the native version of llvm-tblgen when cross-compiling lldb standalone.


bulbazord:
cc @dcci @weliveindetail @compnerd

cmake/modules/LLDBStandalone.cmake
```diff
@@ -116,8 +87,8 @@ if (CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   endif()
 
   # Start Swift Mods
-  find_package(Clang REQUIRED CONFIG
-    HINTS "${LLDB_PATH_TO_CLANG_BUILD}" NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
+  #find_package(Clang REQUIRED CONFIG
+  #  HINTS "${LLDB_PATH_TO_CLANG_BUILD}" NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
```
compnerd: Why keep these around?

compnerd (commented):
I'm excited about where this is heading!  I think that we should go a step further and remove `LLDB_PATH_TO_LLVM_BUILD` and `LLDB_PATH_TO_CLANG_BUILD` and "rename" them in spirit to the CMake *standard* variables `LLVM_DIR` and `Clang_DIR` which CMake will actually look at implicitly when searching for the CMake package.  This avoids the custom logic here and parameters and can rely entirely on the standard in-package behaviour.  In fact, we could also do this for the swift package, and completely eliminate the 6 variables that current need to be specified in favour of 3 parameters that only need to be specified if the package is not in the system search path.

cmake/modules/LLDBStandalone.cmake
bulbazord: Whoops, meant to delete that. Will remove and re-push.

bulbazord:
> 
> 
> I'm excited about where this is heading! I think that we should go a step further and remove `LLDB_PATH_TO_LLVM_BUILD` and `LLDB_PATH_TO_CLANG_BUILD` and "rename" them in spirit to the CMake _standard_ variables `LLVM_DIR` and `Clang_DIR` which CMake will actually look at implicitly when searching for the CMake package. This avoids the custom logic here and parameters and can rely entirely on the standard in-package behaviour. In fact, we could also do this for the swift package, and completely eliminate the 6 variables that current need to be specified in favour of 3 parameters that only need to be specified if the package is not in the system search path.

That might be a good follow up item, since these are just cherry-picks. I like the ability to specify multiple paths in `LLDB_PATH_TO_{CLANG,LLVM}_BUILD` but I don't see a problem dropping that in favor of setting `{CLANG,LLVM}_DIR` directly. Definitely should be done upstream first though. :)

compnerd:
@swift-ci please test

dcci:
@xiaobai I don't have any objections to this patch but @weliveindetail should really sign it off as he touched this code extensively in the last months.

dcci:
@weliveindetail Stefan, can you please take a look?

bulbazord:
@weliveindetail friendly ping! Could you please look when you get the chance? Thanks!

weliveindetail (approved):
Looks good to me. In case you add https://reviews.llvm.org/D57402 when it's though, I believe we still need the `HINTS "${LLDB_PATH_TO_LLVM/CLANG_BUILD}"` as long as we don't use `LLVM/Clang_DIR`.

PR #1254, opened 2019-01-29
[Swift] Fix printing of nested C-like enums.

dcci:
<rdar://problem/47537143>


dcci:
@swift-ci test and merge

PR #1255, opened 2019-01-29
Use compactMap

moiseev:
The code in this test has been using the `flatMap` variant that is now marked is obsoleted in swift 5.0. See: https://github.com/apple/swift/pull/22138

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/variables/generic_struct_debug_info/generic_flatmap/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/variables/generic_struct_debug_info/generic_flatmap/main.swift
@@ -33,13 +33,13 @@ struct FlatMapper<Type>
             ValueAndIndices(originalIndex: 0, filteredIndex: 0, value: values[0]),
         ]
         
-        let _ = tuples.flatMap { tuple in
+        let _ = tuples.compactMap { tuple in
             return tuple //% self.expect('po tuple', substrs=['originalIndex : 0', 'filteredIndex : 0', 'name : "Coffee"', 'ID : "1"'])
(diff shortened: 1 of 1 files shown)


jimingham:
Is flatMap gone or just obsolete?  If it is present but marked obsolete, then people might still have it around and we should still test it works till it is actually gone.  In that case could you just duplicate the test but for compactMap?  Then we'll get rid of the flatMap test when it is actually gone.

OTOH, if flatMap is actually gone could you rename the test directory and files flatMap -> compactMap as well as the usage?  It's weird to have a test called generic_flatmap but it tests the compactMap.

Thanks!

moiseev:
> Is flatMap gone or just obsolete? If it is present but marked obsolete, then people might still have it around and we should still test it works

Yeah. But how do we compile this test?

When the methods is marked as `obsoleted: 5.0` the method is technically there, but the compiler will pretend it's not when the code is compiled with `-swift-version 5`, which is the default mode now.
It is probably possible to compile the test in Swift 4.2 mode, but that would be postponing the problem until the time when that mode is no longer supported.

jimingham:
While it is possible to use the toolset lldb is a part of to compile a program with "flatMap" in it, lldb should be able to print one of these, and we should test that it can.  So if passing -swift-version 4.2 to the compiler allows this file to compile, then we shouldn't delete the support for it.

If/when it is no longer possible to make a flatMap with our tools, then we should delete the test (and the support in the lldb data formatters for it).

That's not postponing, that's maintaining the support while it still can be used.

And at the same time we should add an independent test for compactMap, preferably one not called "flatmap".

moiseev:
We decided to just obsolete the method in 5.1, instead. This change is no longer needed.

PR #1259, opened 2019-01-30
Assorted Swift generics cleanups

slavapestov:
Fixes <rdar://problem/31822623>.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
@@ -230,7 +230,7 @@ __builtin_logger_initialize()
 
     const char *optional_extension =
         (language_flags & SwiftUserExpression::eLanguageFlagIsWeakSelf)
-            ? "Swift.Optional where Wrapped: "
+            ? "Swift.Optional where Wrapped == "
             : "";
 
     if (generic_info.class_bindings.size()) {
@@ -1446,8 +1446,7 @@ swift::ValueDecl *SwiftASTManipulator::MakeGlobalTypealias(
                            nullptr, &m_source_file);
   swift::Type underlying_type = GetSwiftType(type);
   type_alias_decl->setUnderlyingType(underlying_type);
-  if (underlying_type->hasArchetype())
-    type_alias_decl->markAsDebuggerAlias(true);
+  type_alias_decl->markAsDebuggerAlias(true);
 
   Log *log(lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_EXPRESSIONS));
   if (log) {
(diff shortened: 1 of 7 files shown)


JDevlieghere:
I'm not really familiar with this code, I think Adrian and Davide would be better suited.

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -543,6 +543,13 @@ AddRequiredAliases(Block *block, lldb::StackFrameSP &stack_frame_sp,
   if (!imported_self_type.IsValid())
     return;
 
+  SwiftLanguageRuntime *swift_runtime =
+      stack_frame_sp->GetThread()->GetProcess()->GetSwiftLanguageRuntime();
+  auto *stack_frame = stack_frame_sp.get();
+  imported_self_type =
+      swift_runtime->DoArchetypeBindingForType(*stack_frame,
+                                               imported_self_type);
+
```
adrian-prantl, line 552: I think this is reasonable.
Whether the compiler does the archetype binding for us or we do it for him doesn't make a difference in the end. And the difference between `expr` and `expr -d no-runtime` is one of presenting the result, not really the input expression, at least as far as generic types are concerned.

adrian-prantl: @jimingham may have an opinion on this?

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -557,36 +564,6 @@ AddRequiredAliases(Block *block, lldb::StackFrameSP &stack_frame_sp,
 
   Flags imported_self_type_flags(imported_self_type.GetTypeInfo());
```
adrian-prantl, line 565: This sea of red looks great!

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -474,10 +481,10 @@ static bool GetObjectDescription_ObjectCopy(Process *process, Stream &str,
       process->GetThreadList().GetSelectedThread()->GetSelectedFrame();
   auto *swift_ast_ctx =
       llvm::dyn_cast_or_null<SwiftASTContext>(static_type.GetTypeSystem());
-  if (swift_ast_ctx)
-    static_type =
-        swift_ast_ctx->MapIntoContext(frame_sp,
-                                      static_type.GetOpaqueQualType());
+  if (swift_ast_ctx) {
+    SwiftASTContextLock lock(GetSwiftExeCtx(object));
+    static_type = runtime->DoArchetypeBindingForType(*frame_sp, static_type);
```
adrian-prantl, line 486: This is good for basically the same reason as above. We're asking the object to print its description using the expression evaluator, so somebody has to bind the archetypes and it might as well be us.

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
```diff
@@ -1470,88 +1470,4 @@ swift::ValueDecl *SwiftASTManipulator::MakeGlobalTypealias(
...
-    // result_type = ReplaceInType(result_type,
-    // result_fixup_types.context_alias, result_fixup_types.context_real);
-    if (result_type.getPointer() == result_fixup_types.context_alias) {
-      result_type = result_fixup_types.context_alias->getSinglyDesugaredType();
-    }
-  }
-
-  return result_type;
-}
+}
```
adrian-prantl, line 1472: no newline?

slavapestov:
@swift-ci Please test

slavapestov:
@swift-ci Please test

slavapestov:
@swift-ci Please test

PR #1264, opened 2019-01-31
Adding s390x support in SwiftLanguageRuntime

levivic:
This adds support of s390x on SwiftLanguageRuntime.cpp.


dcci:
@levivic thanks for your contribution! Are you trying to port `swift-lldb ` to your platform?

levivic:
@dcci yeah, however our current main focus is still on `swift`. But since we have been using `swift-lldb`, we contribute to `swift-lldb` when we see issue arising.

dcci:
what test this is fixing?

levivic:
@dcci this doesn't fix any specific test. When i was using `lldb` to debug swift on s390x, there was an crash on lldb which failed at https://github.com/apple/swift-lldb/blob/swift-5.0-branch/source/Target/SwiftLanguageRuntime.cpp#L2671. That was where i spotted the missing support for s390x and provided fix.

dcci:
@swift-ci please test

dcci (approved):

PR #1265, opened 2019-01-31
[SwiftLanguageRuntime] Fix trapping on exclusivity violation.

dcci:
This should also work for the playground, and it's a more accurate
check for the REPL.

<rdar://problem/33820622>


dcci:
@swift-ci build toolchain

dcci:
@swift-ci test and merge

PR #1267, opened 2019-02-01
Fix use of non-existing variable in crashlog.py

shafik:
Summary:
The method find_matching_slice(self) uses uuid_str on one of the paths but the variable does not exist and so this results in a NameError exception if we take that path.

Differential Revision: https://reviews.llvm.org/D57467

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@352772 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit cf4db884f6adf7f1a6901469656a803ae30a0ee7)


shafik:
@swift-ci test and merge

adrian-prantl (approved):

PR #1268, opened 2019-02-01
Run some of the mandatory optimization passes before serialization.

bgogul:
This ensures that the ownership qualifiers are eliminated before we serialize the module. This avoids ownership mismatch errors as in [SR-9805](https://bugs.swift.org/browse/SR-9805).

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
@@ -1938,36 +1938,40 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
   //  - passes like differentation need to see the code before optimizations.
   //  - Some passes may create new functions, but only the functions defined in
   //    the lldb repl line should be serialized.
+  bool done_serialization = false;
   if (swift_ast_ctx->UseSerialization()) {
-    // Run all the passes before differentiation before we serialize.
-    runSILMandatoryOptPreDiffPasses(*sil_module);
-    // Serialize the module now.
-    auto expr_module_dir = swift_ast_ctx->GetReplExprModulesDir();
-    assert(expr_module_dir != nullptr);
-    llvm::SmallString<256> filename(expr_module_dir);
-    std::string module_name;
-    GetNameFromModule(&parsed_expr->module, module_name);
-    llvm::sys::path::append(filename, module_name);
-    llvm::sys::path::replace_extension(filename, ".swiftmodule");
-    // TODO: Check language is swift
-    swift::SerializationOptions serializationOpts;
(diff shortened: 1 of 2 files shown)


source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -1939,6 +1939,9 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
   //  - Some passes may create new functions, but only the functions defined in
   //    the lldb repl line should be serialized.
   if (swift_ast_ctx->UseSerialization()) {
+    // Run all the passes before differentiation before we serialize.
+    runSILMandatoryOptPreDiffPasses(*sil_module);
```
marcrasi: Does it work to run all the sil diagnostic passes before we serialize? That would be more consistent with what the normal compiler does in `FrontendTool.cpp : performCompileStepsPostSILGen()`

bgogul: Unfortunately, that messes up auto-diff. Specifically, mandatory inlining converts the high-level operators like `.+` to builtins, which causes autodiff to not compute correct gradients.

marcrasi: Oh, I see that the comment above the `if` explains it.

I guess that the reason we don't encounter the same problems when using the normal compiler is that we never try to differentiate functions defined in different modules when using the normal compiler? Any time we do cross-module differentiation in the normal compiler, we require that the differentiated function is marked `@differentiable` so that it gets differentiated when the module gets compiled.

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -1959,6 +1962,8 @@ unsigned SwiftExpressionParser::Parse(DiagnosticManager &diagnostic_manager,
                      sil_module.get());
   }
 
+  // TODO(https://bugs.swift.org/browse/SR-9805): We should avoid re-running all
+  // the passses in runSILMandatoryOptPreDiffPasses(...) here.
```
marcrasi: My suggestion in https://github.com/apple/swift/pull/22303 will also resolve this TODO!

marcrasi (changes requested):

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
bgogul: Done!

bgogul:
@marcrasi Addressed your comments, PTAL.

marcrasi (approved):

PR #1269, opened 2019-02-02
add code completion to the SBAPI

marcrasi:
This PR has two things:
* A code completer in SwiftCodeCompletion.cpp. This is a simplified copy of [REPLCodeCompletion.cpp](https://github.com/apple/swift/blob/master/lib/IDE/REPLCodeCompletion.cpp). It is not very good right now, but I plan to experiment with it and iterate on it. I want it separate from REPLCodeCompletion.cpp because REPLCodeCompletion.cpp has a bunch of features that are irrelevant for the jupyter kernel that would slow down my experimentation and iteration.
* A bunch of plumbing to expose this to Python through the SBAPI. I think the plumbing is not quite abstract and indirect as it should be: SBTarget things aren't supposed to be language-specific. But I think I can deal with this later.

### DIFF
--- a/include/lldb/Target/Language.h
+++ b/include/lldb/Target/Language.h
@@ -25,6 +25,8 @@
 #include "lldb/DataFormatters/DumpValueObjectOptions.h"
 #include "lldb/DataFormatters/FormatClasses.h"
 #include "lldb/DataFormatters/StringPrinter.h"
+// SWIFT_ENABLE_TENSORFLOW
+#include "lldb/Target/CompletionResponse.h"
 #include "lldb/lldb-private.h"
 #include "lldb/lldb-public.h"
 
@@ -230,6 +232,10 @@ public:
 
   static void GetDefaultExceptionResolverDescription(bool catch_on,
                                                      bool throw_on, Stream &s);
+
+  // SWIFT_ENABLE_TENSORFLOW
+  virtual CompletionResponse CompleteCode(Target &target,
+                                          std::string entered_code);
 
   // These are accessors for general information about the Languages lldb knows
   // about:
--- a/include/lldb/Target/Language.h
+++ b/include/lldb/Target/Language.h
@@ -235,7 +235,7 @@ public:
 
   // SWIFT_ENABLE_TENSORFLOW
   virtual CompletionResponse CompleteCode(Target &target,
-                                          std::string entered_code);
+                                          const std::string &entered_code);
 
   // These are accessors for general information about the Languages lldb knows
   // about:
--- a/include/lldb/API/LLDB.h
+++ b/include/lldb/API/LLDB.h
@@ -58,6 +58,9 @@
 #include "lldb/API/SBStream.h"
 #include "lldb/API/SBStringList.h"
 #include "lldb/API/SBStructuredData.h"
+// SWIFT_ENABLE_TENSORFLOW
+#include "lldb/API/SBCompletionMatch.h"
+#include "lldb/API/SBCompletionResponse.h"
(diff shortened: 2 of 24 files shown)


jimingham (changes requested):
I think you hinted at this, but this is not structured correctly for lldb.  The completion should be a feature of the Language plugin, and all the public API's should take code and a language (and possibly a SBSymbolContext, since that will effect completion as well.

Then the Swift Language plugin will implement the swift version of this, and for now the other languages will say this is unsupported.

I'm a little hesitant to let this go in structured as it is, because then it will grow more tendrils and that and inertia will make it harder to restructure correctly as time goes on.

jimingham:
I see this is just to the tensor-flow branch.  I don't have any say what you do there, so I'll remove the "changes requested".  But it won't be acceptable to any or the main line branches as is...  Also, the upper side of all these changes need to go through llvm.org to lldb.  But that's for the future...

jimingham (approved):

marcrasi:
I'll update this PR to take a language and dispatch to an implementation based on the language, as you suggest. Maybe I'll add SBSymbolContext later if I discover that I need it.

@jimingham do you have a suggested way to do the dispatch? I could start with a simple switch over the language that calls a language-specific implementation.

marcrasi:
Actually I see this `Language` class with a bunch of virtual methods. I'll try adding a virtual code completion method to it.

jimingham:
Yes, virtual methods on the Language seems like the right way to do it.  

At some point you will need a SymbolContext.  For instance, if the debugger is stopped in a method that defines a local variable foo, then "fo<TAB>" should complete to "foo" among other things.  The only way you would know to do that is because you consulted the list of locals in the current symbol context.  

Even if you don't use the symbol context at first, it would be good to design it into the API's because we don't remove SB API's and I'd rather not end up with two API's, one that takes a SymbolContext and one that doesn't.

marcrasi:
Ok! I made it dispatch using a virtual method on Language, and I also added an SBSymbolContext.

I think I might still iterate on the SBAPI in the tensorflow branch a bit before I end up with a good one that covers everything.

jimingham:
I haven't read through the whole patch yet, but we don't allow inlineable code in SB API .h files.  All actual code should go in the .cpp files.  Otherwise a library building against the SB API's will get code from the version of the SB API it was built against - which may not work with the version of the LLDB library it actually loads.

jimingham:
Also a few comments in the SB headers would not go amiss...

marcrasi:
> I haven't read through the whole patch yet, but we don't allow inlineable code in SB API .h files. All actual code should go in the .cpp files. Otherwise a library building against the SB API's will get code from the version of the SB API it was built against - which may not work with the version of the LLDB library it actually loads.

That makes sense, thanks. Fixed

> Also a few comments in the SB headers would not go amiss...

Also done :)

include/lldb/Target/Language.h
```diff
@@ -231,6 +233,10 @@ class Language : public PluginInterface {
   static void GetDefaultExceptionResolverDescription(bool catch_on,
                                                      bool throw_on, Stream &s);
 
+  // SWIFT_ENABLE_TENSORFLOW
+  virtual CompletionResponse CompleteCode(Target &target,
+                                          std::string entered_code);
```
bgogul: Should this be `const std::string&` instead of `std::string`? If so, there are also other places that take a std::string argument instead of `const std::string&`

bgogul (approved):

include/lldb/Target/Language.h
marcrasi: Done!

PR #1270, opened 2019-02-05
[Swift] Disable round-trip verification of debug types from the expression evaluator

slavapestov:


slavapestov:
https://github.com/apple/swift/pull/22260
@swift-ci Please test

PR #1271, opened 2019-02-06
Relative source map

jimingham:
lldb changed how relative paths in the line table are represented - normalizing them to remove "./" - which broke applying relative target.source-maps.  This patch fixes that by doing the same check that was done in RemapPaths in the function where we actually find the source files.

This patch is slightly changed from the llvm.org version because swift5.0 is at a slightly earlier stage in the FileSpec API's than llvm.org TOT.

### DIFF
--- a/source/Target/PathMappingList.cpp
+++ b/source/Target/PathMappingList.cpp
@@ -205,27 +205,46 @@ bool PathMappingList::ReverseRemapPath(const FileSpec &file, FileSpec &fixed) co
 
 bool PathMappingList::FindFile(const FileSpec &orig_spec,
                                FileSpec &new_spec) const {
-  if (!m_pairs.empty()) {
-    char orig_path[PATH_MAX];
-    const size_t orig_path_len =
-        orig_spec.GetPath(orig_path, sizeof(orig_path));
-    if (orig_path_len > 0) {
-      const_iterator pos, end = m_pairs.end();
-      for (pos = m_pairs.begin(); pos != end; ++pos) {
-        const size_t prefix_len = pos->first.GetLength();
+  if (m_pairs.empty())
+    return false;
+  
+  std::string orig_path = orig_spec.GetPath();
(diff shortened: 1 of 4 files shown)


jimingham:

source/Target/PathMappingList.cpp
```diff
@@ -205,27 +205,46 @@ bool PathMappingList::ReverseRemapPath(const FileSpec &file, FileSpec &fixed) co
...
+  
+  std::string orig_path = orig_spec.GetPath();
+    
+  if (orig_path.empty())
+    return false;
+      
+  bool orig_is_relative = orig_spec.IsRelative();
+
+  const_iterator pos, end = m_pairs.end();
+  for (pos = m_pairs.begin(); pos != end; ++pos) {
```
JDevlieghere, line 219: This could be a range, but looks like this was existing code.

source/Target/PathMappingList.cpp
```diff
@@ -205,27 +205,46 @@ bool PathMappingList::ReverseRemapPath(const FileSpec &file, FileSpec &fixed) co
...
+    llvm::StringRef orig_ref(orig_path);
+    llvm::StringRef prefix_ref = pos->first.GetStringRef();
+    if (orig_ref.size() >= prefix_ref.size()) {
```
JDevlieghere, line 222: You could invert the condition and do an "early continue".

source/Target/PathMappingList.cpp
jimingham: I changed this upstream, but for swift-5.0 this is fine, its just a formal difference.

source/Target/PathMappingList.cpp
jimingham: You guys and your indento-phobia!  I changed this downstream, but for swift-5.0 this version should be fine.  And it's already merged...

PR #1277, opened 2019-02-06
[SwiftASTContext] Stop asking size for types that don't exist in memory.

dcci:
<rdar://problem/47194104>


dcci:
@swift-ci test and merge

PR #1279, opened 2019-02-07
include parameter labels in matches

marcrasi:
Before this change, parameter labels were not included in matches. For example, the completions for
```
import TensorFlow
let t: Tensor<Float> = Tensor(ones: [5])
t.su<TAB>
```
were
```
sum(
sum(
sum(
```
After this change, the matches are
```
sum()
sum(alongAxes: )
sum(squeezingAxes: )
```

### DIFF
--- a/source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
+++ b/source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
@@ -26,54 +26,57 @@ using namespace swift;
 using namespace ide;
 
 static std::string toInsertableString(CodeCompletionResult *Result) {
+  using ChunkKind = CodeCompletionString::Chunk::ChunkKind;
   std::string Str;
-  for (auto C : Result->getCompletionString()->getChunks()) {
-    switch (C.getKind()) {
-    case CodeCompletionString::Chunk::ChunkKind::AccessControlKeyword:
-    case CodeCompletionString::Chunk::ChunkKind::OverrideKeyword:
-    case CodeCompletionString::Chunk::ChunkKind::ThrowsKeyword:
-    case CodeCompletionString::Chunk::ChunkKind::RethrowsKeyword:
-    case CodeCompletionString::Chunk::ChunkKind::DeclAttrKeyword:
-    case CodeCompletionString::Chunk::ChunkKind::DeclIntroducer:
-    case CodeCompletionString::Chunk::ChunkKind::Text:
-    case CodeCompletionString::Chunk::ChunkKind::LeftParen:
-    case CodeCompletionString::Chunk::ChunkKind::RightParen:
-    case CodeCompletionString::Chunk::ChunkKind::LeftBracket:
-    case CodeCompletionString::Chunk::ChunkKind::RightBracket:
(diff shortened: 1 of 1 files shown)


jimingham:
That's definitely better what what was there, but that's not quite how the completions work in lldb (or most of the systems I've seen with completions).  

The way the lldb command line works, when you hit tab, if there's a common substring among all the completions, we just complete to the common substring.  So I would expect:

`t.su<TAB>
`
to complete to 

`t.sum(
`
as a single completion, and that just gets added to the line.  Then a tab at

`t.sum(<TAB>
`
should show the options as your change makes it do.

That's also what most file completer I've seen do.  If you do:

`$ /foo/ba<TAB>
`
that will complete to 

`$ /foo/bar/
`
and not:

```
$ /foo/ba
Available completions:
r/baz
r/lala

```
etc.

That's also nicer when you are typing, because then the "um(" part gets filled in for you, and if you were trying to remember the spelling of "sum" you've gotten what you needed, and if you were trying to disambiguate among parameters, you can type another `<TAB>` and then refine the search by starting to type parameter labels.

Would it be hard to make  your change do that?

marcrasi:
I'm approaching this completer as an API that can be consumed by different UIs, and I think the behavior you described should be the responsibility of the UI.

For example, if we refactor the LLDB REPL to use this completer, then it can check the list for a common prefix and automatically fill that in.

An example of a UI with different behavior: IDE-style completions pop up a menu as you type, even if you don't hit tab. That UI does not immediately fill in common prefixes, and that UI can also use this API.

As I develop this API in the tensorflow branch, the Jupyter UI is the consumer at the top of my mind, so I'm  being a bit biased towards doing things that are good for it. (There are actually two Jupyter UIs that I'm thinking about, and ones of them completes on TAB while the other completes as you type.) But getting this into master and making the LLDB REPL use it too is something that I want to eventually acheive.

jimingham:
It seems reasonable to have the completer have maximalist and minimalist modes.  OTOH, it doesn't seem reasonable to have clients implement either of these modes.  We should do that for them.  That argues for a flag to control the behavior.  That way we will give a consistent experience to everybody who wants to use this.

marcrasi:
A flag does sound pretty good. The completer has more information about the completions than the clients, so it can probably adjust the behavior more easily than the clients can.

I'm going to forge ahead, implementing the specific behavior that I want for Jupyter, without any flags, for now, because I have some specific things that I want to acheive in Jupyter pretty soon.

jimingham:
That seems like a reasonable approach.

jimingham:
Note, I'm not commenting on the actual implementation on purpose.  I didn't write any of this code and haven't studied how the swift completer works, so I am unlikely to catch subtle errors, etc.

marcrasi:
Understood! Thanks for the comments!

jimingham:
BTW, the way the lldb completer does this is that it returns two things, a "longest common substring" and a list of completions.  Since it doesn't take that much effort to construct the "longest common substring" that might be another good way to do this.

bgogul (approved):
Seems reasonable to me.

PR #1282, opened 2019-02-07
 Fix build failure and enable test on Linux.  … 23b6e00

adrian-prantl:
rdar://problem/47861577


adrian-prantl:
@swift-ci test and merge

PR #1283, opened 2019-02-07
[Swift] Get the correct metadata pointer, instead of crashing.

dcci:
SR-9882
<rdar://problem/47884443>

### DIFF
--- a/source/Plugins/Language/Swift/SwiftMetatype.cpp
+++ b/source/Plugins/Language/Swift/SwiftMetatype.cpp
@@ -26,7 +26,7 @@ using namespace lldb_private::formatters::swift;
 
 bool lldb_private::formatters::swift::SwiftMetatype_SummaryProvider(
     ValueObject &valobj, Stream &stream, const TypeSummaryOptions &options) {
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci test and merge

source/Plugins/Language/Swift/SwiftMetatype.cpp
```diff
@@ -26,7 +26,7 @@ using namespace lldb_private::formatters::swift;
 
 bool lldb_private::formatters::swift::SwiftMetatype_SummaryProvider(
     ValueObject &valobj, Stream &stream, const TypeSummaryOptions &options) {
-  lldb::addr_t metadata_ptr = valobj.GetValueAsUnsigned(LLDB_INVALID_ADDRESS);
```
dcci, line 29: As this might not be entirely obvious from the patch, the issue here is that GetValueAsUnsigned() doesn't work for a zero-sized type, e.g. MetaTypes, but we can call `GetPointerValue()` because we always have the location of the value in our hands.

PR #1285, opened 2019-02-09
completion with persistent expression state

marcrasi:
Before this PR, the completer "remembered" decls from previously-evaluated code by concatenating all the previously-evaluated code in front of the code to be completed. The completer then sent this whole thing to the parser and typechecker, and used the parse and typecheck results to provide completions. This caused incorrect behavior with redefinitions, because the parse&typecheck steps did not handle redefinitions correctly.

This PR completely changes the approach. Now, the completer works like this:
0. Creates two `SourceFile`s named `PreviousDeclsFile` and `EnteredCodeFile` in the same module.
1. Put all the decls that are in `SwiftPersistentExpressionState` into `PreviousDeclsFile`.
2. Put all the imports that are in `SwiftPersistentExpressionState` into a `EnteredCodeFile`.
3. Parse just the current snippet of code, to find out if it declares anything. If it does declare something that was already in `SwiftPersistentExpressionState`, remove that decl from `PreviousDeclFile`.
4. Put the just current snippet of code into `EnteredCodeFile`.
5. Do parsing, typechecking, and completion on `EnteredCodeFile`. (Since previous decls are in `PreviousDeclFile`, which is in the same module, it is able to see them).

This approach handles redefinitions correctly, as far as I can tell with my initial tests.

This PR also adds some completion tests.

### DIFF
--- a/source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
+++ b/source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
@@ -195,8 +195,166 @@ doCodeCompletion(SourceFile &SF, StringRef EnteredCode,
   return BufferID;
 }
 
-CompletionResponse SwiftCompleteCode(SourceFile &SF, StringRef EnteredCode) {
-  ASTContext &Ctx = SF.getASTContext();
+static void AddSourceFile(ASTContext &Ctx, ModuleDecl *Module,
+                          SourceFileKind Kind) {
+  Optional<unsigned> BufferID;
+  SourceFile *SF = new (Ctx) SourceFile(
+      *Module, Kind, BufferID, SourceFile::ImplicitModuleImportKind::Stdlib,
+      /*Keep tokens*/ false);
+  SF->ASTStage = SourceFile::TypeChecked;
+  Module->addFile(*SF);
+}
+
+static SourceFile *GetSingleSourceFile(ModuleDecl *Module,
+                                       SourceFileKind Kind) {
+  SourceFile *Result = nullptr;
+  for (auto *File : Module->getFiles()) {
+    auto *SF = dyn_cast<SourceFile>(File);
+    if (!SF)
+      continue;
+    if (SF->Kind != Kind)
+      continue;
+    assert(!Result && "multiple source files of the requested kind");
+    Result = SF;
+  }
+  return Result;
+}
+
+CompletionResponse
+SwiftCompleteCode(SwiftASTContext &SwiftCtx,
+                  SwiftPersistentExpressionState &PersistentExpressionState,
+                  StringRef EnteredCode) {
+  Status Error;
+  ASTContext &Ctx = *SwiftCtx.GetASTContext();
+
+  // == Prepare a module and source files ==
+
+  // Get or create the module that we do completions in.
+  static ConstString CompletionsModuleName("completions");
+  ModuleDecl *CompletionsModule =
+      SwiftCtx.GetModule(CompletionsModuleName, Error);
+  if (!CompletionsModule) {
+    CompletionsModule = SwiftCtx.CreateModule(CompletionsModuleName, Error);
+    if (!CompletionsModule)
+      return CompletionResponse::error("could not make completions module");
+
+    // This file accumulates all of the "hand imports" (imports that the user
+    // made in previous executions). We also put the current entered code in
+    // this file.
+    AddSourceFile(Ctx, CompletionsModule, SourceFileKind::REPL);
+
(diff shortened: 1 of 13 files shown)


source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
```diff
@@ -192,8 +192,166 @@ doCodeCompletion(SourceFile &SF, StringRef EnteredCode,
...
+  {
+    std::vector<Decl *> PersistentDecls;
+    PersistentExpressionState.GetAllDecls(PersistentDecls);
+    PreviousDeclsFile->Decls.clear();
+    for (auto *PersistentDecl : PersistentDecls)
+      PreviousDeclsFile->Decls.push_back(PersistentDecl);
+    PreviousDeclsFile->clearLookupCache();
+  }
+
+  // `PreviousDeclsFile` might now contain decls that a re-defined in
```
bgogul, line 300: This approach might not work correctly with extensions. e.g., if you redefine a nested type in an extension, you might still have both the definitions in the auto complete. 

I think a proper fix for https://bugs.swift.org/browse/TF-10 shoud take care of the issue automatically.

source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
```diff
@@ -192,8 +192,166 @@ doCodeCompletion(SourceFile &SF, StringRef EnteredCode,
...
+    for (auto it = EnteredCodeFile->Decls.begin() + OriginalDeclCount;
+         it != EnteredCodeFile->Decls.end(); it++)
+      if (auto *NewValueDecl = dyn_cast<ValueDecl>(*it))
+        NewDecls.FindAndConstruct(NewValueDecl->getBaseName().getIdentifier())
+            .second.push_back(NewValueDecl);
+    EnteredCodeFile->Decls.resize(OriginalDeclCount);
+    DelayedDiags.abort();
+
+    // Subtract `NewDecls` from the decls in `PreviousDeclsFile`.
+    auto ContainedInNewDecls = [&](Decl *OldDecl) -> bool {
```
bgogul, line 329: Can't you use the swift_persistent_state->FindMatchingDecls / GetSwiftPersistentDecls to check if a decl is already there?

source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
```diff
@@ -192,8 +192,166 @@ doCodeCompletion(SourceFile &SF, StringRef EnteredCode,
...
+  }
+
+  // This file accumulates all of the "hand imports" (imports that the user
+  // made in previous executions). We also put the current entered code in
+  // this file.
+  SourceFile *EnteredCodeFile =
+      GetSingleSourceFile(CompletionsModule, SourceFileKind::REPL);
+  assert(EnteredCodeFile);
+
+  // Accumulate new hand imports into the file.
```
bgogul, line 257: I wonder if you can simply reuse SwiftExpressionParser::AutoImport?

bgogul (commented):
I think the approach sounds reasonable. However, it seems like we might be able to reuse some existing code for some of the functionality here.

source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
marcrasi: Indeed, it behaves weirdly with extensions. I have added some tests documenting the current behavior. We'll defer actually fixing it until later.

source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
marcrasi: We discussed this a bit in person.

In summary, this sounds good to do eventually. A current blocker is that `swift_persistent_state->FindMatchingDecls / GetSwiftPersistentDecls` requires the AST to be typechecked, but I need to do my deduping before typechecking. I think it might be possible to modify `swift_persistent_state->FindMatchingDecls / GetSwiftPersistentDecls` to work on non-typechecked ASTs, but I will do that later in a separate PR (and probably into master because it's a change that would also make sense in master).

source/Plugins/Language/Swift/SwiftCodeCompletion.cpp
marcrasi: Also discussed a bit in person.

It sounds like a good goal to dedupe this with SwiftExpressionParser::AutoImport eventually, but it'll be a lot of work because SwiftExpressionParser::AutoImport does a lot of things and deals with a lot of state that might not be good for the completer to touch.

So we'll defer this until later.

PR #1286, opened 2019-02-11
Use RemoteAST in GetDynamicTypeAndAddress()

slavapestov:
There was a lot of redundant logic here that predates RemoteAST.

I started off by trying to simplify debug info emission in the Swift side by only emitting debug values for top-level generic parameter types and not member types; in lldb we should be able to look up member types just by calling `Type::subst()`. Unfortunately, this broke test cases where we were getting the dynamic type of values whose type is a member type. Turns out this was handled as part of a special "generic parameter or member type" code path, which would map the type to a concrete type, and then get the dynamic type of the concrete type.

This "get dynamic type of concrete type resulting from concretizing a generic parameter type" code path was totally different from the "get dynamic type of concrete type" code path. The former used RemoteAST, the latter did not. Merging these two code paths is what this PR is all about.

The non-RemoteAST code path handled a few cases that the RemoteAST code path did not, such as values materialized in host memory -- this is how results come back from the expression evaluator, which wants to "freeze" them so that they're available later. It also handled imported types and some other edge cases.

Now that RemoteAST on the Swift side is more capable, all it took was adding some support to lldb's MemoryReader implementation for reading memory from a specified local buffer to use RemoteAST for all the other cases as well.

Fixes <rdar://40706055> (but really a whole lot more than that).

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -2240,38 +2240,6 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_GenericTypeParam(
   return true;
 }
 
-bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Tuple(
-    ValueObject &in_value, SwiftASTContext &scratch_ctx,
-    lldb::DynamicValueType use_dynamic, TypeAndOrName &class_type_or_name,
-    Address &address) {
-  Status error;
-  std::vector<CompilerType> dyn_types;
-  for (size_t idx = 0; idx < in_value.GetNumChildren(); idx++) {
-    ValueObjectSP child_sp(in_value.GetChildAtIndex(idx, true));
-    TypeAndOrName type_and_or_name;
-    Address address;
-    Value::ValueType value_type;
-    CompilerType child_type;
-    if (!GetDynamicTypeAndAddress(*child_sp.get(), use_dynamic,
-                                  type_and_or_name, address, value_type))
-      child_type = child_sp->GetCompilerType();
-    else
-      child_type = type_and_or_name.GetCompilerType();
-    dyn_types.push_back(scratch_ctx.ImportType(child_type, error));
-  }
-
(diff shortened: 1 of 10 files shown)


source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1729,380 +1734,88 @@ bool SwiftLanguageRuntime::IsValidErrorValue(
...
-  auto ptr_size = m_process->GetAddressByteSize();
+  if (in_value.GetValueType() == eValueTypeConstResult) {
+    if (log)
+      log->Printf("existential value is a const result");
+
+    // Horrible hack -- we have a locally materialized value, so let
+    // our MemoryReader know where it is so that RemoteAST can read
+    // from it. Also, there's something going on with Error values
+    // I don't quite understand.
+    is_error = GetSwiftType(protocol_type)
```
dcci: This is the precise reason I didn't switch to RemoteAST this codepath after adding the support in the compiler. Apparently the expression parser does something weird with materializing/dematerializing values, @jimingham has a rdar assigned to him about this.

slavapestov: For anyone else internal the radar is rdar://38194957. While it would be nice to fix this it's not a pre-requisite since we can teach the RemoteAST memory reader to read values from known local ranges of memory too. It would clean up the code here if we didn't have to worry about that, though.

slavapestov:
@swift-ci Please test

slavapestov:
@swift-ci Please test

slavapestov:
@dcci I figured out the problem with Error existentials. It was in SwiftLanguageRuntime::CalculateErrorValue(). There was also a RemoteAST bug on Linux with error values. This should now be ready to merge.

slavapestov:
https://github.com/apple/swift/pull/22975
@swift-ci Please test

dcci (approved):
LGTM. Thanks.

slavapestov:
@swift-ci Please test

PR #1290, opened 2019-02-12
[SwiftExpressionParser] Add tests for the fixed closure behaviour.

dcci:
<rdar://problem/47982630>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/main.swift
@@ -8,7 +8,11 @@ func main() -> Int {
               //%self.expect('frame var $1', substrs=['foo'])
   )
 
-  return 0
+  var tinky = [1,2].map({$0 * 2})
+
+  return 0 //%self.expect('expr tinky.map({$0 * 2})', substrs=['[0] = 2', '[1] = 4'])
+           //%self.expect('expr [2,4].map({$0 * 2})', substrs=['[0] = 4', '[1] = 8'])
+           //%self.expect('expr $0', substrs=['unresolved identifier \'$0\''], error=True)
 }
 
 _ = main()
--- a/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/main.swift
@@ -8,9 +8,11 @@ func main() -> Int {
               //%self.expect('frame var $1', substrs=['foo'])
   )
 
-  var tinky = [1,2].map({$0 * 2})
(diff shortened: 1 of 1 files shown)


dcci:
https://github.com/apple/swift/pull/22552

@swift-ci please test

dcci:
[apple/swift#22552](https://github.com/apple/swift/pull/22552)

@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/closure_shortcuts/main.swift
```diff
@@ -8,7 +8,9 @@ func main() -> Int {
               //%self.expect('frame var $1', substrs=['foo'])
   )
 
-  return 0
+  var tinky = [1,2].map({$0 * 2}) //%self.expect('expr [1,2].map({$0 * 2})', substrs=['[0] = 2', '[1] = 4'])
```
jrose-apple: Might be better to break this out into multiple lines, so you can be sure you're testing this from inside the closure. Or, actually, this should just be part of the previous test, no?

dcci: honestly, this is probably better rewritten as 

```
var tinky = [1,2]
//%self.expect('expr tinky.map({$0 * 2}, substr=[...]
```

jimingham: I thought Davide was being clever and getting two test for the price of one.  After all, the breakpoint on the line will actually get two locations, one in the main function, and one in the closure.  So as written the test will actually make sure we can use $0 in a being defined closure both in a closure context and out of it.  Since the two had different failure modes, it seems good to include both tests.

jrose-apple: I guess I'd say that's *too* clever! If we ever break the "breakpoint on a line gets both locations" behavior (unlikely, I guess), we might not notice.

jimingham: Yeah, if you want to make it more explicit, then I think you would have to write:

var tinky = [1,2].map(
{$0 * 2}
)

To make sure each breakpoint takes in the appropriate context.  But I do think it is worthwhile checking those two cases as they broke in different ways.

dcci:
We should have all cases covered between `reversedNames` and `tinky`.

jrose-apple:
Now you're missing the test of doing the `map` inside a closure.

dcci:
oh, I see now.

dcci:
[apple/swift#22552](https://github.com/apple/swift/pull/22552)

@swift-ci please test

jrose-apple (approved):

PR #1291, opened 2019-02-13
[CMake] Link against host swiftReflection instead of target swiftRefl…

bulbazord:
…ection

I'm doing this because I think we should be linking against the swiftReflection library built for the host. The other swift libraries being linked against are all host libraries, so I don't think swiftReflection is a special case.

### DIFF
--- a/source/Target/CMakeLists.txt
+++ b/source/Target/CMakeLists.txt
@@ -1,50 +1,5 @@
 include_directories(../Plugins/Process/Utility)
 
-set(REFLECTION_LIB "swiftReflection")
-
-if(CMAKE_SYSTEM_NAME MATCHES "Linux")
-    # CMAKE_SYSTEM_PROCESSOR is equivalent to `uname -p` on Linux.
-    if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
-      set(ARCH "x86_64")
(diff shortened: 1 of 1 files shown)


bulbazord:
cc @compnerd @dcci @weliveindetail

weliveindetail (changes requested):
While I'm not familiar with the reasoning behind this code, looking at the history shows that https://github.com/apple/swift-lldb/commit/cb495cf645d3b22083c7b5df4e8797adc15e2354 did this exact same change some months ago and https://github.com/apple/swift-lldb/commit/209bc1fb8ecaeb34c13a4047f2f78a79e9dcde20 reverted it.

Thus, swiftReflection seems to be some kind of special case and we should understand the details before landing this.

fredriss:
Hmmm, I don't remember reverting this, thus I don't remember the reasons and my commit message was rather terse. I suppose it broke something in LLDB. We did change the Reflection build in the meantime to unconditionally build a static host library. The change looks reasonable to me.

weliveindetail (approved):
Ok, let's give it a try then.

compnerd:
@swift-ci please test

bulbazord:
It seems like the reverts were to match whatever was on Swift-5.0's branch, which probably meant that we were not building Reflection for the Host there. I don't think that's the case anymore though, so I think this should be okay.

PR #1296, opened 2019-02-14
"LSP Snippets" option for completion API

marcrasi:
This adds an "LSP Snippets" option to the completion API, documented in SBCompletionOptions.i.

Basically, these allow us to insert placeholders for arguments when the user selects a function from the autocomplete menu. Monaco-based editors automatically highlight the placeholders and let the user tab between them, so that it's easy for them to fill in all the arguments. It looks like this:
lsp_snippet_placeholders


bgogul (approved):
Nice!

marcrasi:
fyi, I'm probably not going to get around to fully releasing this for the next few weeks, so I will just leave this PR open here until I have time.

PR #1299, opened 2019-02-15
[Swift] Rework import of module `repl` for completion.

dcci:


source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
```diff
@@ -549,9 +549,13 @@ int SwiftREPL::CompleteCode(const std::string &current_code,
...
     static ConstString g_repl_module_name("repl");
-    swift::ModuleDecl *repl_module =
+
+    swift::ModuleDecl *repl_module = nullptr;
+    if (m_swift_repl_initialized)
+      repl_module =
         swift_ast->GetModule(g_repl_module_name, error);
-    if (repl_module == NULL) {
+    if (repl_module == nullptr) {
+      swift::ModuleDecl *repl_module = nullptr;
```
adrian-prantl: Isn't that redundant?

source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
```diff
@@ -548,19 +548,22 @@ int SwiftREPL::CompleteCode(const std::string &current_code,
...
-    swift::ModuleDecl *repl_module =
-        swift_ast->GetModule(repl_module_info, error);
-    if (!repl_module) {
-      repl_module = swift_ast->CreateModule(repl_module_info, error);
+    static ConstString g_repl_module_name("repl");
+
+    swift::ModuleDecl *repl_module = nullptr;
+    if (m_swift_repl_initialized)
+      repl_module =
+        swift_ast->GetModule(g_repl_module_name, error);
```
adrian-prantl: Are LHS and RHS from two different branches?

source/Plugins/ExpressionParser/Swift/SwiftREPL.h
```diff
@@ -80,6 +80,7 @@ class SwiftREPL : public REPL {
 
 private:
   lldb::SwiftASTContextSP m_swift_ast_sp;
+  bool m_swift_repl_initialized = false;
```
marcrasi: This variable name is a bit confusing because it sounds like it keeps track of whether the whole repl is initialized. Maybe something involving the word "completion" would be better. Like `m_completion_module_initialized`.

(I had the same confusion with `g_repl_module_name` and also it's string value "repl" when I initially came across this code, so it might be good to change them too, to something like `static ConstString g_completion_module_name("completion")`.

marcrasi (approved):

source/Plugins/ExpressionParser/Swift/SwiftREPL.h
dcci: I agree, let me do this.

source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
dcci: This should be rebased now.

dcci:
@swift-ci test and merge

PR #1301, opened 2019-02-15
[SwiftFormatter] Fix SIMD formatter to work with expr.

dcci:
`expr` doesn't have the memory available so we can't really
call GetAddressOf() in the formatter. While here, simplify it a bit.

<rdar://problem/46330565>


dcci:
@swift-ci test and merge

PR #1302, opened 2019-02-15
Make RemapClangImporterOptions more robust.

adrian-prantl:
Due to a bug absolute path remapping would be applied to -iquote
*options* rather than their arguments. The fixed code only whitelists
a few search path options that it actually understands.

rdar://problem/48078809

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1402,6 +1402,30 @@ bool HasSwiftModules(Module &module) {
 }
 
 namespace {
+
+/// Calls arg.consume_front(<options>) and returns true on success.
+/// \param prefix contains the consumed prefix.
+bool ConsumeIncludeOption(StringRef &arg, StringRef &prefix) {
+  static StringRef options[] = {"-I",
+                                "-F",
+                                "-iquote",
+                                "-idirafter",
+                                "-iframeworkwithsysroot",
+                                "-iframework",
+                                "-iprefix",
+                                "-iwithprefixbefore",
+                                "-iwithprefix",
+                                "-isystemafter",
+                                "-isystem",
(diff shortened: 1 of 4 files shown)


adrian-prantl:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1468,18 +1468,30 @@ void SwiftASTContext::RemapClangImporterOptions(
...
   }
+  bool remap_next = false;
   for (auto &arg_string : options.ExtraArgs) {
     StringRef prefix;
     StringRef arg = arg_string;
-    if (arg.consume_front("-I"))
+
+    if (!remap_next) {
+      /* previous argument was -iquote */;
+    } else if (arg == "-I" || arg == "-F" || arg == "-iquote") {
```
allevato: Clang allows `-iquote` to be used in joined form as well (e.g., `-iquotefoo/bar` is legal): https://github.com/llvm/llvm-project/blob/fcbec02/clang/include/clang/Driver/Options.td#L1951-L1952

Does that case need to be handled here as well? It's ugly, but we actually use it in our Swift Bazel rules to remove some excess `<space>-Xcc<space>`s in our command lines.

adrian-prantl: Done.

adrian-prantl:
@swift-ci test

DavidGoldman:
@adrian-prantl Is it possible to merge this into the Swift 5.0 branch?

adrian-prantl:
@DavidGoldman it is! -> https://github.com/apple/swift-lldb/pull/1305

DavidGoldman:
Thanks, is it in an Xcode 10.2 beta?

PR #1305, opened 2019-02-16
Make RemapClangImporterOptions more robust.

adrian-prantl:
Due to a bug absolute path remapping would be applied to -iquote
*options* rather than their arguments. The fixed code only whitelists
a few search path options that it actually understands.

This patch also adds support for -iquote/some/path.

rdar://problem/48078809


adrian-prantl:
@swift-ci test

PR #1317, opened 2019-02-21
Cherry-pick "Merge target triple into module triple when constructing module from memory"

bulbazord:
I'm cherry-picking this change because I'm trying to debug swift code on android and ran into issues when doing expression evaluation. Specifically, constructing `SwiftASTContext`s presented a problem because the modules being used to construct the SwiftASTContext had incomplete triples (e.g `aarch64---`). swift-lldb makes the assumption that if the OS is not set in the triple, then it should use the host OS (when not an apple platorm). That doesn't work for remote debugging, so let's try to make sure triples are constructed as accurately as possible.


bulbazord:
cc @compnerd @adrian-prantl

compnerd:
@swift-ci please test

adrian-prantl (approved):

PR #1320, opened 2019-02-22
TF-223: When a type does not conform, explain why

marcrasi:
The LLDB diagnostic logic hides diagnostics pointing to things that the user did not enter. For example, if you do `struct Foo : Sequence {}`, then you should get this:
```
error: repl.swift:1:8: error: type 'Foo' does not conform to protocol 'Sequence'
struct Foo : Sequence {}
       ^

Swift.Sequence:3:20: note: unable to infer associated type 'Iterator' for protocol 'Sequence'
    associatedtype Iterator : IteratorProtocol
                   ^

Swift.Sequence:2:40: note: candidate would match and infer 'Iterator' = 'Foo' if 'Foo' conformed to 'IteratorProtocol'
    @inlinable public __consuming func makeIterator() -> Self
```
but LLDB hides the two notes because they don't point at the user's expression.

This PR fixes that.

I'm not sure if there are any situations where this fix breaks some desired behavior. I'll create a PR into master soon, and I'll ask in that PR whether there are such situations.

Fixes https://bugs.swift.org/browse/TF-223

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -2807,55 +2807,61 @@ public:
       const DiagnosticSeverity severity = SeverityForKind(diagnostic.kind);
       const DiagnosticOrigin origin = eDiagnosticOriginSwift;
 
-      if (first_line > 0 && bufferID != UINT32_MAX &&
-          diagnostic.bufferID == bufferID && !diagnostic.bufferName.empty()) {
-        // Make sure the error line is in range
-        if (diagnostic.line >= first_line && diagnostic.line <= last_line) {
-          // Need to remap the error/warning to a different line
-          StreamString match;
-          match.Printf("%s:%u:", diagnostic.bufferName.str().c_str(),
-                       diagnostic.line);
-          const size_t match_len = match.GetString().size();
-          size_t match_pos = diagnostic.description.find(match.GetString());
-          if (match_pos != std::string::npos) {
-            // We have some <file>:<line>:" instances that need to be updated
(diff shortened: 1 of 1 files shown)


source/Symbol/SwiftASTContext.cpp
```diff
@@ -2807,55 +2807,61 @@ class StoringDiagnosticConsumer : public swift::DiagnosticConsumer {
...
+          fixed_description.Printf("%s:%u:",
+                                   diagnostic.bufferName.str().c_str(),
+                                   diagnostic.line - first_line +
+                                       line_offset + 1);
+          start_pos = match_pos + match_len;
+          match_pos =
+              diagnostic.description.find(match.GetString(), start_pos);
+        } while (match_pos != std::string::npos);
+
+        // Append any last remainging text
```
bgogul, line 2846: typo: remainging -> remaining

(The old code has the typo too.)

bgogul (approved):

jimingham:
The problem is, for instance, every time an expression fails to parse we get a bunch of warnings that have nothing to do with the compile failure (e.g. that we didn't use the variable we use to capture errors.  Those are just confusing, so we suppress them.

marcrasi:
I think that this change handles that case, because I have made it still exclude errors from the expression's buffer that fall outside of `[first_line, last_line]`. So if those confusing warnings all appear in the same buffer as the expression, they should still be excluded.

jimingham:
That should be fine, then.  Avoiding printing errors/warnings for the implementation details of the expression parser was the whole point of this exercise.

PR #1325, opened 2019-02-26
XPASSES are now reported. Fix all of them on Linux.

dcci:
Same as stable, 5.1 version.


dcci:
@swift-ci test and merge

PR #1329, opened 2019-02-27
 Show original Archetype names in backtraces.

adrian-prantl:
Since this information is now available in DWARF, LLDB can display the
correct names for archetypes when printing backtraces.

<rdar://problem/48259889>

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -85,6 +85,7 @@ using namespace lldb;
 using namespace lldb_private;
 
 static constexpr std::chrono::seconds g_po_function_timeout(15);
+static const char *g_dollar_tau_underscore = u8"$\u03C4_";
 
 namespace lldb_private {
 swift::Type GetSwiftType(void *opaque_ptr) {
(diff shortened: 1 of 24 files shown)


adrian-prantl:
test with https://github.com/apple/swift/pull/22954
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/22954
@swift-ci test

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -689,26 +691,54 @@ bool SwiftLanguageRuntime::IsSwiftMangledName(const char *name) {
...
+  llvm::DenseMap<std::pair<uint64_t, uint64_t>, StringRef> dict;
+  if (simplified)
+    options = swift::Demangle::DemangleOptions::SimplifiedUIDemangleOptions();
+
+  if (sc) {
+    bool can_create = true;
+    // This terrifying cast avoids having too many differences with llvm.org.
+    if (Block *block = const_cast<SymbolContext *>(sc)->GetFunctionBlock())
+      if (VariableListSP var_list = block->GetBlockVariableList(can_create))
+        for (unsigned i = 0; i < var_list->GetSize(); ++i) {
```
dcci: Instead of having this lexing process here, can we move this logic to `swift::Demangle`, and maybe add some unit tests for it?

dcci (changes requested):
See comment above.

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: I don't think that makes sense. That format is debug-info-specific and has nothing to do with (de)mangling. It's the naming format we use for the type metadata variables in DWARF.

dcci: If you feel lldb is the right place where this function should live, I think we should at least factor this out so we can test it in isolation :)

adrian-prantl: I factored out common code between SwiftASTContext and SwiftLanguageRuntime into a new `SwiftLanguageRuntime::GetArchetypeNamesForFunction()`.

dcci (approved):
This is much better.

adrian-prantl:
test with https://github.com/apple/swift/pull/22954
@swift-ci test

adrian-prantl:
Works for me. I'll open a separate PR to rename everything.

adrian-prantl:
test with https://github.com/apple/swift/pull/22954
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/22954
@swift-ci test

adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

PR #1331, opened 2019-02-28
[Swift] Add a setting to control the Swift module loading mode

nathawes:
This adds a new setting that controls the Swift module loading mode, equivalent to the SWIFT_FORCE_MODULE_LOADING environment variable supported by the Swift compiler:

- `settings set symbols.swift-module-loading-mode only-parseable`
Only load swift modules via .swiftinterface files
- `settings set symbols.swift-module-loading-mode only-serialized`
Only load swift modules via .swiftmodule files
- `settings set symbols.swift-module-loading-mode prefer-parseable`
Try to load via a .swiftinterface, but fall back to the .swiftmodule if that fails (e.g. if it doesn't exist)
- `settings set symbols.swift-module-loading-mode prefer-serialized`
Try to load via a .swiftmodule, but fall back to the .swiftinterface if that fails (e.g. if it doesn't exist). This is the default.

This patch also updates lldb's `--repl` mode to respect the `-O` option so that it's possible to change this setting before the Swift module loader is initialized when using the REPL.

### DIFF
--- a/lit/Driver/TestRepl.test
+++ b/lit/Driver/TestRepl.test
@@ -1,0 +1,7 @@
+# RUN: echo ':quit' | %lldb -x --repl -O 'expr 42' -S %S/Inputs/Print2.in -o 'expr 999999' -s %s 2>&1 | FileCheck %s
+# CHECK: {{w}}arning: commands specified to run after file load (via -o or -s) are ignored in REPL mode
+# CHECK: (int) $0 = 42
+# CHECK: (int) $1 = 2
+# CHECK-NOT: 999999
+
+expr 999999
--- a/lit/SwiftREPL/SwiftInterfaceForceModuleLoadMode.test
+++ b/lit/SwiftREPL/SwiftInterfaceForceModuleLoadMode.test
@@ -1,0 +1,87 @@
+// This test checks that the value of the symbols.swift-module-loading-mode
+// setting is respected when loading Swift modules.
+
+// RUN: rm -rf %t && mkdir %t
+
+// Setup generates a dylib and .swiftinterface for A.swift as is, but generates
+// the .swiftmodule from a modified version a A.swift where the 'FromInterface'
+// type has been renamed to 'FromSerialized'. This means that the 'testValue'
+// variable will have a different type depending on whether the module was
+// loaded via the .swiftinterface or .swiftmodule.
+
+// RUN: mkdir %t/mcp
+// RUN: mkdir %t/lib
+// RUN: cp %S/Inputs/A.swift %t/AA.swift
+// RUN: %target-swiftc -module-name AA -emit-parseable-module-interface-path %t/lib/AA.swiftinterface -emit-library -o %t/lib/libAA%target-shared-library-suffix %t/AA.swift
+// RUN: sed -e 's/FromInterface/FromSerialized/g' %t/AA.swift | %target-swiftc -module-name AA -emit-module -o %t/lib/AA.swiftmodule -
+// RUN: rm %t/AA.swift
+
+
+// This is the input provided to lldb. It intentionally assigns 'testValue' to
+// a variable of the incorrect type so we can tell via the produced diagnostic
+// whether the .swiftinterface or .swiftmodule was used.
+
+import AA
+
+let x: OtherType = testValue
+// NOT-LOADED: use of undeclared type 'OtherType'
+// FROM-INTERFACE: cannot convert value of type 'FromInterface' to specified type 'OtherType'
+// FROM-SERIALIZED: cannot convert value of type 'FromSerialized' to specified type 'OtherType'
+// INVALID: error: invalid enumeration value '{{.*}}', valid values are: {{.*}}
+
+
+// 1. Neither .swiftinterface or .swiftmodule present
+//
+// RUN: mv %t/lib %t/lib-backup
+// RUN: %lldb --repl="-I%t/lib -L%t/lib -lAA" -O "settings set symbols.clang-modules-cache-path %t/mcp" \
+// RUN:   -O "settings set symbols.swift-module-loading-mode prefer-parseable" < %s 2>&1 | FileCheck %s -check-prefix=NOT-LOADED
+// RUN: %lldb --repl="-I%t/lib -L%t/lib -lAA" -O "settings set symbols.clang-modules-cache-path %t/mcp" \
+// RUN:   -O "settings set symbols.swift-module-loading-mode prefer-serialized" < %s 2>&1 | FileCheck %s -check-prefix=NOT-LOADED
+// RUN: %lldb --repl="-I%t/lib -L%t/lib -lAA" -O "settings set symbols.clang-modules-cache-path %t/mcp" \
+// RUN:   -O "settings set symbols.swift-module-loading-mode only-parseable" < %s 2>&1 | FileCheck %s -check-prefix=NOT-LOADED
+// RUN: %lldb --repl="-I%t/lib -L%t/lib -lAA" -O "settings set symbols.clang-modules-cache-path %t/mcp" \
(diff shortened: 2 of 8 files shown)


nathawes:
@swift-ci test

nathawes:
@swift-ci test

nathawes:
@swift-ci test

nathawes:
@swift-ci test

harlanhaskins (approved):

source/Symbol/SwiftASTContext.cpp
```diff
@@ -2992,20 +2993,44 @@ swift::ASTContext *SwiftASTContext::GetASTContext() {
...
-      swift::ParseableInterfaceModuleLoader::create(
-        *m_ast_context_ap, ModuleCachePath, PrebuiltModuleCachePath,
-        /*tracker=*/nullptr, swift::ModuleLoadingMode::PreferSerialized));
 
-    if (parseable_module_loader_ap) {
-      m_parseable_module_loader =
-      (swift::ParseableInterfaceModuleLoader *)parseable_module_loader_ap.get();
-      m_ast_context_ap->addModuleLoader(std::move(parseable_module_loader_ap));
+    auto MLM = swift::ModuleLoadingMode::PreferSerialized;
+    if (auto forceModuleLoadingMode =
```
jrose-apple: Nitpick: this should use the LLDB variable naming style, even though it's mostly copied from Swift. (Alternately, we could wrap this up in a helper function that LLDB can call.)

lit/SwiftREPL/SwiftInterfaceForceModuleLoadMode.test
```diff
@@ -0,0 +1,65 @@
...
+// NOT-LOADED: use of undeclared type 'OtherType'
+// FROM-INTERFACE: cannot convert value of type 'FromInterface' to specified type 'OtherType'
+// FROM-SERIALIZED: cannot convert value of type 'FromSerialized' to specified type 'OtherType'
+// INVALID: error: unknown value for SWIFT_FORCE_MODULE_LOADING variable: '{{.*}}'
+}
+
+// 1. Neither .swiftinterface or .swiftmodule present
+//
+// RUN: mv %t/lib %t/lib-backup
+// RUN: env SWIFT_FORCE_MODULE_LOADING=prefer-parseable %lldb --repl="-I%t/lib -L%t/lib -lAA -enable-parseable-module-interface" < %s 2>&1 | FileCheck %s -check-prefix=NOT-LOADED
```
jrose-apple: I don't think you need the `-enable-parseable-module-interface` here; the environment variable implies it for the compiler and we're just always enabling it for LLDB.

tools/driver/Driver.cpp
```diff
@@ -662,100 +662,101 @@ int Driver::MainLoop() {
   bool handle_events = true;
   bool spawn_thread = false;
 
-  if (m_option_data.m_repl) {
```
nathawes: The change here is just always taking the else branch, and instead checking if we're in REPL mode just before where we were previously calling m_debugger.RunCommandInterpreter(...) and calling m_debugger.RunREPL(...) instead. The rest is due to indentation change.

This was needed for --repl mode to still run the commands passed via -O (and written to commands_stream in the code above this line)

nathawes:
@swift-ci test

nathawes:
@swift-ci test

nathawes:
@swift-ci test

source/Core/ModuleList.cpp
```diff
@@ -85,12 +100,17 @@ static constexpr PropertyDefinition g_properties[] = {
          "when debugging Swift code"},
     {"clang-modules-cache-path", OptionValue::eTypeFileSpec, true, 0, nullptr,
      {},
-     "The path to the clang modules cache directory (-fmodules-cache-path)."}};
+     "The path to the clang modules cache directory (-fmodules-cache-path)."},
+    {"swift-module-loading-mode", OptionValue::eTypeEnum, false,
+     eSwiftModuleLoadingModePreferSerialized, nullptr,
+     OptionEnumValues(g_swift_module_loading_mode_enums),
+     "The module loading mode to use when loading modules for Swift."}};
```
jrose-apple, line 107: Is there a way to mark that this is only meant for testing?

tools/driver/Driver.cpp
```diff
@@ -612,47 +612,51 @@ int Driver::MainLoop() {
...
-                             EscapeString(m_option_data.m_args[0]).c_str());
-
-    if (!m_option_data.m_core_file.empty()) {
-      commands_stream.Printf(" --core %s",
-                             EscapeString(m_option_data.m_core_file).c_str());
-    }
-    commands_stream.Printf("\n");
+  // If we're not in --repl mode, add the commands to process the file
+  // arguments, and the commands specified to run afterwards.
+  if (!m_option_data.m_repl) {
```
jrose-apple, line 618: Definitely want to make sure an LLDB person signs off on this.

adrian-prantl: @jimingham I wanted to get your opinon on this, too. Currently`lldb --repl` always ignores `-O` options, but for testing they would be really useful.

Nathan made sure that the Swift driver always passes `-x` to `lldb --repl` to preserve the old behavior for users of `swift` but if you call `lldb --repl` you can now also specify `-O` options.

jimingham:
I'm fine with the code for the change, that looks ok.  Thanks for the tests!

I don't think many people run the REPL from the lldb driver, most folks do it by running swift.

So making the run through the lldb command line a bit more wild west is fine too.

I wonder, however, if there is any reason to allow the driver in --repl mode to handle commands after creating the target?  You could for instance do "-o run" and we'd get really confused...  This doesn't seem necessary for your purposes, so it would be nice to outlaw it.  IIUC, you only want to add settings that modify how we ingest the target.  The only other use for passing commands to lldb on the command line is  to add breakpoints in code you were planning to exercise in the REPL.  But you can set breakpoints before a target is created, and they will get added to the REPL target when it is made.  So -O will do for that as well.

So a good compromise might be to only allow -O and -S options when passing --repl, and not -o and -s.  That would be simple to implement, just don't do the call to WriteCommandsForSource(eCommandPlacementAfterFile, commands_stream).  If we do that we should also check that m_option_data.m_after_file_commands is empty and warning that we're going to drop those commands if not.  And we should drop a note into the driver's --help output to this effect.

source/Core/ModuleList.cpp
nathawes: I'm not sure, sorry. @adrian-prantl do you know if there's a way?

tools/driver/Driver.cpp
nathawes: I think there was some confusion here, sorry. I *was* going to do that, but then I realized the existing behavior with `--repl` (and from swift) was actually already processing the .lldbinit from the home directory (it does so a little earlier in this function, before any of the command_stream/repl code changed here) and that behavior is actually useful as it gives us a way to set the new `swift-module-loading-mode` setting when running the repl via swift. I think this PR will change the behavior from swift, but only for users who have lldbinit files in their project directories and whose home directory lldbinit file sets the `target.load-cwd-lldbinit` setting to true (the default is to ignore them warn that they exist).

nathawes:
@swift-ci please test

nathawes:
@swift-ci please test

nathawes:
@swift-ci please test

nathawes:
Updated this to use the upstreamed commit with the driver changes (supporting -O and -S with --repl). Thanks for your help with that @adrian-prantl!

lit/Driver/TestRepl.test
```diff
@@ -0,0 +1,7 @@
+# RUN: echo ':quit' | %lldb -x --repl -O 'expr 42' -S %S/Inputs/Print2.in -o 'expr 999999' -s %s 2>&1 | FileCheck %s
```
jrose-apple, line 1: Nitpick: you can use `-b` to get LLDB to exit after it's processed all the -O/-o/-S/-s options.

adrian-prantl:
@swift-ci test and merge

PR #1332, opened 2019-02-28
[Host] Include Config.h to make sure READLINE_HAS_WCHAR_T is honoured.

dcci:
<rdar://problem/48401960>

### DIFF
--- a/include/lldb/Host/Editline.h
+++ b/include/lldb/Host/Editline.h
@@ -33,6 +33,8 @@
 #define liblldb_Editline_h_
 #if defined(__cplusplus)
 
+#include "lldb/Host/Config.h"
+
 #if LLDB_EDITLINE_USE_WCHAR
 #include <codecvt>
 #endif


dcci:
@swift-ci test and merge

dcci:
@adrian-prantl can you please review this one?

include/lldb/Host/Editline.h
```diff
@@ -33,6 +33,8 @@
 #define liblldb_Editline_h_
 #if defined(__cplusplus)
 
+#include "lldb/Host/Config.h"
```
adrian-prantl, line 36: For the record that file has
```
#if defined(__APPLE__)

// This block of code only exists to keep the Xcode project working in the
// absence of a configuration step.
#define LLDB_CONFIG_TERMIOS_SUPPORTED 1

#define LLDB_EDITLINE_USE_WCHAR 1
...
```

adrian-prantl (approved):
That should do the trick, yes.

include/lldb/Host/Editline.h
adrian-prantl: We'll need to find something cleaner for upstream / llvm.org

PR #1333, opened 2019-02-28
Match Objective-C timeout

JDevlieghere:
Make timeout match for Swift and Objective-C


JDevlieghere:

PR #1340, opened 2019-03-01
Allow LLDB repl to properly handle redefinition of expressions. 

bgogul:
Currently, lldb's REPL mode does not work well with extensions. Specifically, we cannot redefine an extension. Consider the following LLDB repl session:

```
Welcome to Swift version 5.0-dev (LLVM dcb9eb74a7, Clang 95cdf7c9af, Swift 56a798a3ea).
Type :help for assistance.
  1> struct Foo {}
  2> extension Foo { func f() -> Int { return 1 } }
  3> Foo().f()
$R0: Int = 1
  4> extension Foo { func f() -> Int { return 2 } }
  5> Foo().f()
error: repl.swift:5:1: error: ambiguous use of 'f()'
Foo().f()
^

repl.swift:2:22: note: found this candidate
extension Foo { func f() -> Int { return 1 } }
                     ^

repl.swift:4:22: note: found this candidate
extension Foo { func f() -> Int { return 2 } }
                     ^
```
As you can see redefining a function in an extension causes an ambiguous use diagnostic. This PR uses the additional hooks declared in https://github.com/apple/swift/pull/23031 to deal with redefinition of extensions. 

See the discussion on the [swift forum](https://forums.swift.org/t/lldb-repl-does-not-work-well-with-extensions/20057/15) for more context.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
@@ -543,7 +543,8 @@ public:
      auto sl = decl->getLoc();
      if (!sl.isValid())
        return 0;
-     if (m_source_manager.getDisplayNameForLoc(sl) != "repl.swift")
+     if (m_source_manager.getDisplayNameForLoc(sl) !=
+         SwiftREPL::GetSourceFileBasename())
        return 0;
      std::pair<unsigned, unsigned> line_col =
          m_source_manager.getLineAndColumn(sl);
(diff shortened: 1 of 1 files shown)


jimingham:
Please use SwiftREPL::GetSourceFileBasename rather than hardcoding "repl.swift".  Other than that, this should do the trick.

Jim

bgogul:
> Please use SwiftREPL::GetSourceFileBasename rather than hardcoding "repl.swift".  Other than that, this should do the trick.

Thanks for the quick response. Done!

dcci (changes requested):
This needs a test.

dcci:
Also, somebody working on the frontend should sign this off.

bgogul:
> This needs a test.

Thanks @dcci. I will certainly add a test. I am waiting for the https://github.com/apple/swift/pull/23031 PR to be merged in.

PR #1342, opened 2019-03-04
Set Swift ABI bit variable when creating a SwiftRuntime

mikeash:
The Swift metadata for classes uses a different bit to identify themselves in the pre-
and post-stable ABI eras. We need to be able to debug both, so the code in the MetadataReader
needs to be configured at runtime.

Companion to https://github.com/apple/swift/pull/23064.

rdar://problem/47935062


mikeash:
https://github.com/apple/swift/pull/23064

@swift-ci please test

mikeash:
apple/swift#23064

@swift-ci please test

PR #1344, opened 2019-03-05
 Avoid querying funciton debug info while building the symbol table.

adrian-prantl:
This fixes a testcase breakage introduced by 7eba1df14c04ce4ce812e5f0ab8f108899550001.

rdar://problem/48580111


adrian-prantl:
@swift-ci test and merge

PR #1345, opened 2019-03-05
[ExpressionParser] Remove line_offset parameter from UserExpression::Parse

bulbazord:
The parameter `line_offset` is set to 0 everywhere. It was being used to adjust swift diagnostic messages when they needed to be modified, but as-is it's basically unused.

This will also slightly reduce the diff between llvm.org/master and swift-lldb/stable, so that's nice.


bulbazord:
cc @compnerd @adrian-prantl @dcci @JDevlieghere

dcci (approved):

JDevlieghere:
LGTM, Thanks!

JDevlieghere:

PR #1352, opened 2019-03-06
Repair upstream-with-swift build

bulbazord:
This branch stopped building because of what appears to be a bad merge.

cc @compnerd @jimingham @JDevlieghere @dcci


dcci:
Jim is in the middle of fixing this.  The conflict is because ExpressionSourceCode pretends to be generic but actually has lots of both Clang & Swift specific stuff in it.  It's one of the last bits of the segregation of swift & clang that didn't get done. Should be done end of today or early tomorrow.

bulbazord:
Good to know, thanks. I'll go ahead and close this then.

PR #1354, opened 2019-03-06
Change the scanning algorithm in DynamicLoaderDarwinKernel::SearchFor…

jasonmolenda:
Change the scanning algorithm in DynamicLoaderDarwinKernel::SearchForKernelNearPC.

Currently when lldb might be doing a kernel debug session, it scans through
memory by taking the current pc value and looking for a kernel at megabyte
boundaries, up to 32MB behind $pc.  This adjusts the algorithm to
scan back at every 16k page boundary and to stop scanning as soon
as we hit a memory read error.  The addition of stopping at a memory read
error saves us from tons of unnecessary packet traffic on generic 
targets where lldb might look for a kernel binary.

I've been trying to think of how to construct a test for this; it's a bit
tricky.  A gdb-remote protocol test with the contents of a fake tiny kernel
mach-o binary would satisify part of it, but this kernel path also directly
calls over to dsymForUUID or DebugSymbols framework lookups to find the 
kernel binary as well.  I'll keep thinking about this one, but it's so
intertangled with these two external systems that it may be hard to do.

<rdar://problem/48578197> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@355476 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 1f74bbf12d19201c918fcd2766cee275f5aa3050)

### DIFF
--- a/source/Plugins/DynamicLoader/Darwin-Kernel/DynamicLoaderDarwinKernel.cpp
+++ b/source/Plugins/DynamicLoader/Darwin-Kernel/DynamicLoaderDarwinKernel.cpp
@@ -294,9 +294,11 @@ DynamicLoaderDarwinKernel::SearchForKernelNearPC(Process *process) {
     return LLDB_INVALID_ADDRESS;
   addr_t pc = thread->GetRegisterContext()->GetPC(LLDB_INVALID_ADDRESS);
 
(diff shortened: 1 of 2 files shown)


jasonmolenda:
@swift-ci test and merge

PR #1355, opened 2019-03-06
 Pass ConstString by value (NFC)

adrian-prantl:
My apologies for the large patch. With the exception of ConstString.h
itself it was entirely produced by sed.

ConstString has exactly one const char * data member, so passing a
ConstString by reference is not any more efficient than copying it by
value. In both cases a single pointer is passed. But passing it by
value makes it harder to accidentally return the address of a local
object.

(This fixes rdar://problem/48640859 for the Apple folks)

Differential Revision: https://reviews.llvm.org/D59030

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@355553 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 77fa533323ec662db8fd4af0a8643797214dc166)


adrian-prantl:
@swift-ci test and merge

PR #1356, opened 2019-03-06
Remove the warning in

jasonmolenda:
Remove the warning in
DynamicLoaderDarwinKernel::KextImageInfo::LoadImageUsingMemoryModule
which would list every kext that failed to load when doing kernel
debugging.  Instead, in DynamicLoaderDarwinKernel::ParseKextSummaries,
print a summary of how many kexts lldb was unable to load at the end.

I want to reduce the amount of output at the start of kernel debug
sessions a bit; we'll see if anyone really wanted to see the list of
which kexts specifically were unable to be loaded.

No functional change, only changing lldb's output at the start of
a kernel debug session.

<rdar://problem/48654569> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@355565 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit ee24b4b6d468ae57bd511caacb8dad6ed996bafa)


jasonmolenda:
@swift-ci test and merge

PR #1357, opened 2019-03-07
When disassembling Aarch64 target and vendor Apple, set the cpu to

jasonmolenda:
When disassembling Aarch64 target and vendor Apple, set the cpu to
"apple-latest" which llvm uses to indicate the newest supported ISA.
Add a unit test; I'm only testing an armv8.1 instruction in this
unit test which would already be disassembled correctly because we
set the disassembler to ARM v8.2 mode, but it ensures that nothing
has been broken by adding this cpu spec.

<rdar://problem/38714781> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@355578 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 53e897c22fc698872154f5feb61434e9302a9019)


jasonmolenda:
@swift-ci test and merge

PR #1361, opened 2019-03-08
Cherry-pick changes for ComputeClangResourceDirectory on Windows

bulbazord:
This is useful for having swift debugging on a Windows host, so I'd like to cherry-pick them into stable.

cc @compnerd @adrian-prantl

### DIFF
--- a/source/Plugins/ExpressionParser/Clang/ClangHost.cpp
+++ b/source/Plugins/ExpressionParser/Clang/ClangHost.cpp
@@ -53,9 +53,7 @@ static bool DefaultComputeClangResourceDirectory(FileSpec &lldb_shlib_spec,
 
   llvm::SmallString<256> clang_dir(parent_dir);
   llvm::SmallString<32> relative_path;
-  llvm::sys::path::append(relative_path,
-                          llvm::Twine("lib") + CLANG_LIBDIR_SUFFIX, "clang",
-                          CLANG_VERSION_STRING);
+  llvm::sys::path::append(relative_path, "lib", "lldb", "clang");
 
   llvm::sys::path::append(clang_dir, relative_path);
   if (!verify || VerifyClangPath(clang_dir)) {
--- a/source/Plugins/ExpressionParser/Clang/ClangHost.cpp
+++ b/source/Plugins/ExpressionParser/Clang/ClangHost.cpp
@@ -76,7 +76,7 @@ bool lldb_private::ComputeClangResourceDirectory(FileSpec &lldb_shlib_spec,
     ++rev_it;
   }
 
-  // Posix-style of LLDB detected.
+  // We found a non-framework build of LLDB
   if (rev_it == r_end)
     return DefaultComputeClangResourceDirectory(lldb_shlib_spec, file_spec,
                                                 verify);
--- a/source/Plugins/ExpressionParser/Clang/ClangHost.cpp
+++ b/source/Plugins/ExpressionParser/Clang/ClangHost.cpp
@@ -18,9 +18,6 @@
 
 #include "lldb/Host/FileSystem.h"
 #include "lldb/Host/HostInfo.h"
-#if !defined(_WIN32)
-#include "lldb/Host/posix/HostInfoPosix.h"
-#endif
 #include "lldb/Utility/FileSpec.h"
 #include "lldb/Utility/Log.h"
 
@@ -28,12 +25,6 @@
 
 using namespace lldb_private;
 
-#if defined(_WIN32)
-static bool ComputeClangResourceDirectory(FileSpec &lldb_shlib_spec,
-                                          FileSpec &file_spec, bool verify) {
-  return false;
-}
-#else
 static bool VerifyClangPath(const llvm::Twine &clang_path) {
   if (FileSystem::Instance().IsDirectory(clang_path))
     return true;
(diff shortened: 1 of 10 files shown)


source/Plugins/ExpressionParser/Clang/ClangHost.cpp
```diff
@@ -53,8 +42,37 @@ static bool VerifyClangPath(const llvm::Twine &clang_path) {
...
+                          CLANG_VERSION_STRING);
+
+  llvm::sys::path::append(clang_dir, relative_path);
+  if (!verify || VerifyClangPath(clang_dir)) {
+    file_spec.GetDirectory().SetString(clang_dir);
+    FileSystem::Instance().Resolve(file_spec);
+    return true;
+  }
+
+  return HostInfoPosix::ComputePathRelativeToLibrary(file_spec, relative_path);
```
compnerd: Why `HostInfoPosix` here?  Should this be conditional on `LLVM_ON_UNIX` and `LLVM_ON_WIN32` perhaps?

source/Plugins/ExpressionParser/Clang/ClangHost.cpp
```diff
@@ -67,8 +85,10 @@ bool lldb_private::ComputeClangDirectory(FileSpec &lldb_shlib_spec,
     ++rev_it;
   }
 
+  // Posix-style of LLDB detected.
```
compnerd: I don't understand this comment.

source/Plugins/ExpressionParser/Clang/ClangHost.cpp
bulbazord: I think you commented on one of the commits in the middle? It's definitely not HostInfoPosix in the final result.

source/Plugins/ExpressionParser/Clang/ClangHost.cpp
bulbazord: Yeah looking back this comment isn't very good. I intended it to indicate we aren't in a framework, but rather a more traditional tree with a `bin/ include/ lib/ share/` set of directories. I can update the comment on llvm.org and cherry-pick it into this PR with the rest of these commits.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
CC: @dcci @adrian-prantl @fredriss

bulbazord:
ping! @adrian-prantl

adrian-prantl:
Can you also add r356057 and r355994? I had to fixup the macOS implementation a little bit to get all tests to pass.

adrian-prantl:
Ah, you did!

adrian-prantl (approved):

compnerd:
@swift-ci please test

source/Plugins/ExpressionParser/Clang/ClangHost.cpp
```diff
@@ -53,9 +36,36 @@ static bool VerifyClangPath(const llvm::Twine &clang_path) {
...
+/// installed with the same prefix as lldb.
+///
+static bool DefaultComputeClangResourceDirectory(FileSpec &lldb_shlib_spec,
                                          FileSpec &file_spec, bool verify) {
   std::string raw_path = lldb_shlib_spec.GetPath();
+  llvm::StringRef parent_dir = llvm::sys::path::parent_path(raw_path);
+
+  llvm::SmallString<256> clang_dir(parent_dir);
+  llvm::SmallString<32> relative_path;
+  llvm::sys::path::append(relative_path, "lib", "lldb", "clang");
```
adrian-prantl, line 50: @xiaobai This is now different from llvm.org, where this line reads:
```
  llvm::sys::path::append(relative_path,
                          llvm::Twine("lib") + CLANG_LIBDIR_SUFFIX, "clang",
                          CLANG_VERSION_STRING);
```

Why is that? And could you please make sure to bring both versions back in sync?

bulbazord: From what I understand, swift-lldb relies on the clang resource copied from swift, which is placed in `$install_location/lib/lldb/clang`, whereas llvm.org's lldb relies on the clang resource dir located at `$install_location/lib/clang/$clang_version`. I'm not sure how I'd synchronize llvm.org and swift-lldb right now. Maybe a cmake variable that fills out Config.h?

adrian-prantl: I think you should be able to do something similar to what the macOS implementation does and just have a list of locations that are searched one after another. You can probably even reuse some of the code in the macOS implementation. If you think that that is too inefficient then we should detect the build type (e.g., by having CMake generate a macro) and make the code depend on that. Depending on the repository/branch is not a good idea, because some people *will* be building the swift-lldb compiler without swift.

bulbazord: Ah yeah looking at what the MacOS implementation does, that sounds like the best option. I should have some time in the next few days to fix that up. Thanks for pointing this out

adrian-prantl: Thanks! These kinds of differences between upstream and swift-lldb often cause extremely difficult to debug problems, especially when related to search paths, so this will pay off quite a lot.

bulbazord: @adrian-prantl I opened #1411 to cherry-pick the change to stable.

PR #1369, opened 2019-03-12
Add `-fmodule-map-file=` to RemapClangImporterOptions whitelist

DavidGoldman:
This flag was missed in the RemapClangImporterOptions whitelist. Let me know if I should instead create this PR against a different branch (stable/swift 5?) or if more information is needed.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1410,6 +1410,7 @@ namespace {
 bool ConsumeIncludeOption(StringRef &arg, StringRef &prefix) {
   static StringRef options[] = {"-I",
                                 "-F",
+                                "-fmodule-map-file=",
                                 "-iquote",
                                 "-idirafter",
                                 "-iframeworkwithsysroot",


DavidGoldman:
Friendly ping @adrian-prantl

adrian-prantl:
Sorry I missed this.

adrian-prantl:
@swift-ci test and merge

DavidGoldman:
@adrian-prantl is it possible to merge this into swift-5.0 as well?

adrian-prantl:
Unfortunately, the bar for any changes to swift-5.0 is *very* high at this point in time. Can you give an estimate for how important this bugfix is to you?

DavidGoldman:
@adrian-prantl At the moment, we can't debug any swift binaries built with Bazel due to this issue.

I'd like to confirm that this fixes it, but I'm unable to properly build swift-lldb:

(building with Xcode 10.2 beta 4, swift-5.0-branch as all active branches for swift repos)

`$ lldb/scripts/build-swift-cmake.py`
```
CMake Error at stdlib/public/core/cmake_install.cmake:36 (file):
  file INSTALL cannot find
  "/Users/davg/dev/github/swift-lldb/build/Ninja-DebugAssert/swift-macosx-x86_64/./lib/swift/iphoneos/armv7/Swift.swiftmodule".
Call Stack (most recent call first):
  stdlib/public/cmake_install.cmake:40 (include)
  stdlib/cmake_install.cmake:37 (include)
  cmake_install.cmake:37 (include)
```

and then during debugging:

```
 ((Target*)0x7fb5fe03bc00)->GetSwiftASTContext() returning NULL - couldn't load the Swift stdlib
Shared Swift state for GBus could not be initialized.
The REPL and expressions are unavailable.
```

DavidGoldman:
I also wonder if this remapping will fail to apply for relative paths. e.g. if we want to map a module a `/BuildRoot/relative/path/module.modulemap`, the `-fmodule-map-file=relative/path/module.modulemap`may not be properly remapped to `/BuildRoot/relative/path/module.modulemap` with a `.=/BuildRoot` remapping.

DavidGoldman:
Friendly ping, still can't build via the `lldb/scripts/build-swift-cmake.py` script:

`$ lldb/scripts/build-swift-cmake.py`
```
~/dev/swift-lldb
--- Installing swift ---
+ env DESTDIR=/Users/davg/dev/swift-lldb/install/ /Users/davg/.homebrew/bin/cmake --build /Users/davg/dev/github/swift-lldb/build/Ninja-DebugAssert/swift-macosx-x86_64 -- install
[0/1] Install the project...
...
-- Installing: /Users/davg/dev/swift-lldb/install/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift_static/iphoneos/libswiftCore.a
CMake Error at stdlib/public/core/cmake_install.cmake:65 (file):
  file INSTALL cannot find
  "/Users/davg/dev/github/swift-lldb/build/Ninja-DebugAssert/swift-macosx-x86_64/./lib/swift/iphonesimulator/i386/Swift.swiftmodule".
Call Stack (most recent call first):
  stdlib/public/cmake_install.cmake:40 (include)
  stdlib/cmake_install.cmake:37 (include)
  cmake_install.cmake:37 (include)
```

adrian-prantl:
I am not familiar with that script. How does it invoke swift/utils/build-script? It's complaining about the 32-bit iphone simulator runtime missing, which either means that it itnvoked build-script without `-i` or that is because the 32-bit iOS SDK isn't present in the SDK that you are building against, which I think may be the case in more recent versions of Xcode. For testing I usually don't bother with installing LLDB, so perhaps you can also just run it from the build directory.

CC: @weliveindetail

DavidGoldman:
It does indeed build lldb but lldb can't find the Swift stdlib modules - any ideas?

```
 ((Target*)0x7fb5fe03bc00)->GetSwiftASTContext() returning NULL - couldn't load the Swift stdlib
Shared Swift state for GBus could not be initialized.
The REPL and expressions are unavailable.
```

adrian-prantl:
Two things I would try is 
- adding -i to build_script_arguments in lldb/scripts/build-swift-cmake.py
- see if you can modify the swift cmake code to not expect an i386 runtime

that said, just invoking build-script -r -l -i -- --lldb-build-with-cmake (like the ci.swift.org cmake lldb bot does) should also result in an lldb that can find everything relative to its build location.

DavidGoldman:
Thanks, the `-i` flag fixed it. I've confirmed that this addition fixes debugging for Bazel-based Swift projects (with dSYMs), can you please merge this into the swift-5.0-branch? Otherwise we won't be able to debug :(

dmaclach:
> Unfortunately, the bar for any changes to swift-5.0 is very high at this point in time. Can you give an estimate for how important this bugfix is to you?

Hey @adrian-prantl , this bugfix is extremely important to us (Google) encouraging the use of Swift in-house. Without it we are basically 2-3x link times and that is dissuading large teams from adopting Swift. Happy to discuss offline if that is useful, and let me know whom I have to bribe 😉 .

dmaclach:
@adrian-prantl apologies.. wrong issue. This one is important to us (Google) because without it we are basically going to be unable to debug Swift 5 apps which would **DEFINITELY** dissuade large teams from adopting Swift. Still happy to discuss offline if that is useful, and let me know whom I have to bribe 😉 .

adrian-prantl:
Just a quick clarification: Do you need this on Darwin, Linux, or both?

DavidGoldman:
Primarily Darwin, I'd imagine debugging on Linux would run into the same issues though, so I guess both?

PR #1372, opened 2019-03-13
Simplify the Loading... message test, the regex test has been

jimingham:
failing intermittently though the printed output looks right.


jimingham:

PR #1375, opened 2019-03-13
[Playground/REPL] This test is flakey, skip it.

dcci:
Alex Brown is going to look at it. Matches a change on stable.


dcci:
@swift-ci test and merge

PR #1379, opened 2019-03-13
Reorder the operations in

jasonmolenda:
Reorder the operations in
DynamicLoaderDarwinKernel::KextImageInfo::LoadImageUsingMemoryModule a
bit so that we only read the binaries out of memory once we've
determined that we can find a real binary on the local system.  

Previously, lldb would read all of the kext binaries out of memory
and then determine if it had the local copy.  The kext table gives
us most the information we need (address, name, uuid) so lldb only
needs the actual in-memory load commands when it comes time to set
the section load addresses.  Delay reading until that point for all
the kexts.

NFC; doing the operations in a different order.

<rdar://problem/41181173> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@356108 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit c857e324bad30097ec3a94f7cc730fc8542687df)


jasonmolenda:
@swift-ci test and merge

PR #1381, opened 2019-03-15
Check the result of creating a node from __next_ in the std::list for…

jimingham:
…matter.

There's a single report of a crash coming from this current_sp being NULL.  I don't
have a repro case, and I couldn't get it to happen by hand-corrupting a list.  We
always get an error instead.  So I don't have a test case.  But checking for null
is clearly right here.

<rdar://problem/48503320>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@355957 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #1382, opened 2019-03-15
Pass arguments correctly to the objc object checker on arm64

jimingham:
Traditionally objc had two entry points, objc_msgSend for scalar
return methods, and objc_msgSend_stret for struct return convention
methods.  But on arm64 the second was not needed (since arm64 doesn't
use an argument register for the struct return pointer) so it was removed.

The code that dispatches to the objc object checker when it sees some
flavor of objc_msgSend was not aware of this change so was sending the
wrong arguments to the checker.

<rdar://problem/48315890>

Differential Revision: https://reviews.llvm.org/D58699

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@355026 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

jimingham:

jimingham:

PR #1383, opened 2019-03-15
When deserializing breakpoints some options may not be present.

jimingham:
The deserializer was not handling this case.  For now we just
accept the absent option, and set it to the breakpoint default.
This will be more important if/when I figure out how to serialize
the options set on breakpont locations.

<rdar://problem/48322664>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@354702 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

jimingham:

PR #1386, opened 2019-03-15
[Swift] Update for removal of -enable-parseable-module-interface

jrose-apple:
This mode is on by default in swiftc now.


jrose-apple:
apple/swift#23331
@swift-ci Please test

PR #1393, opened 2019-03-19
Support resilient values

adrian-prantl:
Specifically LLDB now calls into Remote Mirrors to determine whether a
resilient value that is stored in a fixed-size buffer is stored inline
or not.
    
rdar://problem/48018240

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/resilience/mod.a.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/resilience/mod.a.swift
@@ -12,12 +12,32 @@
 
 public struct S {
   public var a = 1
-  public var s1 = "i"
-  public var s2 = "am"
-  public var s3 = "large"
-  public var s4 = "!"
+  fileprivate var s1 = "i"
+  fileprivate var s2 = "am"
+  fileprivate var s3 = "large"
+  fileprivate var s4 = "!"
 
   public init() {}
+}
+
+fileprivate class Message { fileprivate var s = "hello" }
+fileprivate struct NotBitwiseTakable {
+  fileprivate weak var msg : Message?
+}
+
+fileprivate struct FixedContainer {
+    var s = S()
+}
+
+fileprivate var g_msg = Message()
+fileprivate var g_b = NotBitwiseTakable()
+fileprivate var g_s = S()
+fileprivate var g_t = (S(), S())
+fileprivate var g_c = FixedContainer()
+
+public func initGlobal() -> Int {
+  g_b.msg = g_msg
+  return g_s.a + g_t.0.a + g_c.s.a
 }
 
 public func fA(_ x: S) -> Int {
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1677,24 +1677,26 @@ SwiftLanguageRuntime::GetMemberVariableOffset(CompilerType instance_type,
                 failure.render().c_str());
 
   // Try remote mirrors.
-  if (!reflection_ctx)
-    return llvm::None;
-  ConstString mangled_name(instance_type.GetMangledTypeName());
-  StringRef mangled_no_prefix =
-      swift::Demangle::dropSwiftManglingPrefix(mangled_name.GetStringRef());
-  swift::Demangle::Demangler Dem;
-  auto demangled = Dem.demangleType(mangled_no_prefix);
-  auto *type_ref = swift::Demangle::decodeMangledType(
-      reflection_ctx->getBuilder(), demangled);
-  if (!type_ref)
-    return llvm::None;
-  auto type_info =
-      reflection_ctx->getBuilder().getTypeConverter().getTypeInfo(type_ref);
+  const swift::reflection::TypeInfo *type_info = GetTypeInfo(instance_type);
   if (!type_info)
     return llvm::None;
   auto record_type_info =
       llvm::dyn_cast<swift::reflection::RecordTypeInfo>(type_info);
   if (record_type_info) {
+    // Handle tuples.
+    if (record_type_info->getRecordKind() ==
+        swift::reflection::RecordKind::Tuple) {
+      unsigned tuple_idx;
+      if (member_name.GetStringRef().getAsInteger(10, tuple_idx) ||
+          tuple_idx >= record_type_info->getNumFields()) {
+        if (error)
+          error->SetErrorString("tuple index out of bounds");
+        return llvm::None;
+      }
+      return record_type_info->getFields()[tuple_idx].Offset;
+    }
+
+    // Handle other record types.
     for (auto &field : record_type_info->getFields()) {
       if (ConstString(field.Name) == member_name)
         return field.Offset;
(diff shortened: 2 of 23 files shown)


adrian-prantl:
@dcci  Can you give this a very thorough look please?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2386,6 +2386,48 @@ lldb::addr_t SwiftLanguageRuntime::FixupAddress(lldb::addr_t addr,
...
+  auto *type_ref = swift::Demangle::decodeMangledType(
+      reflection_ctx->getBuilder(), demangled);
+  if (!type_ref)
+    return nullptr;
+  return reflection_ctx->getBuilder().getTypeConverter().getTypeInfo(type_ref);
+}
+
+bool SwiftLanguageRuntime::IsIndirectResilientType(CompilerType type) {
+  swift::CanType swift_can_type(GetCanonicalSwiftType(type));
+  swift::NominalType *nominal_type =
```
slavapestov: Actually you don't need to reject non-nominal types or non-resilient nominals here. Consider these cases:

- A tuple containing two resilient structs
- A Fixed-layout struct containing a resilient struct

In both of these cases the decision to use the inline buffer storage or external storage still depends only on the size of the type. I think if you try these cases with your current implementation, this method will incorrectly return false.

I think just looking at the size is right because you already know statically if you have an inline buffer or not. In the case where you don't have a buffer there's no indirection (regardless of the resilience attribute on the type, etc). If there's a buffer, only the size determines whether there is indirection.

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2386,6 +2386,48 @@ lldb::addr_t SwiftLanguageRuntime::FixupAddress(lldb::addr_t addr,
...
+    return false;
+
+  swift::NominalTypeDecl *nominal_type_decl = nominal_type->getDecl();
+  if (!nominal_type_decl)
+    return false;
+
+  if (!nominal_type_decl->isResilient())
+    return false;
+
+  auto *type_info = GetTypeInfo(type);
```
slavapestov, line 2420: Yeah, this works with any type, not just nominals, so you can remove everything above

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2386,6 +2386,48 @@ lldb::addr_t SwiftLanguageRuntime::FixupAddress(lldb::addr_t addr,
   return addr;
 }
 
+const swift::reflection::TypeInfo *
+SwiftLanguageRuntime::GetTypeInfo(CompilerType type) {
+  swift::CanType swift_can_type(GetCanonicalSwiftType(type));
+  ConstString mangled_name(type.GetMangledTypeName());
```
slavapestov, line 2407: Does GetMangledTypeName use the debugger mangling? I think with a CanType there should be no difference (CanTypes don't contain Type Aliases) but just to be safe, you might want to write some tests that a global variable of type alias type still works (remote mirrors doesn't support type alias types since there's no alias -> underlying type mapping in the reflection metadata).

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2386,6 +2386,48 @@ lldb::addr_t SwiftLanguageRuntime::FixupAddress(lldb::addr_t addr,
...
+  if (!nominal_type_decl)
+    return false;
+
+  if (!nominal_type_decl->isResilient())
+    return false;
+
+  auto *type_info = GetTypeInfo(type);
+  if (!type_info)
+    return false;
+  return !type_info->isBitwiseTakable() || type_info->getSize() > 24;
```
slavapestov: As in the other PR, the 24 bytes is actually only valid on 64-bit, you should just use `word size * 3` instead.

packages/Python/lldbsuite/test/lang/swift/resilience/mod.a.swift
```diff
@@ -12,14 +12,20 @@
...
-  public var s4 = "!"
+  fileprivate var s1 = "i"
+  fileprivate var s2 = "am"
+  fileprivate var s3 = "large"
+  fileprivate var s4 = "!"
 
   public init() {}
 }
 
+var g_s = S()
```
slavapestov: I would add tests to cover a few more cases:

- Tuple containing two structs (total size < 24 bytes)
- Tuple containing two structs (total size > 24 bytes)
- Fixed layout struct containing resilient struct (total size < 24 bytes)
- Fixed layout struct containing resilient struct (total size > 24 bytes)
- A struct that is < 24 bytes but not bitwise takable and hence is still out of line (eg, it contains a `weak var`)

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2386,6 +2386,48 @@ lldb::addr_t SwiftLanguageRuntime::FixupAddress(lldb::addr_t addr,
...
+  swift::Demangle::Demangler Dem;
+  auto demangled = Dem.demangleType(mangled_no_prefix);
+  auto *type_ref = swift::Demangle::decodeMangledType(
+      reflection_ctx->getBuilder(), demangled);
+  if (!type_ref)
+    return nullptr;
+  return reflection_ctx->getBuilder().getTypeConverter().getTypeInfo(type_ref);
+}
+
+bool SwiftLanguageRuntime::IsIndirectResilientType(CompilerType type) {
```
slavapestov: Suggestion for a better name: `IsStoredInlineInBuffer()`

Also I believe this does the same thing as `SwiftASTContext::GetAllocationStrategy()`, except it uses RemoteAST and not IRGen. Should the latter be phased out in favor of the former (same with other IRGen TypeInfo-based queries)?

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
```diff
@@ -701,12 +701,6 @@ void SwiftASTManipulator::MakeDeclarationsPublic() {
                 llvm::dyn_cast<swift::AbstractStorageDecl>(decl))
           var_decl->overwriteSetterAccess(access);
       }
-
-      // FIXME: Remove this once LLDB has proper support for resilience.
-      if (swift::VarDecl *var_decl =
```
slavapestov: Cleanup suggestion: can you either file a radar on me, or prepare a patch to remove VarDecl::setREPLVar()? I don't believe its used for anything else.

slavapestov (commented):
Great to see this coming along!

source/Target/SwiftLanguageRuntime.cpp
slavapestov: In particular, your new implementation will never give an answer of `TypeAllocationStrategy::eDynamic` because RemoteMirrors can always give a definite result using runtime metadata.

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: > Actually you don't need to reject non-nominal types or non-resilient nominals here. 

isResilient() is a method of NominalTypeDecl and AbstractStorageDecl. How would I even query it for a tuple?

I was banking on fixed-size struct members already being handled by GetMemberVariableOffset but it's probably best to at least test this.

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: We got this sorted out: Tuples with resilient elements become one single fixed-size buffer.

adrian-prantl:
This update marks the first version of the patch that actually passes the entire lldb testsuite (including the REPL tests and a bunch of tests I added myself). I still haven't addressed all of @slavapestov 's feedback, I'll follow up with that as soon as I'm done with https://github.com/apple/swift/pull/23465/, which is also a prerequisite for this patch.

adrian-prantl:
Added tests for nested tuples and structs.

adrian-prantl:
test with https://github.com/apple/swift/pull/23425
@swift-ci test

adrian-prantl:
test with https://github.com/apple/swift/pull/23425
@swift-ci test

PR #1397, opened 2019-03-21
Fix gdb remote reproducer issues found while investigating flakiness

JDevlieghere:
Unfortunately this doesn't fix the flakiness itself, but a bug uncovered during the investigation.


JDevlieghere:

JDevlieghere:

PR #1401, opened 2019-03-22
[Reproducers] Fix GDB remote flakiness during replay

JDevlieghere:
This fixes the flakiness of the GDB remote reproducer during replay. It
was caused by a combination sending one ACK to many from the replay
server and the code that "flushes" any queued GDB remote packets in
GDBRemoteCommunicationClient::HandshakeWithServer.

The spurious ACK was the result of combining both implicit and explicit
handling of ACKs in the replay server. The handshake consists of an ACK
followed by an QStartNoAckMode. As long as we haven't seen any
QStartNoAckMode, we were sending implicit acknowledgments. So the first
ACK got acknowledged twice, once implicitly, and once as part of the
replay.

The reason we didn't notice this was the code in HandshakeWithServer
that "waits for any responses that might have been queued up in the
remote GDB server and flush them all". A 10ms timeout is used to move on
when no packets are left. If the second ACK didn't make it within those
10ms, all packets were offset by one.

### DIFF
--- a/lit/Reproducer/Functionalities/TestDataFormatter.test
+++ b/lit/Reproducer/Functionalities/TestDataFormatter.test
@@ -1,4 +1,4 @@
-# UNSUPPORTED: system-windows, system-freebsd, system-linux
+# UNSUPPORTED: system-windows, system-freebsd
 
 # This tests that data formatters continue to work when replaying a reproducer.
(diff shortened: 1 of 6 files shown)


JDevlieghere:

PR #1407, opened 2019-03-26
[CMake] macOS: Find DebugSymbols.framework inside the SDK

jrose-apple:
It's always going to be at / on disk, but that's not what we should be linking against!


jrose-apple:
@swift-ci Please test

dcci (commented):
If you feel like it, it's probably better to make this change upstream as this code is shared.
Otherwise, we can just merge this here and I'll commit it upstream for you.

jimingham:
Do you know why this change was required now.  There haven't been headers for DebugSymbols in /System for ages now, and lldb has always linked against this...

jrose-apple:
@dcci: Oh, right, of course. I'll do it there.

@jimingham: It actually isn't required; like I was saying on Apple-internal chat (before chat went down) I had a bad Xcode install. But it's more correct regardless.

jrose-apple:
https://reviews.llvm.org/D59831

PR #1408, opened 2019-03-27
Revert "[Swift] This test infloops, disable while people fix the unde…

dcci:
…ryling problem."

Erik committed a compiler fix.


dcci:
@swift-ci test and merge

PR #1410, opened 2019-03-28
Support resilient values.

adrian-prantl:
Specifically LLDB now calls into Remote Mirrors to determine whether a
resilient value that is stored in a fixed-size buffer is stored inline
or not.

rdar://problem/48018240
(cherry picked from commit afed7caf4365afc36fe319c9e9d5cd3bd8ca5006)


adrian-prantl:
test with https://github.com/apple/swift/pull/23622
@swift-ci test

PR #1412, opened 2019-03-28
[CMake] Re-add ExpressionSourceCode.cpp to CMake

bulbazord:
This file got removed from llvm.org but remains in swift-lldb. It looks
like this got dropped during a merge but it is needed to build
upstream-with-swift.

cc @dcci @jimingham @JDevlieghere @adrian-prantl @compnerd

### DIFF
--- a/source/Expression/CMakeLists.txt
+++ b/source/Expression/CMakeLists.txt
@@ -6,6 +6,7 @@ add_lldb_library(lldbExpression
   DiagnosticManager.cpp
   DWARFExpression.cpp
   Expression.cpp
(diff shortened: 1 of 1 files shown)


jimingham:
Yes, that's right.  We have some generic code that handles writing out the expression source to disk so lldb can list it if you stop at a breakpoint in code coming from an expression, which is only present on the Swift side.  That code needs to be migrated to llvm.org, at which point we will re-add ExpressionSourceCode.cpp there.  But until then we have to maintain this diff on upstream-with-swift.

PR #1139, opened 2018-12-14
Support for Swift-side work on "implementation imports"

jrose-apple:
See apple/swift#21313. Mostly here for combined pull-request testing at this point; I'll ask for review later.

PR #1414, opened 2019-03-28

jrose-apple:
More tests to come, but this gets the "don't crash" part in place, and help shake out a bug in the Swift implementation.

Builds on #1139, also depends on a small Swift change that isn't on GitHub yet (to not produce an error when an implementation-only import can't be loaded).

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/Makefile
@@ -1,0 +1,21 @@
+LEVEL = ../../../../make
+
+EXE=a.out
+SWIFT_SOURCES=main.swift
+SWIFTFLAGS_EXTRAS= -I$(shell pwd)
+MODULENAME=main
+LD_EXTRAS=-L$(shell pwd) -lSomeLibrary
+
+all: SomeLibrary.swiftmodule a.out
+
+include $(LEVEL)/Makefile.rules
+
+SomeLibrary.swiftmodule: SomeLibrary.swift
+	$(MAKE) MAKE_DSYM=NO CC=$(CC) SWIFTC=$(SWIFTC) \
+		ARCH=$(ARCH) DSYMUTIL=$(DSYMUTIL) \
+		BASENAME=$(shell basename $< .swift) \
+		VPATH=$(SRCDIR) -I $(SRCDIR) -f $(SRCDIR)/dylib.mk all
+
+clean::
+	rm -rf main *.dylib *.so *.dll *.dSYM *.swiftdoc *.swiftmodule *.swiftinterface
+
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
@@ -1,0 +1,65 @@
+# TestMainExecutable.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2019 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""
+Test `@_implementationOnly import` in the main executable
+"""
+import commands
+import lldb
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test.decorators import *
+import lldbsuite.test.lldbutil as lldbutil
+import os
+import os.path
+import time
+import unittest2
+
+class TestMainExecutable(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+    
+    @swiftTest
+    def test_implementation_only_import_main_executable(self):
+        """Test `@_implementationOnly import` in the main executable"""
+        self.build()
+        def cleanup():
+            lldbutil.execute_command("make cleanup")
+        self.addTearDownHook(cleanup)
+        lldbutil.run_to_source_breakpoint(self, "break here", lldb.SBFileSpec("main.swift"))
+
+        self.expect("fr var", substrs=[
+            "(SomeLibrary.TwoInts) value = (first = 2, second = 3)",
+            "(main.ContainsTwoInts) container = {\n  wrapped = (first = 2, second = 3)\n  other = 10\n}"])
+        self.expect("e value", substrs=["(SomeLibrary.TwoInts)", "= (first = 2, second = 3)"])
+        self.expect("e container", substrs=["(main.ContainsTwoInts)", "wrapped = (first = 2, second = 3)", "other = 10"])
+        self.expect("e TwoInts(4, 5)", substrs=["(SomeLibrary.TwoInts)", "= (first = 4, second = 5)"])
+    
+    @swiftTest
+    def test_implementation_only_import_main_executable_no_library_module(self):
+        """Test `@_implementationOnly import` in the main executable, after removing the library's swiftmodule"""
+        self.build()
+        os.remove(os.path.join(self.getBuildDir(), "SomeLibrary.swiftmodule"))
+        os.remove(os.path.join(self.getBuildDir(), "SomeLibrary.swiftinterface"))
+        def cleanup():
+            lldbutil.execute_command("make cleanup")
+        self.addTearDownHook(cleanup)
+        lldbutil.run_to_source_breakpoint(self, "break here", lldb.SBFileSpec("main.swift"))
+
+        self.expect("fr var", substrs=[
+            "value = <could not resolve type>",
+            "container = (other = 10)"])
+        # FIXME: Ideally we would still show that 'value' exists but make it unusable somehow.
+        self.expect("e value", error=True, substrs=["error: use of unresolved identifier 'value'"])
+        # FIXME: The representation of 'container' is garbage because we've dropped a field.
+        # The compiler should keep track of this.
+        self.expect("e container", substrs=["(main.ContainsTwoInts)", "other = "])
+        # This one's probably unavoidable, though.
+        self.expect("e TwoInts(4, 5)", error=True, substrs=["error: use of unresolved identifier 'TwoInts'"])
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/main.swift
@@ -1,0 +1,13 @@
+@_implementationOnly import SomeLibrary
+
+struct ContainsTwoInts {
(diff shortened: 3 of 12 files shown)


jrose-apple:
@adrian-prantl, @dcci, am I doing this correctly? Any test-writing feedback?

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/Makefile
```diff
@@ -0,0 +1,21 @@
...
+SWIFTFLAGS_EXTRAS= -I$(shell pwd)
+LD_EXTRAS=-L$(shell pwd) -lSomeLibrary
+
+all: SomeLibrary.swiftmodule main
+
+include $(LEVEL)/Makefile.rules
+
+SomeLibrary.swiftmodule: SomeLibrary.swift
+	$(MAKE) MAKE_DSYM=NO CC=$(CC) SWIFTC=$(SWIFTC) \
+		SWIFTFLAGS_EXTRAS="$(MOD$*_FLAGS)" \
```
adrian-prantl: How is `$(MOD$*_FLAGS)` expanded?

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
```diff
@@ -0,0 +1,86 @@
...
+import os
+import os.path
+import time
+import unittest2
+
+class TestMainExecutable(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
+    def launch_and_break(self):
```
adrian-prantl: This entire function can be replaced by `lldbutil.run_to_source_breakpoint(self, "break here", SBFileSpec("main.swift"), exe_name=self.getBuildArtifact("main"))`

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
```diff
@@ -0,0 +1,86 @@
...
+    @decorators.swiftTest
+    def test_implementation_only_import_main_executable(self):
+        """Test `@_implementationOnly import` in the main executable"""
+        self.build()
+        def cleanup():
+            lldbutil.execute_command("make cleanup")
+        self.addTearDownHook(cleanup)
+        self.launch_and_break()
+
+        self.expect("fr var", substrs=[
```
adrian-prantl, line 38: A slightly more robust way of writing this is
```
var = frame.FindVariable("value")
first = var.GetChildMemberWithName("first")
lldbutil.check_variable(self, first, value="2")
...
```

this way you don't depend on LLDB's formatting

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
```diff
@@ -0,0 +1,86 @@
...
+            "(SomeLibrary.TwoInts) value = (first = 2, second = 3)",
+            "(main.ContainsTwoInts) container = {\n  wrapped = (first = 2, second = 3)\n  other = 10\n}"])
+        self.expect("e value", substrs=["(SomeLibrary.TwoInts)", "= (first = 2, second = 3)"])
```
adrian-prantl, line 41: For expressions, this is as good as it gets, though

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
```diff
@@ -0,0 +1,86 @@
...
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""
+Test `@_implementationOnly import` in the main executable
+"""
+import commands
+import lldb
+from lldbsuite.test.lldbtest import *
+import lldbsuite.test.decorators as decorators
```
adrian-prantl: We usually `from lldbsuite.test.decorators import *`

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/Makefile
jrose-apple: Oops, that's left over from what I copied from. I should simplify this.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
jrose-apple: Ah! Where would I learn about this? Or should I just take a look at everything in lldbutil?

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/main.swift
```diff
@@ -0,0 +1,13 @@
+@_implementationOnly import SomeLibrary
```
adrian-prantl, line 1: What does this do?
And perhaps add a comment explaining it?

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
jrose-apple: I was deliberately not doing that because it made it harder to observe the structure and gave worse error messages when it failed. I admit to thrashing around on this test case, though.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable_resilient/TestMainExecutableResilient.py
```diff
@@ -0,0 +1,86 @@
...
+import os.path
+import time
+import unittest2
+
+class TestMainExecutable(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
+    def launch_and_break(self):
+        exe_name = "main"
```
adrian-prantl: See above.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable_resilient/TestMainExecutableResilient.py
```diff
@@ -0,0 +1,86 @@
...
+    @decorators.swiftTest
+    def test_implementation_only_import_main_executable_resilient(self):
+        """Test `@_implementationOnly import` in the main executable with a resilient library"""
+        self.build()
+        def cleanup():
+            lldbutil.execute_command("make cleanup")
+        self.addTearDownHook(cleanup)
+        self.launch_and_break()
+
+        self.expect("fr var", substrs=[
```
adrian-prantl, line 38: ditto

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/main.swift
jrose-apple: This is the new feature being tested. I don't think it makes sense to explain it here; eventually it'll be an actual Swift language feature.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable_resilient/TestMainExecutableResilient.py
jrose-apple: Yeah, this whole directory is identical to the other one but with `-enable-library-evolution` added to the Makefile.

adrian-prantl:
Look pretty good so far! Could you perhaps create the pull request for swift-5.1-branch instead? We have an automerger from 5.1 -> stable -> upstream-with-swift.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
```diff
@@ -0,0 +1,86 @@
...
+        # The compiler should keep track of this.
+            "container = (other = 2)"])
+        # FIXME: Ideally we would still show that 'value' exists but make it unusable somehow.
+        self.expect("e value", error=True, substrs=["error: use of unresolved identifier 'value'"])
+        # FIXME: Same problem as above.
+        self.expect("e container", substrs=["(main.ContainsTwoInts)", "other = 2"])
+        # This one's probably unavoidable, though.
+        self.expect("e TwoInts(4, 5)", error=True, substrs=["error: use of unresolved identifier 'TwoInts'"])
+
+if __name__ == '__main__':
```
dcci: You don't really need this `if __name__ == main` and below. Some tests have it, but it's just an historical artifact.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
dcci: I don't think it's documented anywhere, but arguably it should.

jrose-apple:
> Could you perhaps create the pull request for swift-5.1-branch instead? We have an automerger from 5.1 -> stable -> upstream-with-swift.

The Swift change hasn't made it to 5.1 yet (it hasn't even landed in master yet), but sure, I'll do the overall change that way.

dcci:
I think the structure of the tests is generally really good. I guess what we have are just minor comments.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
adrian-prantl: The example test uses it: https://github.com/apple/swift-lldb/blob/stable/packages/Python/lldbsuite/test/sample_test/TestSampleTest.py

I know.. "We have an example test?"

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable_resilient/TestMainExecutableResilient.py
```diff
@@ -0,0 +1,86 @@
...
+        source_spec = lldb.SBFileSpec(source_name)
+        breakpoint = target.BreakpointCreateBySourceRegex("break here", source_spec)
+        self.assertTrue(breakpoint.GetNumLocations() > 0, "Breakpoint set sucessfully")
+        
+        process = target.LaunchSimple(None, None, os.getcwd())
+        self.assertTrue(process, PROCESS_IS_VALID)
+    
+    @decorators.swiftTest
+    def test_implementation_only_import_main_executable_resilient(self):
+        """Test `@_implementationOnly import` in the main executable with a resilient library"""
```
adrian-prantl, line 31: Could you please just add a sentence or two outlining why `@_implementationOnly` is special and how it is expected to deviate or not from the normal behavior?
Inevitably, at some point somebody else will have to do maintenance on the test and without a comment explaining what behavior needs to be preserved, it can easily happen that the test devolves into something that isn't meaningful any more.

jrose-apple:
Moving to the 5.1 branch.

PR #1417, opened 2019-03-29
[Swift] The AST printer now qualifies type names more consistently

jrose-apple:
LLDB counterpart of apple/swift#22830.

This does have one breaking effect: formatters for protocols and typealiases will now have to match against the module name as well. Since this is consistent with structs and enums, this is probably worth it.

### DIFF
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -1187,9 +1187,11 @@ bool lldb_private::formatters::swift::GLKit_SummaryProvider(
 
   Process &process = *process_sp.get();
 
-  // Get the type name without the "simd.simd_" prefix.
+  // Get the type name without the "GLKit." prefix.
   ConstString full_type_name = valobj.GetTypeName();
   llvm::StringRef type_name = full_type_name.GetStringRef();
+  if (type_name.startswith("GLKit."))
+    type_name = type_name.drop_front(6);
 
   // Get the type of object this is.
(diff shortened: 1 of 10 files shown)


jrose-apple:
apple/swift#22830
@swift-ci Please test

jrose-apple:
apple/swift#22830
@swift-ci Please test

harlanhaskins (approved):

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -1187,9 +1187,11 @@ bool lldb_private::formatters::swift::GLKit_SummaryProvider(
 
   Process &process = *process_sp.get();
 
-  // Get the type name without the "simd.simd_" prefix.
+  // Get the type name without the "GLKit." prefix.
```
adrian-prantl, line 1190: Isn't this behavior inconsistent with the new printing behavior now? We probably do want the prefix if we print it everywhere else?

jrose-apple: 👍 This name is used for matching, not printing. The logic here is similar to what we were already doing for the (old) simd types above.

jrose-apple:
apple/swift#22830
@swift-ci Please test Linux

source/Plugins/Language/Swift/SwiftFormatters.cpp
adrian-prantl: Ah thanks!

dcci (approved):
Looks good. Thanks for updating the tests, Jordan.

PR #1419, opened 2019-03-29
Improve the error handling in GetInstanceVariableOffset[_offset].

adrian-prantl:
The old code would return LLDB_INVALID_OFFSET, which was interpreted
by other code as a valid offset of -1. Found by our ASAN bot.

rdar://problem/48981079


adrian-prantl:
@swift-ci test and merge

PR #1424, opened 2019-03-30
[Obj-C] Fix undefined behaviour(s) in the new NSTaggedDate formatter.

dcci:
Type punning through a union -> no good.
double to uint64 to double again -> no good either.

The nice side effect, other than silencing the sanitizer bot
is that it fixes the formatting of some dates, e.g. Jan 1st 1970.

<rdar://problem/47617983>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@353191 91177308-0d34-0410-b5e6-96231b3b80d8


dcci:
@swift-ci test and merge

dcci:
@fredriss looks like I just forgot to cherry-pick this after all :)

PR #1425, opened 2019-04-01
Fix the swig typemap for "uint32_t *versions, uint32_t num_versions".

jimingham:
It was making a list of a certain size but not always filling in that
many elements, which would lead to a crash iterating over the list.

Differential Revision: https://reviews.llvm.org/D59913

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@357207 91177308-0d34-0410-b5e6-96231b3b80d8


jimingham:

PR #1428, opened 2019-04-02
When iterating over breakpoint site owners::ShouldStop, don't hold the list mutex

jimingham:
The breakpoint site ShouldStop iterates over the list of owners calling ShouldStop, while holding the owner's list mutex.  If a ShouldStop call hits this same breakpoint on another thread we will deadlock against this mutex.  So copy the list and drop the lock and iterate over the copy to avoid the deadlock.

This patch also includes a fix-up for a bunch of our operator='s that were taking the pair of mutexes w/o worrying about priority inversion.


jimingham:

PR #1430, opened 2019-04-02
Simplify test and enable it on Linux.

adrian-prantl:
Note that self.frame is a function and I don't know what adverse
effects overwriting it may have had.


adrian-prantl:
@swift-ci test and merge

PR #1432, opened 2019-04-02
Renew the static bindings for the bots that don't have SWIG.

jimingham:
<rdar://problem/49520686>


jimingham:

PR #1436, opened 2019-04-03
Implement ComputeSwiftResourceDirectory work on all platforms

kusmour:
Implemented a new class SwiftHost taking care of Swift directory. Made ComputeSwiftResourceDirectory work for all platforms.
@xiaobai

### DIFF
--- a/source/API/CMakeLists.txt
+++ b/source/API/CMakeLists.txt
@@ -172,6 +172,7 @@ endif()
 set(lib_dir "${CMAKE_BINARY_DIR}/${CMAKE_CFG_INTDIR}/lib${LLVM_LIBDIR_SUFFIX}")
 set(CLANG_RESOURCE_PATH "${LLDB_PATH_TO_SWIFT_BUILD}/lib${LLVM_LIBDIR_SUFFIX}/swift/clang")
 set(clang_headers_target ${CLANG_RESOURCE_PATH}/include)
+set(SWIFT_RESOURCE_PATH "${LLDB_PATH_TO_SWIFT_BUILD}/lib${LLVM_LIBDIR_SUFFIX}/swift")
 if(NOT LLDB_BUILT_STANDALONE)
   set(clang_headers_target clang-headers symlink_clang_headers)
 endif()
@@ -185,9 +186,22 @@ add_custom_command_target(
     ALL
     DEPENDS ${clang_headers_target})
 
+add_custom_command_target(
+    unused_var
+    COMMAND "${CMAKE_COMMAND}" "-E" "copy_directory" "${SWIFT_RESOURCE_PATH}" "${lib_dir}/lldb/swift/"
+    OUTPUT "${lib_dir}/lldb/swift/"
+    VERBATIM
+    ALL
+    DEPENDS ${SWIFT_RESOURCE_PATH})
+
 install(
   CODE "file(MAKE_DIRECTORY ${lib_dir}/lldb)")
 
 install(
   DIRECTORY "${lib_dir}/lldb/clang"
   DESTINATION lib${LLVM_LIBDIR_SUFFIX}/lldb/)
+
+install(
+  DIRECTORY "${lib_dir}/lldb/swift"
+  DESTINATION lib${LLVM_LIBDIR_SUFFIX}/lldb/)
+
--- a/source/Plugins/ExpressionParser/Swift/SwiftHost.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftHost.cpp
@@ -1,0 +1,102 @@
+//===-- SwiftHost.cpp -------------------------------------------*- C++ -*-===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "SwiftHost.h"
+
+#include "lldb/Host/Config.h"
+#include "lldb/Host/FileSystem.h"
+#include "lldb/Host/HostInfo.h"
+#include "lldb/Utility/FileSpec.h"
+#include "lldb/Utility/Log.h"
+
+#include <string>
+
+using namespace lldb_private;
+
+static bool VerifySwiftPath(const llvm::Twine &swift_path) {
+  if (FileSystem::Instance().IsDirectory(swift_path))
+    return true;
+  Log *log = lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_HOST);
+  if (log)
+    log->Printf("VerifySwiftPath(): "
+                "failed to stat swift resource directory at \"%s\"",
+                swift_path.str().c_str());
+  return false;
+}
+
+static bool DefaultComputeSwiftResourceDirectory(FileSpec &lldb_shlib_spec,
+                                                 FileSpec &file_spec,
+                                                 bool verify) {
+  if (!lldb_shlib_spec)
+    return false;
+  Log *log = lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_HOST);
+  std::string raw_path = lldb_shlib_spec.GetPath();
+  // Drop bin (windows) or lib
+  llvm::StringRef parent_path = llvm::sys::path::parent_path(raw_path);
+
+  static const llvm::StringRef kResourceDirSuffixes[] = {
+      "lib/swift",
+      "lib" LLDB_LIBDIR_SUFFIX "/lldb/swift",
+  };
+  for (const auto &Suffix : kResourceDirSuffixes) {
+    llvm::SmallString<256> swift_path(parent_path);
+    llvm::SmallString<32> relative_path(Suffix);
+    llvm::sys::path::append(swift_path, relative_path);
+    if (!verify || VerifySwiftPath(swift_path)) {
+      if (log)
+        log->Printf("DefaultComputeSwiftResourceDir: Setting SwiftResourceDir "
+                    "to \"%s\", verify = %s",
+                    swift_path.str().str().c_str(), verify ? "true" : "false");
+      file_spec.GetDirectory().SetString(swift_path);
+      FileSystem::Instance().Resolve(file_spec);
+      return true;
+    }
+  }
+  return false;
+}
+
+bool lldb_private::ComputeSwiftResourceDirectory(FileSpec &lldb_shlib_spec,
+                                                 FileSpec &file_spec,
+                                                 bool verify) {
+#if !defined(__APPLE__)
+  return DefaultComputeSwiftResourceDirectory(lldb_shlib_spec, file_spec,
+                                              verify);
+#else
+  if (!lldb_shlib_spec)
+    return false;
+
+  std::string raw_path = lldb_shlib_spec.GetPath();
+  size_t framework_pos = raw_path.find("LLDB.framework");
+  if (framework_pos == std::string::npos)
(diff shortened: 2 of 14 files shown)


dcci:
Did you forget to rebase by any chance?

kusmour:
yea, i forgot to rebase on stable. everything should be good now

compnerd:
@swift-ci please test

dcci (approved):
LGTM, thanks.

dcci:
@swift-ci test and merge

compnerd:
@swift-ci please test Linux platform

dcci:
@kusmour this seems to break linux, can you please take a look?

kusmour:
> @kusmour this seems to break linux, can you please take a look?

will do

dcci:
@swift-ci please test

source/API/CMakeLists.txt
```diff
@@ -175,6 +175,8 @@ set(clang_headers_target ${CLANG_RESOURCE_PATH}/include)
 if(NOT LLDB_BUILT_STANDALONE)
   set(clang_headers_target clang-headers symlink_clang_headers)
 endif()
+set(SWIFT_RESOURCE_PATH "${LLDB_PATH_TO_SWIFT_BUILD}/lib${LLVMLIBDIR_SUFFIX}/swift")
+set(swift_headers_target "${SWIFT_RESOURCE_PATH}")
```
bulbazord: Hmm, you can probably get rid of `swift_headers_target` and just use `SWIFT_RESOURCE_PATH` everywhere. Unless there's a proper swift target we can use?

bulbazord (commented):
The change still looks fine to me.

@compnerd @dcci : I think you might want to do another pass over this change before it goes in. @kusmour had to copy the swift resource directory into the lldb build tree. It appears that `SwiftASTContext::GetResourceDir` was assuming that the `swift_dir` value it was passed would never be empty, but on Linux that appears to be the case because we now verify that the swift dir that we compute.
I think we should also probably clean up some of the logic in `SwiftASTContext::GetResourceDir`, as it seems kind of fragile.

@kusmour I think you also might want to cherry-pick this change to the swift-5.1-branch branch

dcci:
@xiaobai thanks. I'll review. In the meanwhile let's run the tests.

dcci:
@swift-ci please test

kusmour:
updated the SwiftHostTest check and replace the swift_headers_target with SWIFT_RESOURCE_PATH

dcci:
@swift-ci please test

source/API/CMakeLists.txt
```diff
@@ -185,9 +186,22 @@ add_custom_command_target(
     ALL
     DEPENDS ${clang_headers_target})
 
+# Copy the swift resource directory.
+add_custom_command_target(
```
dcci, line 189: This bit worries me. @JDevlieghere ?

dcci: or @adrian-prantl

dcci:
@swift-ci please test

gmittert:
@swift-ci please test

dcci:
This is not your fault, rerunning.

```
/home/buildnode/jenkins/workspace/swift-lldb-PR-Linux/branch-stable/swift-corelibs-foundation/Foundation/DispatchData+DataProtocol.swift:14:8: error: missing required module '_SwiftDispatchOverlayShims'
22:12:10 import Dispatch
22:12:10        
```

dcci:
@swift-ci please test linux

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

kusmour:
@compnerd @dcci
All test passed, please review the change. Thank you very much :D

source/API/CMakeLists.txt
```diff
@@ -185,9 +186,23 @@ add_custom_command_target(
     ALL
     DEPENDS ${clang_headers_target})
 
+add_custom_command_target(
+    unused_var
+    COMMAND "${CMAKE_COMMAND}" "-E" "copy_directory"
+    "${SWIFT_RESOURCE_PATH}" "${lib_dir}/lldb/swift/"
+    OUTPUT "${lib_dir}/lldb/swift/"
```
adrian-prantl, line 192: What is this doing?

unittests/Expression/SwiftParserTest.cpp
```diff
@@ -0,0 +1,69 @@
...
+    HostInfo::Initialize();
+  }
+  static void TearDownTestCase() {
+    HostInfo::Terminate();
+    FileSystem::Terminate();
+  }
+};
+} // namespace
+
+static std::string
```
adrian-prantl, line 33: There is already another unit test (host?) that tests Resource directory paths. Could we put this there?

source/Plugins/ExpressionParser/Swift/SwiftHost.cpp
```diff
@@ -0,0 +1,102 @@
...
+                                                 bool verify) {
+#if !defined(__APPLE__)
+  return DefaultComputeSwiftResourceDirectory(lldb_shlib_spec, file_spec,
+                                              verify);
+#else
+  if (!lldb_shlib_spec)
+    return false;
+
+  std::string raw_path = lldb_shlib_spec.GetPath();
+  size_t framework_pos = raw_path.find("LLDB.framework");
```
adrian-prantl, line 75: Not your code, but this should really use a sys::fs iterator...

source/Plugins/ExpressionParser/Swift/SwiftHost.cpp
```diff
@@ -0,0 +1,102 @@
...
+  if (framework_pos == std::string::npos)
+    return DefaultComputeSwiftResourceDirectory(lldb_shlib_spec, file_spec,
+                                                verify);
+
+  framework_pos += strlen("LLDB.framework");
+  raw_path.resize(framework_pos);
+  raw_path.append("/Resources/Swift");
```
adrian-prantl, line 82: and this should use sys::fs::append

source/API/CMakeLists.txt
kusmour: `SwiftASTContext::GetResourceDir` was assuming that the `swift_dir` passed in would never be empty, but on Linux it would be empty because we now verify the `swift_dir` we compute. So I had to copy the swift resource directory into the lldb build tree.

source/API/CMakeLists.txt
kusmour: `SwiftASTContext::GetResourceDir` was assuming that the `swift_dir` passed in would never be empty, but on Linux it would be empty because we now verify the `swift_dir` we compute. So I had to copy the swift resource directory into the lldb build tree.

unittests/Expression/SwiftParserTest.cpp
kusmour: You mean the `ClangParserTest` one? They are different resource directory and have different settings. It's better to have them separate

source/API/CMakeLists.txt
adrian-prantl: I'm not very good at reading CMake: What makes sure this code only runs on Linux?

kusmour: Sorry for the late reply. It doesn't, it copies on every platform. On OSX, the `swift_dir` will get to XCode one, but on other platform, it will have trouble finding the correct swift resource. That's because the `SwiftASTContext::GetResourceDir` makes some assumptions about the resource directory. First, the directory passed in cannot be empty, second if it's a build-script-driven LLDB build, then it will look for sibling directory, either way, we have to provide a valid directory for it.

adrian-prantl: Let me know if I'm misunderstanding something, but that sounds incorrect. On macOS, we distinguish two build schemes, one for local development of LLDB, where copying the Swift resource directory is useful/necessary, and one for building LLDB to ship with Xcode, and in the latter case, we must not copy any resource directories, because LLDB will find the actual Swift compiler's resource directory relative to itself in the Xcode installation.

bulbazord: If you look at SwiftASTContext::GetResourceDir (the one that returns a std::string), it tries several strategies to resolve the swift resource directory. If you built lldb using build-script on Linux, then the last strategy in that function is used. That strategy assumes that the parameter `swift_dir` that gets passed in is not empty, even if it's a bogus path. The last strategy currently takes the bogus path of `build/Ninja-DebugAssert/lldb-linux-x86_64/lib/swift` and transforms it into `build/Ninja-DebugAssert/swift-linux-x86_64/lib/swift`.

The problem is that this patch breaks that assumption. The ComputeSwiftResourceDirectory method always verifies that the path to the swift resource directory exists before returning it. This is the same behavior as ComputeClangResourceDirectory. The option that @kusmour went with (that I also think is the right thing to do) is to copy the swift resource directory into the lldb build tree. This ensures that we always will have a swift resource directory in a well-known location on every platform. I can understand if we do not want to install the swift resource directory into a toolchain for distribution because we expect the swift compiler to do it. But for running in the build tree, I think it makes sense.

source/API/CMakeLists.txt
weliveindetail: Indeed it looks suspicious, but `add_custom_command_target()` is a custom function defined in `swift/cmake/modules/SwiftAddCustomCommandTarget.cmake`. Not sure why it's used here. In other places we copy directories with builtin CMake functionality, e.g. https://github.com/apple/swift-lldb/blob/stable/cmake/modules/LLDBFramework.cmake#L106

weliveindetail (commented):
When touching `lldb.xcodeproj/project.pbxproj`, please always sort it with `scripts/sort-pbxproj.rb` before committing.

In the long term we should move away from custom our path variables (like `LLDB_PATH_TO_SWIFT_BUILD`) and towards CMake defaults, e.g. `Swift_DIR` that we use for `find_package` and the path variables this pulls in from SwiftConfig.cmake like `SWIFT_BINARY_DIR`.

I never got in touch with the C++ functions changed here. I don't know why these paths need to be computed at runtime and don't get passed in a header file via `#cmakedefine`? If it serves your use-case and doesn't break builds, I guess it's ok.

bulbazord:
> In the long term we should move away from custom our path variables (like LLDB_PATH_TO_SWIFT_BUILD) and towards CMake defaults, e.g. Swift_DIR that we use for find_package and the path variables this pulls in from SwiftConfig.cmake like SWIFT_BINARY_DIR.

https://github.com/apple/swift/pull/21388

I need to rebase my PR but this can't happen until that PR goes through.

> I never got in touch with the C++ functions changed here. I don't know why these paths need to be computed at runtime and don't get passed in a header file via #cmakedefine? If it serves your use-case and doesn't break builds, I guess it's ok.

Not sure why we're doing that either. The goal of this PR is to move this swift-specific functionality out of lldbHost and into the SwiftExpressionParser plugin, much like the way clang works. We could change it so the directory is computed ahead of time and stuck in a header file but I think that should be a separate change.

source/API/CMakeLists.txt
bulbazord: Yeah this could be just an `add_custom_command`. We can change the clang resource directory invocation too, in a separate change.

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test linux platform

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

compnerd:
I believe all the review comments have been addressed.  This also enables a bunch of additional things, so going to merge this.

PR #1441, opened 2019-04-03
 Enable previously XFAILed test.

adrian-prantl:
It looks like this was accidentally left XFAILed even though the
underlying bug had been resolved.

<rdar://problem/49526725>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/private_typealias/TestSwiftPrivateTypeAlias.py
+++ b/packages/Python/lldbsuite/test/lang/swift/private_typealias/TestSwiftPrivateTypeAlias.py
@@ -28,43 +28,28 @@ class TestSwiftPrivateTypeAlias(TestBase):
         TestBase.setUp(self)
 
     @swiftTest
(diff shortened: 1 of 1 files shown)


adrian-prantl:
Do the typename changes look right, or is there still a bug?

adrian-prantl:
Here's the source:

```
public struct MyStruct {
    fileprivate typealias IntegerType = Int
    private var m_integer = IntegerType(345)
    fileprivate func Foo(_ i : IntegerType)
    {
        print("i = \(i)") // breakpoint 1
    }
    
    fileprivate func Bar()
    {
        let a : Dictionary<String, IntegerType> = [ "hello" : 234 ]
        print("a = \(a)") // breakpoint 2
    }
    
}

func main() {
    let s = MyStruct()
    s.Foo(123)
    s.Bar()
}

main()
```

adrian-prantl:
@swift-ci test and merge

jrose-apple:
Yes, the intervening ".Type" was always wrong. Thanks, Adrian!

PR #1445, opened 2019-04-03
Disable format-security warnings for the SWIG wrappers.

fredriss:
Sometimes we don't have a choice but to be unsafe when bridging.

rdar://problem/46500945.


fredriss:
@swift-ci test and merge

fredriss:
@swift-ci test and merge

PR #1446, opened 2019-04-04
Ignore "safe" early fatal errors when importing Swift modules.

adrian-prantl:
Some early module import diagnostics do not utterly destroy the Swift
AST context, so add them to the list of ignored diagnostics.

Unbreaks the Swift backwards compatibility tests.


adrian-prantl:
@swift-ci test and merge

PR #1447, opened 2019-04-04
[test] Enable library evolution for swiftinterface test libraries

harlanhaskins:
Since we've changed the behavior of interface printing for non-resilient
types, and since we're not supporting module interfaces for
non-resilient modules, enable library evolution for these libraries so
the tests that exercise information-hiding remain working.


nathawes (approved):

jrose-apple (approved):

harlanhaskins:
https://github.com/apple/swift/pull/23782

@swift-ci please test

PR #1448, opened 2019-04-05
Enable verbose logging to debug an issue only appearing on the bots.

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #1450, opened 2019-04-05
Unify random timeouts throughout LLDB and make them configurable.

adrian-prantl:
Since these timeouts guard against catastrophic error in debugserver,
I also increased all of them to the maximum value among them.

The motivation for this test was the observation that an asanified
LLDB would often exhibit seemingly random test failures that could be
traced back to debugserver packets getting out of sync. With this path
applied I can no longer reproduce the one particular failure mode that
I was investigating.

rdar://problem/49441261

Differential Revision: https://reviews.llvm.org/D60340

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@357829 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 555e11a99fa590bdd39ba50941fb2eeb509c1e2b)

 Conflicts:
	source/Target/Process.cpp

### DIFF
--- a/include/lldb/Target/Process.h
+++ b/include/lldb/Target/Process.h
@@ -67,36 +67,22 @@ public:
   ~ProcessProperties() override;
 
   bool GetDisableMemoryCache() const;
-
   uint64_t GetMemoryCacheLineSize() const;
-
   Args GetExtraStartupCommands() const;
-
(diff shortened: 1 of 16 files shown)


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

PR #1451, opened 2019-04-07
[CMake] Don't force some CMake variables

bulbazord:
I'd like to avoid forcing values on these variables because they overwrite any user set value (e.g. something set in a cmake cache file). I understand these are desirable defaults however, so I set them in build-script-impl here: https://github.com/apple/swift/pull/23842

cc @weliveindetail @dcci @compnerd


compnerd (commented):
I'm not sure that this change is the right thing to do.  Would it not be sufficient to just drop the `FORCE` on the `set`?  That would preserve the default behaviour and allow the user to override it either from the command line or a cache file.

bulbazord:
Yes, that should also work. I would be okay with that solution, so I'll restore the `set`s and just drop the `FORCE`.

compnerd:
@swift-ci please test

compnerd:
CC: @adrian-prantl

weliveindetail:
Yes good point. Original plan here was to move to CMake caches sooner, thus the `FORCE` (but it's been delaying). This would move settings like this from `LLDBConfig.cmake`/`build-script` to `lldb/cmake/cache/<TargetSpecificName>.cmake` à la https://github.com/llvm/llvm-project/tree/master/clang/cmake/caches

PR #1455, opened 2019-04-09
Get the run locker before you ask if your thread is valid.

jimingham:
I have occasional crashes coming from SBThread::GetExtendedBacktraceThread.  The 
symptom is that we got true back from HasThreadScope - so we should have a valid
live thread, but then when we go to use the thread, it is not good anymore and we
crash.
I can't spot any obvious cause for this crash, but in looking for same I noticed
that in the current code we check that the thread is valid, THEN we take the stop 
locker.  We really should do that in the other order, and ensure that the process 
will stay stopped before we check our thread is still good.  That's what this patch does.

<rdar://problem/47478205>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@357963 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/source/API/SBThread.cpp
+++ b/source/API/SBThread.cpp
@@ -1345,9 +1345,9 @@ SBThread SBThread::GetExtendedBacktraceThread(const char *type) {
   ExecutionContext exe_ctx(m_opaque_sp.get(), lock);
   SBThread sb_origin_thread;
 
-  if (exe_ctx.HasThreadScope()) {
-    Process::StopLocker stop_locker;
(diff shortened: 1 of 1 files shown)


jimingham:

jimingham:

jimingham:

dcci:
@jimingham Linux PR testing was broken but this should be good to go now.

dcci:
@swift-ci test and merge

dcci (approved):

dcci:
@swift-ci test and merge

fredriss:
@swift-ci test and merge

PR #1457, opened 2019-04-09
[Batch mode] Cherry-pick fix for error suppression

davidungar:
Addresses rdar://49303574
Requires https://github.com/apple/swift/pull/23879

When compiler issues errors for nonprimaries that do not show up when that file is primary, batch mode never shows the user the specific error. This PR addresses that issue by passing the responsible primary into the `FileSpecificDiagnosticConsumer` so that it can place the error in the primary's .dia file.


davidungar:
https://github.com/apple/swift/pull/23879

@swift-ci please test

dcci (approved):
LGTM, although I thought 5.0 was dead at this point.

PR #1461, opened 2019-04-10
Retrieve the availability platform names from the compiler instead of

adrian-prantl:
hardcoding them. (NFC)

rdar://problem/49659039


adrian-prantl:
@swift-ci test and merge

PR #1462, opened 2019-04-11
Split lit tests to improve testing turnaround

JDevlieghere:
We identified the 3 longest running tests and split them up so lit can run them in parallel.


JDevlieghere:

dcci:
I'm not entirely sure why, but this broke the CMake bots. Given it's not extremely urgent and it's not entirely obvious to me how to fix, I'm going to revert this patch.

PR #1464, opened 2019-04-11
Fix flakyness in TestCommandScriptImmediateOutput

JDevlieghere:
I'm not sure why this surfaced at this particular point, but
TestCommandScriptImmediateOutput (a pexpect test) had a synchronization
issue, where the (lldb) promts it was expecting were getting out of
sync. This happened for two reasons:
- it did not expect the initial (lldb) prompt we print at startup
- launchArgs() returned None, which resulted in an extra "target create
  None" command being issued to lldb (and an extra unhandled prompt
  being printed).

Resolving these two issues seems to fix (or at least, improve) the test.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@357459 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 8909047b1a7e66ea1ecb99e8127a7fca469dedce)


JDevlieghere:

PR #1466, opened 2019-04-11
[Expression] Remove unneeded functions

bulbazord:
These functions are implemented in ClangExpressionSourceCode and
SwiftExpressionSourceCode respectively, so these functions are never
actually used. They also create build failures, because they reference
things from ClangExpressionSourceCode that they can't see (e.g.
c_start_marker).

I understand this is a source of divergence between llvm.org and swift-lldb. If possible, we should pull the relevant changes into stable and push things back to llvm.org to make merges easier.

cc @dcci @adrian-prantl @JDevlieghere @compnerd


compnerd:
@swift-ci please test

JDevlieghere:
Thanks Alex, I think I mis-merged this from stable into upstream-with-swift. I was under the impression that these were used. 

@swift-ci please test

compnerd:
@swift-ci please test

bulbazord:
Looks like upstream-with-swift was rebased and this got taken care of, so I'm going to close this.

PR #1473, opened 2019-04-16
Copy the Clang resource directly directly from the clang build

fredriss:
When building a toolchain version of LLDB, build-script will not
create the Swift resource dir. Get the Clang headers directly from
the Clang build.

### DIFF
--- a/cmake/modules/LLDBFramework.cmake
+++ b/cmake/modules/LLDBFramework.cmake
@@ -98,8 +98,13 @@ add_custom_command(TARGET lldb-framework-headers POST_BUILD
 
 # Copy vendor-specific headers from clang (without staging).
 if(NOT IOS)
+  execute_process(
+     COMMAND ${LLDB_PATH_TO_CLANG_BUILD}/bin/clang -print-resource-dir
+     OUTPUT_VARIABLE CLANG_RESOURCE_PATH
+     OUTPUT_STRIP_TRAILING_WHITESPACE)
+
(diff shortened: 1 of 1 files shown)


fredriss:
@swift-ci test

fredriss:
The second part of lldb-framework-headers (copy_directory ${LLDB_PATH_TO_SWIFT_BUILD}/lib/swift $<TARGET_FILE_DIR:liblldb>/Resources/Swift) is only required to run the TestSuite, and I think it should be made a dependency of the tests. It should not be there in toolchain builds.

dcci (approved):
This seems reasonable.

JDevlieghere (approved):

weliveindetail:
Yes I think it's acceptable. If it works, then let's take it for now.

There's more urgent things than this, but FYI:
* `execute_process(COMMAND ${LLDB_PATH_TO_CLANG_BUILD}/bin/clang ...)` only works as we build standalone; `add_dependencies(lldb-framework-headers clang)` would fix in-tree builds
* Clang saves the path to built-in headers in `$<TARGET_PROPERTY:clang-resource-headers,RUNTIME_OUTPUT_DIRECTORY>`, which could be used instead
https://github.com/llvm/llvm-project/blob/master/clang/lib/Headers/CMakeLists.txt#L161

weliveindetail (approved):

fredriss:
@swift-ci test

PR #1475, opened 2019-04-16
[debugserver] Relax the codesigning identity check

fredriss:
In an effort to help new LLDB developers, we added checks and messaging around
the selection of your codesigning identity on macOS. While helpful, it is not
actually correct. It's perfectly valid to codesign with an identity that is
not named lldb_codesign. Currently this fails the build.

This patch keeps a warning that informs developers how to setup lldb_codesign
and how to pass it to cmake, but it allows the build to proceed with a
different identity.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@358525 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 15fd2e284f5cb4f6a251c5627247cd4e6702785a)


fredriss:
@swift-ci test and merge

fredriss:
@swift-ci test and merge

PR #1482, opened 2019-04-18
[CMake] Emit LLDB.framework.dSYM

weliveindetail:
In order to avoid potential name collision with driver's lldb.dSYM

Summary:
Emit framework's dSYM bundle as LLDB.framework.dSYM instead of LLDB.dSYM, because the latter could conflict with the driver's lldb.dSYM when emitted in the same directory on case-insensitive file systems.
Requires https://reviews.llvm.org/D60862

Reviewers: friss, beanz, bogner

Subscribers: mgorny, lldb-commits, #lldb

Tags: #lldb

Differential Revision: https://reviews.llvm.org/D60863

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@358686 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/source/API/CMakeLists.txt
+++ b/source/API/CMakeLists.txt
@@ -14,6 +14,10 @@ if(NOT LLDB_DISABLE_PYTHON)
   set(lldb_python_wrapper ${lldb_scripts_dir}/LLDBWrapPython.cpp)
 endif()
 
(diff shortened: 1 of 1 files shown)


weliveindetail:
@swift-ci test and merge

PR #1486, opened 2019-04-19
Cherry-pick changes from llvm.org for LLDBStandalone.cmake

bulbazord:
I want these changes for the following reasons:
- They help reduce the diff between llvm.org and swift.org (and thus reduce merge conflicts for upstream-with-swift).
- They allow you to run the swift-lldb test suite on Windows
- They allow you to build lldb-server for android from swift-lldb

One of these commits relies on https://github.com/apple/swift-llvm/pull/150, which has been merged for a few days. I think this is probably safe at this point.

If I understand correctly, merging this into swift-5.1-branch means it will get automerged into stable. Seems like the right thing to do to prevent merge conflicts between swift-5.1-branch and stable.


bulbazord:
cc @adrian-prantl @JDevlieghere @dcci @weliveindetail

bulbazord:
cc @compnerd

dcci (approved):
I have no problems with these changes but Stefan or Fred should sign them off.

fredriss:
@xiaobai One of the commits claims that standalone builds are broken, but that's how we build swift-lldb all the time (build-script uses separate cmake builds for clang, swift and swift-lldb) . Care to explain a little more?

bulbazord:
@fredriss Sure! The context around that claim is that building and testing lldb standalone against an install tree of LLVM/Clang didn't have the correct values of `LLVM_MAIN_INCLUDE_DIR`, `LLVM_LIBRARY_DIR`, and `LLVM_BINARY_DIR`. They were using `LLVM_BUILD_*` values, which are only set if you build against an LLVM/Clang build tree. My understanding is that when swift-lldb is built using build-script, swift-lldb is built against an llvm/clang build tree instead of an install tree. I think that saying standalone builds are broken lacks nuance, but I didn't author that commit and I didn't want to needlessly modify it.

fredriss:
Makes sense. If this passes PR testing, I'm fine with it.

fredriss:
@swift-ci test and merge

PR #1490, opened 2019-04-19
make lldb work with python3.6 and python3.7 at the same time

marcrasi:
This will allow people to use our prebuilt toolchains (which are built with python3.6) even if they have python3.7.

In principle, this is something that we should be able to upstream to main lldb. But this is a slightly hacky temporary solution, so I think we should wait until we have a more proper implementation before doing that.


dan-zheng (approved):
LGTM (as a temporary solution)!

PR #1491, opened 2019-04-20

jrose-apple:
More tests to come, but this gets the "don't crash" part in place, and helped shake out a bug in the Swift implementation. Continuing from #1414 now that the compiler side is in the 5.1 branch.


jrose-apple:
@swift-ci Please test

jrose-apple:
Did not make 5.1, moving back to `stable` in #1838.

PR #1494, opened 2019-04-22
[SwiftREPL] Relax REPL tests to account for vendor substring.

dcci:
"Welcome to Apple Swift" VS "Welcome to Swift".
This is user-configurable, so it makes sense to relax the string
matching.


dcci:
@swift-ci test and merge

PR #1496, opened 2019-04-23
Add a small check to PlatformDarwin::LoadScriptingResourceForModule

jasonmolenda:
Add a small check to PlatformDarwin::LoadScriptingResourceForModule
which reads the python files in a dSYM bundle, to check that the
SymbolFile is actually a dSYM bundle filepath; delay any fetching
of the ScriptInterpreter until after we've done that check.

When debugging a binary without a dSYM on darwin systems, the
SymbolFile we fetch is actually the ObjectFile -- so we would do
an unnecessary trip into Python land and stat around the filesystem
looking for a python file to read in.  There's no reason to do any
of this unless the SymbolFile's file path includes the .dSYM bundle
telltale path components.

<rdar://problem/50065315> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@358938 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 11331029bcca42047f70953e7a990c27bd3f78fd)


jasonmolenda:
@swift-ci test and merge

PR #1497, opened 2019-04-23
Cherry-pick fix for writing of non-gpr registers on newer processors

GaynullinDima:
Cherry-pick from: [https://reviews.llvm.org/rG38a82413210009f6796249a63ae65c15bd548ddd](url)

Explanations: We're using ptrace(PTRACE_SETREGSET, NT_X86_XSTATE) to write all non-gpt
registers on x86 linux. Unfortunately, this method has a quirk, where
the kernel rejects all attempts to write to this area if one supplies a
buffer which is smaller than the area size (even though the kernel will
happily accept partial reads from it).

This means that if the CPU supports some new registers/extensions that
we don't know about (in my case it was the PKRU extension), we will fail
to write *any* non-gpr registers, even those that we know about.

Since this is a situation that's likely to appear again and again, I add
code to NativeRegisterContextLinux_x86_64 to detect the runtime size of
the area, and allocate an appropriate buffer. This does not mean that we
will start automatically supporting all new extensions, but it does mean
that the new extensions will not prevent the old ones from working.

This fixes tests attempting to write to non-gpr registers on new intel
processors (cca Kaby Lake Refresh).

### DIFF
--- a/source/Plugins/Process/Linux/NativeRegisterContextLinux_x86_64.cpp
+++ b/source/Plugins/Process/Linux/NativeRegisterContextLinux_x86_64.cpp
@@ -18,7 +18,7 @@
 
 #include "Plugins/Process/Utility/RegisterContextLinux_i386.h"
 #include "Plugins/Process/Utility/RegisterContextLinux_x86_64.h"
-
+#include <cpuid.h>
 #include <linux/elf.h>
 
 using namespace lldb_private;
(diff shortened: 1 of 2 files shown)


GaynullinDima:
@swift-ci Please test

dcci:
@swift-ci please test

dcci:
@GaynullinDima Thanks for your contribution. Which test fixes exactly?

GaynullinDima:
@dcci It fixes 
lldb-Suite :: functionalities/register/register_command/TestRegisters.py
lldb-Suite :: tools/lldb-server/TestGdbRemoteRegisterState.py
on Skylake (and newer) CPUs

dcci:
```
12:10:42 
/home/buildnode/jenkins/workspace/swift-lldb-PR-Linux/branch-stable/lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_x86_64.cpp:283:8: error: use of undeclared identifier '__get_cpuid_count'
12:10:42   if (!__get_cpuid_count(0x0d, 0, &eax, &ebx, &ecx, &edx))
12:10:42        ^
```

@GaynullinDima this is failing to build, can you take a look? (you can access the log)

GaynullinDima:
@dcci It fixes 
lldb-Suite :: functionalities/register/register_command/TestRegisters.py
lldb-Suite :: tools/lldb-server/TestGdbRemoteRegisterState.py
on Skylake (and newer) CPUs

GaynullinDima:
Yes, I'll look

compnerd:
@swift-ci please test Linux platform

compnerd:
@dcci - that error has been extremely common recently - it's not related to this change.

dcci:
@swift-ci test linux

GaynullinDima:
@dcci Hello,
I've got stuck because it's actually building on my machine and as I understood there is the problem with the lack of function in the header which is [there](https://github.com/apple/swift-clang/blob/c1ff7aaac57a47d4064268adc6d0e2ccd949c41b/lib/Headers/cpuid.h#L300). Or maybe it uses other header on a build system.

dcci (changes requested):
You might want to try this on the same setup of the bots (ubuntu 16.04). Which OS are you running?

dcci:
This has been inactive for a while, closing (we'll get the change for free the next time we rebranch anyway).

PR #1499, opened 2019-04-23
[EditLineTests] Call setenv() before editline is initialized.

dcci:
This was the real reason why the tests were failing.


dcci:
@swift-ci test

JDevlieghere (approved):

PR #1500, opened 2019-04-23
Lock accesses to OptionValueFileSpecList objects

fredriss:
Before a Debugger gets a Target, target settings are routed to a global set
of settings. Even without this, some part of the LLDB which exist independently
of the Debugger object (the Module cache, the Symbol vendors, ...) access
directly the global default store for those settings.

Of course, if you modify one of those global settings while they are being read,
bad things happen. We see this quite a bit with FileSpecList settings. In
particular, we see many cases where one debug session changes
target.exec-search-paths while another session starts up and it crashes when
one of those accesses invalid FileSpecs.

This patch addresses the specific FileSpecList issue by adding locking to
OptionValueFileSpecList and never returning by reference.

Reviewers: clayborg

Subscribers: lldb-commits

Differential Revision: https://reviews.llvm.org/D60468

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359028 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 4cf4c548788c81efd22f448e494ec4b62278311a)

### DIFF
--- a/include/lldb/Interpreter/OptionValueFileSpecList.h
+++ b/include/lldb/Interpreter/OptionValueFileSpecList.h
@@ -9,6 +9,8 @@
 #ifndef liblldb_OptionValueFileSpecList_h_
 #define liblldb_OptionValueFileSpecList_h_
 
+#include <mutex>
+
 #include "lldb/Core/FileSpecList.h"
 #include "lldb/Interpreter/OptionValue.h"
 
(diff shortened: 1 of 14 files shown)


fredriss:
@swift-ci test and merge

fredriss:
@swift-ci test and merge

PR #1502, opened 2019-04-24
Fix infinite recursion when calling C++ template functions

fredriss:
Summary:
When we encounter a templated function in the debug information, we
were creating an AST that looked like this:

FunctionTemplateDecl 0x12980ab90 <<invalid sloc>> <invalid sloc> foo<int>
|-TemplateTypeParmDecl 0x12980aad0 <<invalid sloc>> <invalid sloc> class depth 0 index 0 T
|-FunctionDecl 0x12980aa30 <<invalid sloc>> <invalid sloc> foo<int> 'int (int)' extern
| |-TemplateArgument type 'int'
| `-ParmVarDecl 0x12980a998 <<invalid sloc>> <invalid sloc> t1 'int'
`-FunctionDecl 0x12980aa30 <<invalid sloc>> <invalid sloc> foo<int> 'int (int)' extern
  |-TemplateArgument type 'int'
  `-ParmVarDecl 0x12980a998 <<invalid sloc>> <invalid sloc> t1 'int'

Note that the FunctionTemplateDecl has 2 children which are identical (as
in have the same address). This is not what Clang is doing:

FunctionTemplateDecl 0x7f89d206c6f8 </tmp/template.cpp:1:1, line:4:1> line:2:5 foo
|-TemplateTypeParmDecl 0x7f89d206c4a8 <line:1:10, col:19> col:19 referenced typename depth 0 index 0 T
|-FunctionDecl 0x7f89d206c660 <line:2:1, line:4:1> line:2:5 foo 'int (T)'
| `-ParmVarDecl 0x7f89d206c570 <col:9, col:11> col:11 t1 'T'
`-FunctionDecl 0x7f89d206cb60 <line:2:1, line:4:1> line:2:5 used foo 'int (int)'
  |-TemplateArgument type 'int'
  `-ParmVarDecl 0x7f89d206ca68 <col:9, col:11> col:11 t1 'int':'int'

The 2 chidlren are different and actually repesent different things: the first
one is the unspecialized version and the second one is specialized. (Just looking
at the names shows another major difference which is that we create the parent
with a name of "foo<int>" when it should be just "foo".)

The fact that we have those 2 identical children confuses the ClangImporter
and generates an infinite recursion (reported in https://llvm.org/pr41473).
We cannot create the unspecialized version as the debug information doesn't
contain a mapping from the template parameters to their use in the prototype.

This patch just creates 2 different FunctionDecls for those 2 children of the
FunctionTemplateDecl. This avoids the infinite recursion and allows us to
call functions. As the XFAILs in the added test show, we've still got issues
in our handling of templates. I believe they are mostly centered on the fact
that we create do not register "foo" as a template, but "foo<int>". This is
a bigger change that will need changes to the debug information generation.
I believe this change makes sense on its own.

Reviewers: shafik, clayborg, jingham

Subscribers: aprantl, javed.absar, kristof.beyls, lldb-commits

Differential Revision: https://reviews.llvm.org/D61044

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359140 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 8d0aad6ddf65a5e52cf1626a5b3ff6d91d13a054)

### DIFF
--- a/packages/Python/lldbsuite/test/lang/cpp/template-function/Makefile
+++ b/packages/Python/lldbsuite/test/lang/cpp/template-function/Makefile
@@ -1,0 +1,5 @@
+LEVEL = ../../../make
+
+CXX_SOURCES := main.cpp
+
+include $(LEVEL)/Makefile.rules
--- a/packages/Python/lldbsuite/test/lang/cpp/template-function/TestTemplateFunctions.py
+++ b/packages/Python/lldbsuite/test/lang/cpp/template-function/TestTemplateFunctions.py
@@ -1,0 +1,30 @@
+"""
+Test that we can call C++ template fucntions.
+"""
+import lldb
+from lldbsuite.test.decorators import *
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test import lldbutil
+
+
+class TemplateFunctionsTestCase(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
+    def do_test_template_function(self, add_cast):
+        self.build()
+        (_, _, thread, _) = lldbutil.run_to_name_breakpoint(self, "main")
+        frame = thread.GetSelectedFrame()
+        expr = "foo(42)"
+        if add_cast:
+            expr = "(int)" + expr
+        expr_result = frame.EvaluateExpression(expr)
+        self.assertTrue(expr_result.IsValid())
+        self.assertEqual(expr_result.GetValue(), "42")
+
+    def test_template_function_with_cast(self):
+        self.do_test_template_function(True)
+
+    @expectedFailureAll(debug_info=["dwarf", "gmodules"])
+    def test_template_function_without_cast(self):
+        self.do_test_template_function(False)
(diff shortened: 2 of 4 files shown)


fredriss:
@swift-ci test and merge

fredriss:
@swift-ci test and merge

shahmishal:
@swift-ci test and merge

dcci:
@fredriss The Linux buildbot started failing recently, could this change be the cause?

```
14:38:37 
14:38:37 ********************
14:38:37 Testing: 0 .. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.. 
14:45:20 Testing Time: 578.83s
14:45:20 ********************
14:45:20 Failing Tests (1):
14:45:20     lldb-Suite :: lang/cpp/template-function/TestTemplateFunctions.py
14:45:20 
```

dcci:
Link to the failure: https://ci.swift.org/job/oss-lldb-incremental-linux-ubuntu-16_04/7910/console

dcci:
Only fails in DWO variant.

PR #1505, opened 2019-04-25
[DataFormatters] Adjusting libc++ std::list formatter to act better with pointers and references and adding a test to cover a previous related fix

shafik:
Summary:
This previous fix https://github.com/llvm-mirror/lldb/commit/5469bda296c183d1b6bf74597c88c9ed667b3145 did not have a test since we did not have a reproducer.

This is related to how formatters deal with pointers and references. The added tests both the new behavior and covers the previous bug fix as well.

Differential Revision: https://reviews.llvm.org/D60588

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359118 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 65c885bf9a215d6b05432f5f138378eb96f0e730)

### DIFF
--- a/packages/Python/lldbsuite/test/functionalities/data-formatter/data-formatter-stl/libcxx/list/TestDataFormatterLibcxxList.py
+++ b/packages/Python/lldbsuite/test/functionalities/data-formatter/data-formatter-stl/libcxx/list/TestDataFormatterLibcxxList.py
@@ -162,6 +162,10 @@ class LibcxxListDataFormatterTestCase(TestBase):
                              '[2] = ', '3',
                              '[3] = ', '4'])
 
+        ListPtr = self.frame().FindVariable("list_ptr")
+        self.assertTrue(ListPtr.GetChildAtIndex(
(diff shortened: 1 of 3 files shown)


shafik:
@swift-ci test

JDevlieghere:
@shafik Should we upstream this change? I'm asking because this caused a merge conflict on upstream-with-swift, which I've resolved to your change here.

shafik:
@fredriss what do you think about @JDevlieghere question? Should we upstream the change?

fredriss:
@shafik I'm missing something. Was this not done upstream?

shafik:
@fredriss Actually I was confused for a moment there, it should be.

@JDevlieghere this should already be upstream, so let's take a look at it together when you come in.

JDevlieghere:
Turns out I merged it incorrectly, I've fixed it in 8403de8084. I got confused because the regexes got changed after the stable branch got cut, and the change was only part of the line that conflicted.

PR #1508, opened 2019-04-26
[swig] Provide a way to force static bindings

JDevlieghere:
The current logic in prepare_bindings.py makes it impossible to force it
to use the static bindings. You either have to pass the swig executable
or allow the script to try to find it itself. This is undesirable when
you just want to use the static bindings, for example when the version
of swig is too old.

### DIFF
--- a/scripts/prepare_bindings.py
+++ b/scripts/prepare_bindings.py
@@ -159,7 +159,7 @@ def process_args(args):
 
     group = parser.add_argument_group("static binding usage")
     group.add_argument(
-        "--allow-static-binding",
+        "--use-static-binding",
         action="store_true",
         help=(
             "Specify the pre-baked binding can be used if "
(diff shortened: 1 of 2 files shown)


scripts/prepare_bindings.py
```diff
@@ -277,11 +277,11 @@ def main(args):
         if options.find_swig:
```
compnerd, line 266: Why not just ignore the swig handling if static bindings are in use?  That is `--find-swig` is meaningless.  Simply hoist the check above the `if options.find_swig`.

JDevlieghere: That would break the scenario where you are okay with using the static bindings if swig doesn't exist, but you still want to try and find it. I'm honestly not sure if that's a scenario that we need to support, but it's what the script does currently.

compnerd: Hmm, I see.  Personally, I think that is something we should explicitly prohibit.  You select between the static or dynamic behaviour.  Using swig if available makes it unpredictable.  What happens if I uninstalled swig and forgot about that?  Suddenly the build switches behaviour.  But, if there is a strong reason to support that, then, sure, the change LGTM.

JDevlieghere:
Saleem, what do you think of this approach?

compnerd (approved):
Thanks!

PR #1511, opened 2019-04-26
UN-XFail TestRegisters.py

fredriss:
All bots should be running a recent enough debugserver for this test to pass.


fredriss:
@swift-ci test and merge

PR #1514, opened 2019-04-26
Pass explicit C++ version to test

fredriss:
stop-hook-threads.cpp uses C++11 features, so ask for C++11 explicitely.
This isn't broken on mainstream because clang defaults to a newer C++
version now, but it breaks if testing against an older compiler.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359349 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit e5d2e62e50092bc5604054ff9f123359c4255fcc)


fredriss:
@swift-ci test and merge

PR #1517, opened 2019-04-26
[dotest.py] Remove swiftpr category

JDevlieghere:
Now that we use CMake for Swift PR testing, we have the lit filter  to
distinguish swift tests. The swift annotations are therefore no longer
necessary. This commit removes them.


JDevlieghere:
Oops, looks like I removed the whole line for the inline tests.

dcci (approved):
LGTM

PR #1518, opened 2019-04-27
[SwiftASTContext] Library search paths should be consistent with the …

dcci:
…ones in repl_swift.

The general idea behind the REPL is that of creating a dummy
target for lldb to attach to run swift expressions. The dummy target
dlopen()s libswiftCore, so that people don't have to type `import Swift`
when they open it. In order to find libswiftCore, the build system sets
a couple of @rpath, giving priority to the just built stdlib (if any)
and falling back to the OS standard library (in /usr/lib/swift).

Once the REPL has been launched, if somebody tries to import, e.g.
Foundation, lldb calls SwiftASTContext::LoadLibraryUsingPaths, which
asks GetLibrarySearchPaths() where to look. GetLibrarySearchPaths()
sets the first rpath to /usr/lib/swift and then falls back to the
stdlib just built.

In this tragedy of commons, lldb ends up loading two different
libSwiftCore which might be not compatible with each other. This
results in weird crashes, because in presence of symbols implemented
in multiple libraries, the one that's picked is undefined.

<rdar://problem/50022178>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -3472,9 +3472,16 @@ bool SwiftASTContext::LoadOneImage(Process &process, FileSpec &link_lib_spec,
 
 static std::vector<std::string>
 GetLibrarySearchPaths(const swift::SearchPathOptions &search_path_opts) {
-  std::vector<std::string> paths(search_path_opts.LibrarySearchPaths.begin(),
-                                 search_path_opts.LibrarySearchPaths.end());
+  // The order in which we look up the libraries is important. The REPL
+  // dlopen()s libswiftCore, and gives precedence to the just built standard
+  // library instead of the one in the OS. When we type `import Foundation`,
+  // we want to make sure we end up loading the correct library, i.e. the
+  // one sitting next to the stdlib we just built, and then fall back to the
+  // one in the OS if that's not available.
+  std::vector<std::string> paths;
   paths.push_back(search_path_opts.RuntimeLibraryPath);
+  for (std::string path : search_path_opts.LibrarySearchPaths)
+    paths.push_back(path);
   return paths;
 }
 
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -3479,7 +3479,8 @@ GetLibrarySearchPaths(const swift::SearchPathOptions &search_path_opts) {
   // one sitting next to the stdlib we just built, and then fall back to the
   // one in the OS if that's not available.
   std::vector<std::string> paths;
-  paths.push_back(search_path_opts.RuntimeLibraryPath);
+  if (!search_path_opts.RuntimeLibraryPath.empty())
+    paths.push_back(search_path_opts.RuntimeLibraryPath);
   for (std::string path : search_path_opts.LibrarySearchPaths)
     paths.push_back(path);
   return paths;


dcci:
@swift-ci please test

dcci:
I guess I can write an unittest for this, but I would like to get extra pair of eyes on this to validate/confute my reasoning.

mikeash (approved):

jimingham (commented):
We should consult @bob-wilson about this.  The reason why we force /usr/lib/swift first is because if somebody were to load a swift framework into the REPL that has a copy of swift stdlib in it on a machine that has swift in the OS, then we want to make sure the REPL loads the /usr/lib/swift version rather than the one from the framework.  IIUC what we want to do is make sure the order is:

locally built swift standard library path
/usr/lib/swift
anything else

So maybe we need to know explicitly whether and if yes where the locally build swift is explicitly so we can enforce this ordering.  We probably do know that somewhere else.

dcci:
@jimingham Interesting.
When I first looked at this in a debugger I found out that this is `search_paths` for the REPL.

```
(lldb) p search_paths
(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >) $1 = size=2 {
  [0] = "/usr/lib/swift"
  [1] = "/Users/local/stash/build/Ninja-DebugAssert+stdlib-Debug/lldb-macosx-x86_64/bin/LLDB.framework/Resources/Swift/macosx"
}
```

and after the fix is of course the same swapped.

So, the list is composed looking at two components:
`search_path_opts.RuntimeLibraryPath` and `search_path_opts.LibrarySearchPaths` 

To the best of my understanding, the former is always set in lldb to the library path sitting in your build. The latter has always `/usr/lib/swift` first (and then something else, or nothing as in this case).

So it looks like the code is set up in a way that the order you propose is enforced naturally? [as long as we're carefully to keep `/usr/lib/swift` always the first fallback].

Does this make sense?

jimingham:
That makes sense to me.  We should probably document that we require this ordering somewhere if it isn't already.  But otherwise this seems right to me.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3472,9 +3472,16 @@ bool SwiftASTContext::LoadOneImage(Process &process, FileSpec &link_lib_spec,
...
 GetLibrarySearchPaths(const swift::SearchPathOptions &search_path_opts) {
-  std::vector<std::string> paths(search_path_opts.LibrarySearchPaths.begin(),
-                                 search_path_opts.LibrarySearchPaths.end());
+  // The order in which we look up the libraries is important. The REPL
+  // dlopen()s libswiftCore, and gives precedence to the just built standard
+  // library instead of the one in the OS. When we type `import Foundation`,
+  // we want to make sure we end up loading the correct library, i.e. the
+  // one sitting next to the stdlib we just built, and then fall back to the
+  // one in the OS if that's not available.
+  std::vector<std::string> paths;
```
JDevlieghere, line 3481: (Nit) We might want to reserve the vector.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3472,9 +3472,16 @@ bool SwiftASTContext::LoadOneImage(Process &process, FileSpec &link_lib_spec,
...
   paths.push_back(search_path_opts.RuntimeLibraryPath);
```
JDevlieghere: I’m curious how we mirror the situation of the different rpaths for builds and installs. What’s the value in the later scenario? Is it empty?

JDevlieghere (commented):
The idea makes sense. One question and nit inline.

source/Symbol/SwiftASTContext.cpp
dcci: Are you talking about the value of `RuntimeLibraryPath`? If so, yes, it's empty.

JDevlieghere: Alright, that's what I expected.

source/Symbol/SwiftASTContext.cpp
dcci: This is really a micro-optimization, as this codepath is called only once per dylib loaded. I looked what's the typical # of times this is called on some projects, and it's ~100 (ballpark).
Also, the number of elements we end up having in the array is almost always 2 (one search path is `/usr/lib/swift`, and the other is the `stdlib` you just built). 

That said, your comment made me realize that we should only push the path if it's not an empty string. I guess I'll make that change.

source/Symbol/SwiftASTContext.cpp
bob-wilson: For Swift 5.1, we are moving to have Apple's toolchains not contain the Swift dylibs so the only case where this code will find libraries in the toolchain is when doing development or using a downloaded Swift.org toolchain. In that case, you should set DYLD_LIBRARY_PATH to prefer the toolchain libraries. If you don't set DYLD_LIBRARY_PATH, it may still not work because other dylibs may have dependencies on /usr/lib/swift and you'll still end up with multiple versions of the same library getting loaded. I'm not sure about the context of this code, but if that's what it is doing, then you should be ok.

dcci:
@swift-ci test and merge

PR #1521, opened 2019-04-30
repl: correct linking against `libdl`

compnerd:
Not all targets link against libdl.  NetBSD is one such target.  Use
`CMAKE_DL_LIBS` which will correctly specify the spelling as well as
configure the value as appropriate for the target.


compnerd:
CC: @dcci @xiaobai

compnerd:
@swift-ci please test

dcci (approved):
LGTM.

PR #1523, opened 2019-04-30
 Support local Swift variables called "self" that shadow the "self" fu...

adrian-prantl:
…nction argument.

This specifically makes the common "guard let self = self" pattern work.

<rdar://problem/39611934>


adrian-prantl:
@swift-ci test and merge

PR #1525, opened 2019-05-02
 Change the implementation of SwiftLanguageRuntime::IsRuntimeSupportValue …

adrian-prantl:
from a blacklist to a whitelist.

rdar://problem/45322477

### DIFF
--- a/include/lldb/Target/SwiftLanguageRuntime.h
+++ b/include/lldb/Target/SwiftLanguageRuntime.h
@@ -453,7 +453,6 @@ protected:
 
   CompilerType m_box_metadata_type;
 
-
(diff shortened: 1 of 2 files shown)


adrian-prantl:
@swift-ci test

jimingham (approved):
This is okay on the lldb side.  If there were some way to tell clang/swift developers that IsArtificial has the meaning - among other things - of: "if not self, won't get shown in the debugger" then I'd be happier, so nobody will accidentally cause parameters or locals to get hidden.  But that's on the swift/clang side.

adrian-prantl:
> If there were some way to tell clang/swift developers that IsArtificial has the meaning

Since that developer is in both cases myself, consider this done.

PR #1527, opened 2019-05-02
[Swift] Set the ParseableInterfaceModuleLoader's prebuilt module cache path

nathawes:
It's now being set to `<resource dir>/<platform>/prebuilt-modules` to match its default value in the Swift compiler.

Resolves rdar://problem/47611329

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/parseable_interfaces/shared/TestSwiftInterfaceNoDebugInfo.py
+++ b/packages/Python/lldbsuite/test/lang/swift/parseable_interfaces/shared/TestSwiftInterfaceNoDebugInfo.py
@@ -48,6 +48,31 @@ class TestSwiftInterfaceNoDebugInfo(TestBase):
             open(self.getBuildArtifact(module), 'w').close()
         self.do_test()
 
+    @swiftTest
+    @skipUnlessPlatform(["macosx"])
+    def test_prebuilt_cache_location(self):
+        """Verify the prebuilt cache path is correct"""
+        self.build()
+        log = self.getBuildArtifact("types.log")
+        self.runCmd('log enable lldb types -f "%s"' % log)
+
+        # Set a breakpoint in and launch the main executable so we load the
+        # ASTContext and log the prebuilt cache path
+        lldbutil.run_to_source_breakpoint(
+           self, "break here", lldb.SBFileSpec("main.swift"),
+           exe_name=self.getBuildArtifact("main"))
+
+        # Check the prebuilt cache path in the log output
+        found = False
+        prefix = 'Using prebuilt module cache path: '
+        logfile = open(log, "r")
+        for line in logfile:
+            if prefix in line:
+                self.assertTrue(line.endswith('/macosx/prebuilt-modules\n'), 'unexpected prebuilt cache path: ' + line)
+                found = True
+        self.assertTrue(found, 'prebuilt cache path log entry not found')
+
+
     def setUp(self):
         TestBase.setUp(self)
 
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -65,6 +65,7 @@
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/Path.h"
 #include "llvm/Support/Process.h"
 #include "llvm/Support/TargetRegistry.h"
 #include "llvm/Support/TargetSelect.h"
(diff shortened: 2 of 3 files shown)


nathawes:
@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/parseable_interfaces/shared/TestSwiftInterfaceNoDebugInfo.py
```diff
@@ -48,6 +48,31 @@ def test_swift_interface_fallback(self):
...
+
+        # Check the prebuilt cache path in the log output
+        found = False
+        prefix = 'Using prebuilt module cache path: '
+        logfile = open(log, "r")
+        for line in logfile:
+            if prefix in line:
+                self.assertTrue(line.endswith('/macosx/prebuilt-modules\n'), 'unexpected prebuilt cache path: ' + line)
+                found = True
+        self.assertTrue(found, 'prebuilt cache path log entry not found')
```
adrian-prantl, line 73: would it make sense to also add a check that tests that there is such a directory in `$(xcrun --show-sdk-path)` even if it won't be the same one (because the test is run on a just-built compiler)? This way we could catch the case where the directory is renamed in a future SDK and we need to update the hardcoded path.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3009,9 +3010,17 @@ swift::ASTContext *SwiftASTContext::GetASTContext() {
...
   // Install the parseable interface module loader.
-  std::string ModuleCachePath = GetClangImporterOptions().ModuleCachePath;
-  StringRef PrebuiltModuleCachePath =
-      GetCompilerInvocation().getFrontendOptions().PrebuiltModuleCachePath;
+  std::string moduleCachePath = GetClangImporterOptions().ModuleCachePath;
+
+  // Use the default location of the prebuilt module cache:
+  // <resource-dir>/<platform>/prebuilt-modules
+  llvm::Triple triple(GetTriple());
+  llvm::SmallString<64> prebuiltModuleCachePath = GetResourceDir(triple);
```
adrian-prantl: Is 64 enough for a path deep inside Xcode.app?

adrian-prantl: Otherwise we should either make it much smaller or large enough for /Applications/Xcode.app

nathawes: Good point. Bumped it up to 128 (the default Xcode toolchain case needs 115).

nathawes:
@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/parseable_interfaces/shared/TestSwiftInterfaceNoDebugInfo.py
jrose-apple: The directory is in the toolchain, not the SDK…but it also won't exist for a just-built compiler.

nathawes:
OS X failure is:
```
SwiftASTContext.cpp:2995:37: error: VALID_OR_RETURN not found on method 'GetASTContext'; consider adding (or whitelisting if necessary, --help for further information)
```
But that method does have `VALID_OR_RETURN` in it. Took a look with Adrian and we're not sure what about this change triggered the checker to start complaining. We decided to add it to the white list.

nathawes:
@swift-ci please test

packages/Python/lldbsuite/test/lang/swift/parseable_interfaces/shared/TestSwiftInterfaceNoDebugInfo.py
adrian-prantl: *Because* it doesn't exist for a just-built compiler, I thought it would be valuable to have an integration test that verifies that the path being constructed here matches the one that the host toolchain also uses.

jrose-apple: Ah, I see what you mean. That would set a minimum Xcode requirement for building, though. Certainly no released Xcodes have this folder, and we're in the open source repo!

nathawes: @adrian-prantl I filed rdar://problem/50501694 to track adding such this test once public CI is using a new enough Xcode. Are you happy for this to go in in the meantime?

adrian-prantl (approved):

jrose-apple (approved):

PR #1528, opened 2019-05-02
Disabling test in TestClassTemplateParameterPack.py until we do template lookup correctly

shafik:
Summary:
Some tests currently only work because we are pulling all the local variables when we are evaluating an expression. This will soon
change and these test are working but for the wrong reasons. The details can be found in the discussion here:
http://lists.llvm.org/pipermail/lldb-commits/Week-of-Mon-20180507/040689.html

Differential Review:  https://reviews.llvm.org/D61266

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359699 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 3ce4f605628023708c30581f84467544b3187492)


shafik:
@swift-ci test

PR #1530, opened 2019-05-02
[SwiftASTContext] Check the enum is fixed type before going to IRGen.

dcci:
<rdar://problem/50142178>

### DIFF
--- a/lit/SwiftREPL/FoundationTypes.test
+++ b/lit/SwiftREPL/FoundationTypes.test
@@ -41,3 +41,6 @@ n
 // CHECK:      {{.*R.*}}: Notification = {
 // CHECK-NEXT:   name = "abc"
 // CHECK-NEXT:   object = 42
+
+let tinky : Optional<UUID> = nil
+// CHECK: {{tinky}}: UUID? = nil
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -737,7 +737,8 @@ SwiftEnumDescriptor::CreateDescriptor(swift::ASTContext *ast,
       elements_with_payload = enum_impl_strategy.getElementsWithPayload();
   llvm::ArrayRef<swift::irgen::EnumImplStrategy::Element>
       elements_with_no_payload = enum_impl_strategy.getElementsWithNoPayload();
-  if (enum_decl->isResilient())
+  swift::SILType swift_sil_type = irgen_module.getLoweredType(swift_can_type);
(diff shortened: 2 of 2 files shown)


lit/SwiftREPL/FoundationTypes.test
```diff
@@ -41,3 +41,6 @@ n
 // CHECK:      {{.*R.*}}: Notification = {
 // CHECK-NEXT:   name = "abc"
 // CHECK-NEXT:   object = 42
+
+let tinky : Optional<UUID> = nil
```
dcci, line 45: I ended up finding a super easy test case, incidentally.

dcci:
@swift-ci test and merge

lit/SwiftREPL/FoundationTypes.test
adrian-prantl: This depends on UUID being resilient, but it's better than nothing.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -737,7 +737,8 @@ SwiftEnumDescriptor::CreateDescriptor(swift::ASTContext *ast,
       elements_with_payload = enum_impl_strategy.getElementsWithPayload();
   llvm::ArrayRef<swift::irgen::EnumImplStrategy::Element>
       elements_with_no_payload = enum_impl_strategy.getElementsWithNoPayload();
-  if (enum_decl->isResilient())
+  swift::SILType swift_sil_type = irgen_module.getLoweredType(swift_can_type);
+  if (!irgen_module.getTypeInfo(swift_sil_type).isFixedSize())
```
adrian-prantl, line 741: I would feel more comfortable with the change if we could move this case further down, since it's a bit of a catch-all. Are we sure all of the remaining cases have a fixed-size typeinfo?

lit/SwiftREPL/FoundationTypes.test
dcci: I can write something more elaborate and probably robust if you want, but my attempt to understand how to write a makefile failed (if you want to explain me the magic, happy to hear).

source/Symbol/SwiftASTContext.cpp
dcci: Should we put it as _first_ case instead? Going to IRGen without a fixed type is basically a recipe for trouble.

adrian-prantl: I agree with 

> Going to IRGen without a fixed type is basically a recipe for trouble.

but my question remains: does non-fixed-type always imply resilient?

PR #1532, opened 2019-05-02
Replace opaque types with their runtime underlying type.

jckarter:
rdar://problem/46138166

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1994,6 +1994,83 @@ SwiftLanguageRuntime::DoArchetypeBindingForType(StackFrame &stack_frame,
     // all conformances are always visible
     auto *module_decl = scratch_ctx->GetASTContext()->getStdlibModule();
 
+    // Replace opaque types with their underlying types when possible.
+    swift::Mangle::ASTMangler mangler(true);
+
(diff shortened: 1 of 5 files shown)


jckarter:
https://github.com/apple/swift/pull/24445

@swift-ci Please test

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1994,6 +1994,66 @@ SwiftLanguageRuntime::DoArchetypeBindingForType(StackFrame &stack_frame,
...
+          // Try to find the symbol for the opaque type descriptor in the
+          // process.
+          auto mangled_name = ConstString(
+                    mangler.mangleOpaqueTypeDescriptor(opaque_type->getDecl()));
+          
+          SymbolContextList found;
+          target.GetImages().FindSymbolsWithNameAndType(mangled_name,
+                                                        eSymbolTypeData, found);
+          
+          if (found.GetSize() != 1)
```
dcci: What if you have two symbols from two different `dylibs` which resolve with the same name or type? At that point, shouldn't you probably just check that their value agree and pick one instead of returning early?

jckarter:
apple/swift#24445

@swift-ci Please test

jckarter:
apple/swift#24445

@swift-ci Please test

PR #1534, opened 2019-05-03
Make lldb-mi optional and change how we deal with missing tools in lit

JDevlieghere:
We don't need the variables in lit, we can use the capabilities to check
if the utility exists.

Differential revision: https://reviews.llvm.org/D61533

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359926 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 097a7f74695bbd3a3256b391ab054d9bb37e06ce)


JDevlieghere:

PR #1536, opened 2019-05-03
Fix for ambiguous lookup in expressions between local variable and namespace

shafik:
Summary:
In an Objective-C context a local variable and namespace can cause an ambiguous name lookup when used in an expression. The solution involves mimicking the existing C++ solution which is to add local using declarations for local variables. This causes a different type of lookup to be used which eliminates the namespace during acceptable results filtering.

Differential Revision: https://reviews.llvm.org/D59960

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359921 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit f235e7581bb43fb891483f3bd51e4e35a83421ac)


shafik:
@swift-ci test

PR #1537, opened 2019-05-03
Set a CXXRecordDecl to not be passed in registers if DW_CC_pass_by_reference when loading from DWARF

shafik:
Summary:
This will fix a bug where during expression parsing we are not setting a CXXRecordDecl to not be passed in registers and the resulting code generation is wrong.
The DWARF attribute DW_CC_pass_by_reference tells us that we should not be passing in registers i.e. RAA_Indirect.
This change depends this clang change which fixes the fact that the ASTImporter does not copy RecordDeclBits for CXXRecordDecl: https://reviews.llvm.org/D61140

Differential Revision: https://reviews.llvm.org/D61146

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359732 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 4285f7f8db7300c72712db5cf08bd071ea3bf90a)

### DIFF
--- a/packages/Python/lldbsuite/test/expression_command/argument_passing_restrictions/Makefile
+++ b/packages/Python/lldbsuite/test/expression_command/argument_passing_restrictions/Makefile
@@ -1,0 +1,5 @@
+LEVEL = ../../make
+
+CXX_SOURCES := main.cpp
+
+include $(LEVEL)/Makefile.rules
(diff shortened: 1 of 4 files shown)


shafik:
@swift-ci test

PR #1538, opened 2019-05-03
[Alias] Add 're' alias for register

JDevlieghere:
This patch makes `re` an alias for `register`. Currently `re<TAB>` gives
you the choice between `register` and `reproducer`. Given that you use
`register` a lot more often, it should win for the common substring.

Differential revision: https://reviews.llvm.org/D61469

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@359927 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 39f7cfb4fed12f352b696e534972b8bcd333f10e)


JDevlieghere:

PR #1548, opened 2019-05-08
Unify quoting in log and error messages.

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #1549, opened 2019-05-08
 Improve logging when deserializing SDK paths and introduce an arbitat… 

adrian-prantl:
…ion rule.

rdar://problem/50488068


adrian-prantl:
@swift-ci test and merge

PR #1550, opened 2019-05-08
 Fix an elusive crash in LLDBMemoryReader::readBytes().

adrian-prantl:
Unfortunately I don't have an easy reproducer for this.

<rdar://problem/49973326>

### DIFF
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -1303,6 +1303,9 @@ public:
       auto addr = address.getAddressData();
       if (addr >= m_local_buffer &&
           addr + size <= m_local_buffer + m_local_buffer_size) {
+        // If this crashes, the assumptions stated in
+        // GetDynamicTypeAndAddress_Protocol() most likely no longer
+        // hold.
         memcpy(dest, (void *) addr, size);
         return true;
       }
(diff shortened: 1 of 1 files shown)


adrian-prantl:
@swift-ci test and merge

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1926,14 +1929,17 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Protocol(
   lldb::addr_t existential_address;
   bool use_local_buffer = false;
 
-  if (in_value.GetValueType() == eValueTypeConstResult) {
+  if (in_value.GetValueType() == eValueTypeConstResult &&
```
adrian-prantl, line 1932: This condition alone was just a proxy for what we actually wanted to test, which is in the next line.

dcci: I thought we were going to replace this with `DataExtractor`? Are you doing it as a follow up?

dcci:
```
/home/buildnode/jenkins/workspace/swift-lldb-PR-Linux/branch-swift-5.1-branch/lldb/source/Target/SwiftLanguageRuntime.cpp:1933:45: error: use of undeclared identifier 'eValueTypeHostAddress'
16:05:33       in_value.GetValue().GetValueType() == eValueTypeHostAddress) {
```
This probably needs to be fully qualified.

source/Target/SwiftLanguageRuntime.cpp
adrian-prantl: It turned out the the `else` code path is already reading memory from the process.

dcci: Very good, thanks.

dcci (approved):

adrian-prantl:
@swift-ci test and merge

PR #1558, opened 2019-05-10
[Swift] Refactor all module loader creation/addition to happen in GetASTContext()

nathawes:
This is necessary to be able to use the ClangImporter's clang::CompilerInstance to get the default clang module cache path to use in Swift's ParseableInterfaceModuleLoader which must be added before adding the ClangImporter. Previously they were created and added in separate methods
(GetASTContext() for the PIML and GetClangImporter() for the ClangImporter).

### DIFF
--- a/lit/Swift/DynamicTyperesolutionConflict.test
+++ b/lit/Swift/DynamicTyperesolutionConflict.test
@@ -1,4 +1,5 @@
 # REQUIRES: system-darwin
+# REQUIRES: rdar50667488
 
 # This testcase causes the scratch context to get destroyed by a
 # conflict that is triggered via dynamic type resolution. The conflict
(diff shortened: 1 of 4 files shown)


nathawes:
@swift-ci please test

nathawes:
@swift-ci please test

adrian-prantl (approved):

nathawes:
@swift-ci please test

lit/Swift/DynamicTyperesolutionConflict.test
```diff
@@ -1,4 +1,5 @@
 # REQUIRES: system-darwin
+# REQUIRES: rdar50667488
```
nathawes, line 2: Adrian is going to look into this failure later.

PR #1560, opened 2019-05-10
[5.1][Swift] Refactor all module loader creation/addition to happen in GetASTContext()

nathawes:
Cherry-pick of https://github.com/apple/swift-lldb/pull/1558 for 5.1

This is necessary to be able to use the ClangImporter's clang::CompilerInstance to get the default clang module cache path to use in Swift's ParseableInterfaceModuleLoader which must be added before adding the ClangImporter. Previously they were created and added in separate methods
(GetASTContext() for the PIML and GetClangImporter() for the ClangImporter).

Resolves rdar://problem/50163552

### DIFF
--- a/lit/Swift/DynamicTyperesolutionConflict.test
+++ b/lit/Swift/DynamicTyperesolutionConflict.test
@@ -1,4 +1,5 @@
 # REQUIRES: system-darwin
+# REQUIRES: rdar50667488
 
 # This testcase causes the scratch context to get destroyed by a
 # conflict that is triggered via dynamic type resolution. The conflict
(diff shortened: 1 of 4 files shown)


nathawes:
@swift-ci please test

adrian-prantl (approved):

lit/Swift/DynamicTyperesolutionConflict.test
```diff
@@ -1,4 +1,5 @@
 # REQUIRES: system-darwin
+# REQUIRES: rdar50667488
```
adrian-prantl, line 2: I will look into this later.

PR #1561, opened 2019-05-10
[lldb] Fix merge

smeenai:
- https://reviews.llvm.org/D59040 deleted source/Expression/ExpressionSourceCode.cpp
  entirely, so we should delete everything except the Swift-specific parts.
- https://reviews.llvm.org/D61501 renamed CompileUnit to Unit, so we should
  complete that rename across swift-lldb.


smeenai:
CC @dcci @JDevlieghere @xiaobai @compnerd

PR #1562, opened 2019-05-10
[Reproducers] Capture return values of functions returning by ptr/ref

JDevlieghere:
For some reason I had convinced myself that functions returning by
pointer or reference do not require recording their result. However,
after further considering I don't see how that could work, at least not
with the current implementation. Interestingly enough, the reproducer
instrumentation already (mostly) accounts for this, though the
lldb-instr tool did not.

This patch adds the missing macros and updates the lldb-instr tool.

Differential revision: https://reviews.llvm.org/D60178

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@357639 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit ef3f3c35ca8aa68ba20cff357a513e45295231b6)


JDevlieghere:
@swift-ci please test

PR #1565, opened 2019-05-13
Merge target and launch info environments

JDevlieghere:
Before this change we were overriding the launch info environment with
the target environment. This meant that the environment variables passed
to `process launch --environment <>` were lost. Instead of replacing the
environment, we should merge them.

Differential revision: https://reviews.llvm.org/D61864

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@360612 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 1467977362168cda2dc4cd75f65c50433621c90b)

### DIFF
--- a/lit/Process/Inputs/env.cpp
+++ b/lit/Process/Inputs/env.cpp
@@ -1,0 +1,7 @@
+#include <cstdlib>
+#include <iostream>
+
+int main() {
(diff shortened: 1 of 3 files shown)


JDevlieghere:

PR #1570, opened 2019-05-13
[CMake] Simplify lldb-server handling

JDevlieghere:
We can piggyback off the existing add_lldb_tool_subdirectory to decide
whether or not lldb-server should be built.

Differential revision: https://reviews.llvm.org/D61872

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@360621 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit e2a0d76f51ce8ca5940c348fde5030e3f83c198d)


JDevlieghere:
@swift-ci please test

PR #1578, opened 2019-05-14
Finish renaming CompileUnit -> Unit

gmittert:
I believe this patch was dropped from upstream-with-swift in one of the merges. Cherry-picking it back in. 
```
D42892 changed a lot of code to use superclass DWARFUnit instead of its
subclass DWARFCompileUnit.

Finish this change more thoroughly for any *CompileUnit* -> *Unit* names.
Later patch will introduce DWARFTypeUnit which needs to be sometimes different
from DWARFCompileUnit and it would be confusing without this renaming.

Differential Revision: https://reviews.llvm.org/D61501

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@360443 91177308-0d34-0410-b5e6-96231b3b80d8
```

### DIFF
--- a/source/Plugins/SymbolFile/DWARF/ManualDWARFIndex.cpp
+++ b/source/Plugins/SymbolFile/DWARF/ManualDWARFIndex.cpp
@@ -32,9 +32,9 @@ void ManualDWARFIndex::Index() {
   Timer scoped_timer(func_cat, "%p", static_cast<void *>(&debug_info));
 
   std::vector<DWARFUnit *> units_to_index;
-  units_to_index.reserve(debug_info.GetNumCompileUnits());
-  for (size_t U = 0; U < debug_info.GetNumCompileUnits(); ++U) {
-    DWARFUnit *unit = debug_info.GetCompileUnitAtIndex(U);
+  units_to_index.reserve(debug_info.GetNumUnits());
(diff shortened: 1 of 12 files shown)


source/Plugins/SymbolFile/DWARF/ManualDWARFIndex.cpp
```diff
@@ -55,22 +55,24 @@ void ManualDWARFIndex::Index() {
...
+  // First figure out which units didn't have their DIEs already
   // parsed and remember this.  If no DIEs were parsed prior to this index
   // function call, we are going to want to clear the CU dies after we are
   // done indexing to make sure we don't pull in all DWARF dies, but we need
-  // to wait until all compile units have been indexed in case a DIE in one
-  // compile unit refers to another and the indexes accesses those DIEs.
-  //----------------------------------------------------------------------
+  // to wait until all units have been indexed in case a DIE in one
+  // unit refers to another and the indexes accesses those DIEs.
+#if 0
```
gmittert, line 66: `#if 0`ing this out was the only change I needed to make to have this merge successfully.

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

gmittert:
I believe this is superceded by https://github.com/apple/swift-lldb/pull/1580

PR #1580, opened 2019-05-15
Merge llvm.org into upstream-with-swift (May 14t)

dcci:
cc: @xiaobai (for some reason I can't put you in the reviewers section)


adrian-prantl:
If I caught all your edits this LGTM.

PR #1581, opened 2019-05-15
Clang's ASTImport now returns Expected<> Types

gmittert:
Clang b211b92b88 changed Import to return an Expected<QualType> so
updated the usage to match

This should fix the upstream-with-swift build.
https://github.com/llvm-mirror/clang/commit/b211b92b8811a6f926a11912ca97291d56486d2a

Clang was changed here:
git-svn-id: https://llvm.org/svn/llvm-project/cfe/trunk@360760
91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/source/Symbol/ClangASTContext.cpp
+++ b/source/Symbol/ClangASTContext.cpp
@@ -1437,8 +1437,13 @@ CompilerType ClangASTContext::CopyType(const CompilerType &src) {
           FileManager file_manager(file_system_options);
           ASTImporter importer(*dst_clang_ast, file_manager, *src_clang_ast,
                                file_manager, false);
-          QualType dst_qual_type(importer.Import(ClangUtil::GetQualType(src)));
-          return CompilerType(this, dst_qual_type.getAsOpaquePtr());
+          if (auto expected = importer.Import(ClangUtil::GetQualType(src))) {
+            QualType dst_qual_type(expected.get());
+            return CompilerType(this, dst_qual_type.getAsOpaquePtr());
+          } else {
+              Log *log = lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_EXPRESSIONS);
+              LLDB_LOG_ERROR(log, expected.takeError(), "Couldn't import type: {0}");
+          }
         }
       }
     }


source/Symbol/ClangASTContext.cpp
```diff
@@ -1437,8 +1437,11 @@ CompilerType ClangASTContext::CopyType(const CompilerType &src) {
           FileManager file_manager(file_system_options);
           ASTImporter importer(*dst_clang_ast, file_manager, *src_clang_ast,
                                file_manager, false);
-          QualType dst_qual_type(importer.Import(ClangUtil::GetQualType(src)));
-          return CompilerType(this, dst_qual_type.getAsOpaquePtr());
+          Expected<QualType> imported_dst_qual_type = importer.Import(ClangUtil::GetQualType(src));
+          if (imported_dst_qual_type) {
```
compnerd: I would fold this into one:

`if (auto T = importer.import(ClangUtil::GetQualType(src))`

compnerd (commented):
Please add a reference to the SVN revision in the commit message.

gmittert:
Don't merge this quite yet, the swift-clang auto rebaser hasn't pulled the corresponding patch into swift-clang yet.

Edit: it's now in swift-clang upstream with swift.

gmittert:
@swift-ci please test

gmittert:
Do the tests actually work on upstream-with-swift PRs? The seem to keep failing on the same thing: something with swig on Linux, and something in swift on macos.

gmittert:
CC: @dcci @adrian-prantl @xiaobai @JDevlieghere

dcci:
@swift-ci test

source/Symbol/ClangASTContext.cpp
```diff
@@ -1437,8 +1437,10 @@ CompilerType ClangASTContext::CopyType(const CompilerType &src) {
           FileManager file_manager(file_system_options);
           ASTImporter importer(*dst_clang_ast, file_manager, *src_clang_ast,
                                file_manager, false);
-          QualType dst_qual_type(importer.Import(ClangUtil::GetQualType(src)));
-          return CompilerType(this, dst_qual_type.getAsOpaquePtr());
+          if (auto expected = importer.Import(ClangUtil::GetQualType(src))) {
+            QualType dst_qual_type(expected.get());
+            return CompilerType(this, dst_qual_type.getAsOpaquePtr());
+          }
```
JDevlieghere, line 1446: I believe this will assert in the destructor of llvm::Error when the expected is empty. You'd have to do something like   in CopyDecl: 

```
if (auto expected = importer.Import(ClangUtil::GetQualType(src))) {
  ...
} else { 
  Log *log = lldb_private::GetLogIfAllCategoriesSet(LIBLLDB_LOG_EXPRESSIONS);
  LLDB_LOG_ERROR(log, expected.takeError(), "Couldn't import type: {0}");	
}

```

JDevlieghere (changes requested):

dcci:
> Do the tests actually work on upstream-with-swift PRs? The seem to keep failing on the same thing: something with swig on Linux, and something in swift on macos.

`upstream-with-swift` is a best-effort branch. Things aren't guaranteed to always work there.
This is true for `master-next` in swift too.

gmittert:
> > Do the tests actually work on upstream-with-swift PRs? The seem to keep failing on the same thing: something with swig on Linux, and something in swift on macos.
> 
> `upstream-with-swift` is a best-effort branch. Things aren't guaranteed to always work there.
> This is true for `master-next` in swift too.

Does that then mean it's fine to merge even if builds are failing?

source/Symbol/ClangASTContext.cpp
gmittert: I can add the logging, but calling operator bool() on it should mark it as checked if it's empty

https://llvm.org/doxygen/Support_2Error_8h_source.html#l00235

gmittert:
Clang issue should now be fixed,

@swift-ci please test

gmittert:
@swift-ci please test

source/Symbol/ClangASTContext.cpp
JDevlieghere: Yup, you are correct. I think the logging here is good, but otherwise it's customary to have an [`llvm::consumeError`](https://llvm.org/doxygen/namespacellvm.html#aee0e70be66536b9ef7b65ae2e5a45959) when you want to explicitly discarding the error.

PR #1585, opened 2019-05-15
[Initialization] Remove swift initialization from SystemInitializerCommon

bulbazord:
Swift initialization should occur in SystemInitializaterFull and not in the SystemInitializerCommon.
This PR has 3 changes:
 - I found that SwiftASTContext was being initialized in both Full and Common, which seems problematic. Removed the one in Common.
 - Moved SwiftREPL initialization from Common to Full.
 - Stopped linking lldbInitialization against the swift ExpressionParser plugin. This is good because it means lldb-server stops linking against it.

### DIFF
--- a/source/API/SystemInitializerFull.cpp
+++ b/source/API/SystemInitializerFull.cpp
@@ -113,6 +113,7 @@
 #endif
 
 #if defined(__APPLE__) || defined(__linux__) || defined(_WIN32)
+#include "Plugins/ExpressionParser/Swift/SwiftREPL.h"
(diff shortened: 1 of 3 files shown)


bulbazord:
cc @compnerd @dcci @JDevlieghere

dcci (approved):
Seems fine to me but I would be happy if @JDevlieghere could take a look as he understand the initialization path better than me.

dcci:
@swift-ci test

compnerd (approved):
Thanks for doing this.

PR #1593, opened 2019-05-16
Remove SwiftASTContext::isTrivialOptionSetType. (NFC)

adrian-prantl:
To serve the SwiftOptionSet data formatter, SwiftASTContext had an
elaborate mechanism in place to determine whether the type is
"trivial" or not. Nowadays, for all practical intents and purposes
quering for the existence of the rawValue child, appears to have the
same effect and was done by the data formatter anyway.

rdar://problem/50735738

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -377,20 +377,6 @@ public:
 
   Status GetFatalErrors();
 
-  union ExtraTypeInformation {
-    uint64_t m_intValue;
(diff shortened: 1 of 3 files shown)


adrian-prantl:
@swift-ci test

adrian-prantl:
Also, it looks like the Swift type system makes it impossible to define a "nontrivial" option set since OptionSet conforms to RawRepresentable, which expects a rawValue property.

dcci (approved):
lgtm, we discussed this extensively in private.

PR #1594, opened 2019-05-16
[StackFrame] Prefer synthetic variable if asked.

dcci:
If a type has a synthetic we should give it preference.
I believe this works fine for most of the types anwyay,
but integers are "special", as lldb treats them both
as primitive types or structures depending on the code
path in the data formatters code. Eventually we should
just have a consistent reprensentation for integers and
floats (i.e. represent them for what they really are, structs).
But until that happens, we need to have this hack in place.

Fixes:
(lldb) frame var patatino[0]
(_ArrayBuffer<Int>) patatino[0] = {
  _storage = (rawValue = 4 values)
}

to print:
(lldb) frame var patatino[0]
(Int) [0] = 1

<rdar://problem/50705707>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/array_element/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/array_element/Makefile
@@ -1,0 +1,3 @@
+LEVEL = ../../../make
+SWIFT_SOURCES := main.swift
+include $(LEVEL)/Makefile.rules
--- a/packages/Python/lldbsuite/test/lang/swift/array_element/TestSwiftArrayElement.py
+++ b/packages/Python/lldbsuite/test/lang/swift/array_element/TestSwiftArrayElement.py
@@ -1,0 +1,4 @@
+import lldbsuite.test.lldbinline as lldbinline
+from lldbsuite.test.decorators import *
+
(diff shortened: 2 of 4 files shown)


dcci:
@swift-ci test

jimingham:
This looks right.  GetChildAtIndex SHOULD do exactly what the code you've added does, but given Ints are weird it should not hurt to spell it out explicitly.  

I would like a slightly better comment than "SWIFT MOD", however, because if you actually understand how the ValueObject system works, this code should NOT be necessary and is pretty confusing.  Even just "This is a workaround for value providing synthetic children like the one used for Swift Int."  Then somebody reading this code would either know where to look for what's going on, or to put their fingers in their ears and say "Nah, nah, nah, I can't hear you".

dcci:
@swift-ci test and merge

jimingham (changes requested):
That's not quite right.  These are "value-providing synthetic children" not "values which provide synthetic children" which is what the way you've written it makes it mean.

dcci:
@jimingham so you want "value(s) providing" to be "value-providing" ?
I can change that, no problem.

dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

PR #1595, opened 2019-05-17
Add a test for an OptionSet with a computed rawValue property.

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

jrose-apple:
Did you want a non-trivial example, though? Like a struct with a class reference or String in it?

PR #1596, opened 2019-05-17
[Swift] Accommodate the change to multiple runtime library paths.

mikeash:
LLDB change corresponding to https://github.com/apple/swift/pull/24838.

rdar://problem/47528005


dcci (approved):
LGTM. We might want to also upstream the changes for the CMakeLists.txt in `tools/repl` so that the RPATH is always consistent with what we have here.

mikeash:
https://github.com/apple/swift/pull/24838

@swift-ci please test

PR #1598, opened 2019-05-17
Merge stable into upstream-with-swift

dcci:
@xiaobai FYI

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -377,20 +377,6 @@ public:
 
   Status GetFatalErrors();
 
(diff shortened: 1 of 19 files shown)


dcci:
@xiaobai now that I look at this I feel like I resolved one of these incorrectly, as you removed the initialization for ClangASTContext in SystemInitializerCommon. Let me double check.

bulbazord:
@dcci Yeah, ClangASTContext should no longer be in SystemInitializerCommon. That change is in both llvm/master and swift-lldb/stable, so after you resolve that here it shouldn't conflict again.

dcci:
Done, thanks.

PR #1601, opened 2019-05-17
[Test] Be less stingent about whitespace in SwiftREPL tests

JDevlieghere:
Be more permissive when it comes to whitespace. Currently, if the regex
matches nothing, we still expect two spaces.


JDevlieghere:

PR #1603, opened 2019-05-18
[crashlog] Use loads() instead of readPlistFromString() for python 3.

dcci:
<rdar://problem/50903413>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@361087 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/examples/python/crashlog.py
+++ b/examples/python/crashlog.py
@@ -43,6 +43,12 @@ import sys
 import time
 import uuid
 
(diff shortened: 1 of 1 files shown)


dcci:
@swift-ci test

JDevlieghere:
I would’ve wrapped it in a try-except but this works so LGTM

PR #1607, opened 2019-05-21
[SwiftASTContext] Remove use of ClusteredBitVector::enumerateSetBits

mundaym:
The set bit enumeration API is being removed as part of
https://github.com/apple/swift/pull/24857.

This change replaces the enumerateSetBits call with a loop over all
the bits in the mask. This is potentially more expensive however it's
unlikely that the difference is noticeable given that we are already
building a mask of the same size byte-by-byte in this function.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -588,14 +588,15 @@ public:
       return nullptr;
     }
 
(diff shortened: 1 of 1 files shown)


jckarter:

https://github.com/apple/swift/pull/24857

@swift-ci Please test

PR #1610, opened 2019-05-21
Windows: define `ComputeSwiftDirectory` for Windows

compnerd:
Assume that the Windows layout is relatively similar to that on Linux.
`lldb.dll` is located in the `bin` directory rather than `lib` and the
`ComputeSupportExeDirectory` provides a `FileSpec` that is off by a
level.  Using `ComputeSupportExeDirectory` with `..\lib\swift` gives us
the actual location of the resource dir when lldb is installed into a
toolchain (as per the MSIs from
https://compnerd.visualstudio.com/windows-swift).

### DIFF
--- a/include/lldb/Host/windows/HostInfoWindows.h
+++ b/include/lldb/Host/windows/HostInfoWindows.h
@@ -36,6 +36,7 @@ public:
   static FileSpec GetProgramFileSpec();
   static FileSpec GetDefaultShell();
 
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @dcci @kusmour

This at least permits the repl to startup on Windows; there is still an issue with IRGen that needs to be investigated.

compnerd:
@kusmour took care of this I believe and this should no longer be needed

PR #1616, opened 2019-05-22
[CommandInterpreter] Accept blanks after `all` or [0-9]+ for bt.

dcci:
Previously "bt all    " would've failed as the regex didn't match
them.

<rdar://problem/50824935>


dcci:
@swift-ci test

JDevlieghere (approved):

PR #1618, opened 2019-05-22
Merge llvm.org(master) into upstream-with-swift

dcci:


dcci:
@xiaobai Alex, I'm under the impression your locking change should be on `stable` too. Do you agree?
On top, we should have a small modification in `swift-lldb` to make sure also the SwiftLanguageRuntime is locked. ACK/NACK? =)

dcci:
[happy to do the work, just trying to collect your opinion here]

bulbazord:
@dcci Yeah putting it on stable is a good idea I think. I don't understand what you mean by "make sure also the SwiftLanguageRuntime is locked". Does Process have its own field for a SwiftLanguageRuntime?

dcci:
@xiaobai What I mean is, we should acquire a mutex in `GetSwiftLanguageRuntime` as we do in `GetCPPLanguageRuntime` and `GetObjCLanguageRuntime` right?

bulbazord:
@dcci: Ah yes, I think that's the right thing to do.

PR #1619, opened 2019-05-22
Expression: correct relocation model for Windows

compnerd:
The Windows Code Generation model cannot generation code with the PIC relocation
model - all code is implicitly position independent due to the DLL load slide
that occurs if it is not loaded at the preferred base address.  Invert the
condition and inline the single use of the variable.  This should also aid the
WASM target.  This significantly improves the state of the (swift) repl on
Windows (and should aid in expression evaluation on Windows).

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@361443 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 3b0388e3746bc88193da27092e837cb099188d29)


compnerd:
CC: @JDevlieghere 

This enables the repl to begin functioning on Windows

compnerd:
@swift-ci please test

PR #1620, opened 2019-05-23
[CommandInterpreter] Refactor SourceInitFile

JDevlieghere:
I was looking at the current implementation of SourceInitFile and there
were a few things that made this function hard to read:

 * The code to find the ~/.lldbinit file is duplicated across the cwd
   and non-cwd branch.
 * The ./.lldbinit is once computed by resolving .lldbinit and once by
   resolving ./.lldbinit.
 * It wasn't clear to me what happened when you're sourcing the
   .lldbinit file in the current working directory. Apparently we do
   nothing when we property to control that is set to warn (makes sense)
   and we don't care when the property is set to true (debatable).
 * There were at least two branches where the status of the
   CommandReturnObject were not set.

This patch attempts to simplify that code.

Differential revision: https://reviews.llvm.org/D61994

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@361080 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 0052c40ce2b39d53d8a6cee9f8265871128eaae3)

### DIFF
--- a/include/lldb/Interpreter/CommandInterpreter.h
+++ b/include/lldb/Interpreter/CommandInterpreter.h
@@ -215,7 +215,8 @@ public:
     return GetStaticBroadcasterClass();
   }
 
-  void SourceInitFile(bool in_cwd, CommandReturnObject &result);
+  void SourceInitFileCwd(CommandReturnObject &result);
+  void SourceInitFileHome(CommandReturnObject &result);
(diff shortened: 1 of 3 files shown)


JDevlieghere:
@swift-ci please test

PR #1623, opened 2019-05-23
Fix upstream-with-swift merge

as106:
Summary: Fixed merge conflict in CMakeLists.txt.

cc: @xiaobai @compnerd @dcci @JDevlieghere


bulbazord:
I think this is obviously right. There are other issues with SymbolFileDWARF but those look a little messier to deal with and might take a little more time.

bulbazord:
I stand corrected, looks like #1624 fixed my other concern. This coulda been one PR but no big deal.

PR #1626, opened 2019-05-23
[Test] Be less stingent about whitespace in SwiftREPL tests

JDevlieghere:
Be more permissive when it comes to whitespace. Currently, if the regex
matches nothing, we still expect two spaces.

(cherry picked from commit d8b00d5a370c0aa4548681e3ae65fa7821e888a9)


JDevlieghere:

PR #1627, opened 2019-05-23
 Clean up triple handling in SwiftASTContext (NFC)

adrian-prantl:
This patch

- replaces all char* and std::string triples in public interfaces with
  llvm::Triple objects

- Avoids initializing module contexts with
  SwiftASTContextForExpression objects

- Fixes a bug in log description of regular SwiftASTContext objects

rdar://problem/50894825

PR #1632, opened 2019-05-28
 Fix several assertion failures when dealing with corrupted/missing Sw…

adrian-prantl:
Fix several assertion failures when dealing with corrupted/missing Swift modules.

<rdar://problem/51049172>

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -920,11 +920,5 @@ private:
   std::unique_ptr<SwiftPersistentExpressionState> m_persistent_state_up;
 };
 
-void printASTValidationInfo(
-    const swift::serialization::ValidationInfo &ast_info,
-    const swift::serialization::ExtendedValidationInfo &ext_ast_info,
-    const Module &module, llvm::StringRef module_buf);
-
-}
-
+} // namespace lldb_private
 #endif // #ifndef liblldb_SwiftASTContext_h_
--- a/packages/Python/lldbsuite/test/lang/swift/deserialization_failure/TestSwiftDeserializationFailure.py
+++ b/packages/Python/lldbsuite/test/lang/swift/deserialization_failure/TestSwiftDeserializationFailure.py
@@ -1,0 +1,54 @@
+import lldb
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test.decorators import *
+import lldbsuite.test.lldbutil as lldbutil
+import unittest2
+
+
+class TestSwiftDeserializationFailure(TestBase):
+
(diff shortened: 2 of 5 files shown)


adrian-prantl:
@swift-ci test and merge

include/lldb/Symbol/SwiftASTContext.h
```diff
@@ -920,11 +918,5 @@ class SwiftASTContextForExpressions : public SwiftASTContext {
   std::unique_ptr<SwiftPersistentExpressionState> m_persistent_state_up;
 };
 
-void printASTValidationInfo(
```
dcci, line 923: The definition of this function was just missing?

packages/Python/lldbsuite/test/lang/swift/deserialization_failure/TestSwiftDeserializationFailure.py
```diff
@@ -0,0 +1,54 @@
...
+        # We should not be able to resolve the types defined in the module.
+        lldbutil.continue_to_breakpoint(process, dynamic_bkpt)
+        self.expect("fr var c", substrs=["<could not resolve type>"])
+
+        lldbutil.continue_to_breakpoint(process, generic_bkpt)
+        # FIXME: this is formatted incorrectly.
+        self.expect("fr var t", substrs=["(T)"]) #, "world"])
+
+    @skipUnlessDarwin
+    @skipIf(debug_info=no_match(["dwarf"]))
```
dcci, line 39: Why is this limited to `dwarf` only?

include/lldb/Symbol/SwiftASTContext.h
dcci: Nevermind, the UI didn't load the changes in `SwiftASTContext.cpp`

packages/Python/lldbsuite/test/lang/swift/deserialization_failure/TestSwiftDeserializationFailure.py
adrian-prantl: I was going to say because I'm manually deleting the .swiftmodule so the dsym case is identical, but for the case with the damaged module it would make sense to test both code paths. Will change this.

include/lldb/Symbol/SwiftASTContext.h
adrian-prantl: I moved the function closer to its use in the file.

adrian-prantl: Sorry, most of the changes are actually https://github.com/apple/swift-lldb/pull/1627

adrian-prantl:
@swift-ci test and merge

PR #1633, opened 2019-05-29
[DWARF] Fix Swift build after r361849

smeenai:
r361849 removed the one-argument constructor to DWARFExpression. In this
case, we can just switch to the default constructor, since the
constructed object is being used with GetDIENamesAndRanges, which has
been updated to fill in the CU itself.


smeenai:
CC @compnerd @xiaobai @dcci @JDevlieghere. This should be merged after r361849 has made its way to upstream-with-swift (or integrated into that merge).

dcci:
@smeenai assuming the automerger works, it's safe to merge this now.

bulbazord:
Yup LGTM

PR #1640, opened 2019-05-29
[SwiftASTContext] Make sure search opts and clang importer are initia…

dcci:
…lized.

Some codepaths, e.g. CompleteCode, try to grab a SwiftASTContext without
these initialized and crash later with an assertion (or, even worse, in release
mode, segfaults when trying to access these fields).

<rdar://problem/50875178>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
+++ b/packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
@@ -1,0 +1,21 @@
+from __future__ import print_function
+import pexpect
+import os
+import lldb
+from lldbsuite.test.decorators import *
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test import lldbutil
+
+
+class TestSwiftREPLCompletion(TestBase):
+    mydir = TestBase.compute_mydir(__file__)
+
+    @skipUnlessDarwin
+    def test_repl_completion(self):
+        prompt = "Welcome to"
+        child = pexpect.spawn('%s --repl' % (lldbtest_config.lldbExec))
+        # Assign to make sure the sessions are killed during teardown
+        self.child = child
+        # Send a <TAB> and make sure we don't crash.
+        child.sendline("import Foundation\t")
+        child.expect("2>")
--- a/packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
+++ b/packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
@@ -17,5 +17,7 @@ class TestSwiftREPLCompletion(TestBase):
         # Assign to make sure the sessions are killed during teardown
         self.child = child
         # Send a <TAB> and make sure we don't crash.
-        child.sendline("import Foundation\t")
+        child.sendline("import Foundatio\t")
         child.expect("2>")
+        child.sendline("print(NSString(\"patatino\"))")
+        child.expect("patatino")
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -864,14 +864,31 @@ SwiftASTContext::SwiftASTContext(const SwiftASTContext &rhs)
 
   m_platform_sdk_path = rhs.m_platform_sdk_path;
 
+  // Initialize search paths and clang importer options, we need
+  // them to grab a SwiftASTContext without asserting.
+  std::vector<std::string> module_search_paths;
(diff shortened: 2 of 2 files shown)


packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
```diff
@@ -0,0 +1,21 @@
...
+    mydir = TestBase.compute_mydir(__file__)
+
+    @skipUnlessDarwin
+    def test_repl_completion(self):
+        prompt = "Welcome to"
+        child = pexpect.spawn('%s --repl' % (lldbtest_config.lldbExec))
+        # Assign to make sure the sessions are killed during teardown
+        self.child = child
+        # Send a <TAB> and make sure we don't crash.
+        child.sendline("import Foundation\t")
```
adrian-prantl: what about sending
print(NSString("ok"))
afterwards and expecting "ok"?

adrian-prantl: Does Linux Foundation have NSString?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -864,14 +864,31 @@ SwiftASTContext::SwiftASTContext(const SwiftASTContext &rhs)
...
-  }
-  GetClangImporter();
+  // As this is a copy constructor, make sure we copy the search path
+  // opts and clang importer options from RHS to LHS.
+  assert(lhs_ast && "Invalid LHS ASTContext");
+  assert(rhs_ast && "Invalid RHS ASTContext");
+  lhs_ast->SearchPathOpts = rhs_ast->SearchPathOpts;
+
+  GetCompilerInvocation().getClangImporterOptions() =
+      rhs.m_compiler_invocation_ap->getClangImporterOptions();
```
adrian-prantl, line 883: Isn't this redundant with the call to InitializeSearchPathOptions() and  GetClangImporterOptions() above?

packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
adrian-prantl: How about child.sendline("import Foundatio\t") ?

source/Symbol/SwiftASTContext.cpp
dcci: Only the copy of search paths, the second one is not because before we just initialise the importer options, we don't copy from RHS.

packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
dcci: Good point.

dcci:
@fredriss I'm merging this to 5.1 FYI.

dcci:
@swift-ci test and merge

packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
```diff
@@ -0,0 +1,23 @@
...
+import lldb
+from lldbsuite.test.decorators import *
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test import lldbutil
+
+
+class TestSwiftREPLCompletion(TestBase):
+    mydir = TestBase.compute_mydir(__file__)
+
+    @skipUnlessDarwin
```
adrian-prantl, line 13: Is this necessary?

adrian-prantl (approved):

packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
```diff
@@ -0,0 +1,23 @@
...
+    def test_repl_completion(self):
+        prompt = "Welcome to"
+        child = pexpect.spawn('%s --repl' % (lldbtest_config.lldbExec))
+        # Assign to make sure the sessions are killed during teardown
+        self.child = child
+        # Send a <TAB> and make sure we don't crash.
+        child.sendline("import Foundatio\t")
+        child.expect("2>")
+        child.sendline("print(NSString(\"patatino\"))")
```
adrian-prantl, line 22: child.sendline('print(NSString("patatino"))')

packages/Python/lldbsuite/test/lang/swift/completion/TestSwiftREPLCompletion.py
dcci: I don't think so, let me remove it.

PR #1644, opened 2019-05-30
Add a target.swift-extra-clang-flags setting for ClangImporter flags.

adrian-prantl:
This is intended as a power-user workaround for ClangImporter troubles
caused by non-modular system or 3rd-party headers that are outside of
the control of the user. See the TestSwiftExtraClangFlags for an
example installing a module map using a VFS overlay.

rdar://problem/51272979


adrian-prantl:
@swift-ci test and merge

PR #1646, opened 2019-05-30
[Target] Generalize language-specific behavior in ThreadPlanStepThrough

bulbazord:
This is a cherry-pick from llvm.org (rL362164) and an accompanying commit on top of that to accomodate for swift.


compnerd:
@swift-ci please test

PR #1647, opened 2019-05-31
[5.1][Swift] Don't test the 'prefer-parseable' and 'only-parseable' modes in SwiftInterfaceForceModuleLoadMode.test

nathawes:
These trigger loading the Swift standard library module via its module interface, which takes quite a while and is causing timeouts in CI (e.g. https://ci.swift.org/view/LLDB/job/oss-lldb-incremental-osx-cmake/4500/consoleText). The tests for the prefer-serializable and only-serializable values ensure the module-load-mode setting works and threads the mode though to the
module loaders, so this doesn't lose us too much.

I tried using the `-parse-stdlib` flag we use on the compiler side to prevent loading the stdlib, but lldb seems to load it anyway.


nathawes:
@swift-ci please test

jrose-apple:
Meh, okay.

PR #1657, opened 2019-06-04
Add support for mid-function epilogues on x86 that end in a non-local jump.

jasonmolenda:
Add support for mid-function epilogues on x86 that end in a non-local jump.
    
    The x86 assembly inspection engine has code to support detecting a
    mid-function epilogue that ends in a RET instruction; add support for
    recognizing an epilogue that ends in a JMP, and add a check that the
    unwind state has been restored to the original stack setup; reinstate
    the post-prologue unwind state after this JMP instruction.
    
    The assembly inspection engine used for other architectures,
    UnwindAssemblyInstEmulation, detects mid-function epilogues by
    tracking branch instructions within the function and "forwards"
    the current unwind state to the targets of the branches.  If
    an epilogue unwinds the stack and exits, followed by a branch
    target, we get back to the correct unwind state.  The x86
    unwinder should move to this same algorithm, or possibly even
    look at implementing an x86 instruction emulation plugin and
    get UnwindAssemblyInstEmulation to work for x86 too.  I added
    a branch instruction recognizier method that will be necessary
    if we want to switch the algorithm.
    
    Differential Revision: https://reviews.llvm.org/D62764
    <rdar://problem/51074422>

### DIFF
--- a/source/Plugins/UnwindAssembly/x86/x86AssemblyInspectionEngine.cpp
+++ b/source/Plugins/UnwindAssembly/x86/x86AssemblyInspectionEngine.cpp
@@ -667,10 +667,207 @@ bool x86AssemblyInspectionEngine::mov_reg_to_local_stack_frame_p(
   return false;
 }
 
+// Returns true if this is a jmp instruction where we can't
+// know the destination address statically. 
+//
+// ff e0                                   jmpq   *%rax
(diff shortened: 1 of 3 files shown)


jasonmolenda:
@swift-ci test and merge

PR #1660, opened 2019-06-04
[Target] Generalize some behavior in Target::SymbolsDidLoad

bulbazord:
Summary:
SymbolsDidLoad is currently only implemented for ObjCLanguageRuntime,
but that doesn't mean that it couldn't be useful for other Langauges. Although
this change seems like it's generalizing for the sake of purity, this removes
Target's dependency on ObjCLanguageRuntime.

Differential Revision: https://reviews.llvm.org/D62796

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@362461 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/include/lldb/Target/LanguageRuntime.h
+++ b/include/lldb/Target/LanguageRuntime.h
@@ -160,6 +160,8 @@ public:
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


compnerd:

PR #1661, opened 2019-06-04
[Target] Remove Process::GetCPPLanguageRuntime

bulbazord:
Summary:
I want to remove this method because I think that Process should be
language agnostic, or at least, not have knowledge about specific language
runtimes. There is "GetLanguageRuntime()" which should be used instead. If the
caller a CPPLanguageRuntime, they should cast it as needed. Ideally, this
should only happen in plugins that need C++ specific knowledge.

The next step I would like to do is remove "GetObjCLanguageRuntime()" as well.
There are a lot more instances of that function being used, so I wanted to
upload this one first to get the general reception to this idea.

Reviewers: compnerd, davide, JDevlieghere, jingham, clayborg, labath, aprantl

Subscribers: lldb-commits

Differential Revision: https://reviews.llvm.org/D62755

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@362544 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/include/lldb/Target/CPPLanguageRuntime.h
+++ b/include/lldb/Target/CPPLanguageRuntime.h
@@ -44,6 +44,11 @@ public:
     return lldb::eLanguageTypeC_plus_plus;
   }
 
+  static CPPLanguageRuntime *GetCPPLanguageRuntime(Process &process) {
+    return static_cast<CPPLanguageRuntime *>(
+        process.GetLanguageRuntime(lldb::eLanguageTypeC_plus_plus));
+  }
+
   virtual bool IsVTableName(const char *name) = 0;
 
   bool GetObjectDescription(Stream &str, ValueObject &object) override;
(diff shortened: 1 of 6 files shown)


compnerd:
This is a cherry-pick, so I guess its okay.  I really don't understand the point of the helper method to do the cast.  Why not have the single site caller do the explicit cast rather than add the `GetCPPLanguageRuntime` method?

compnerd:
@swift-ci please test

bulbazord:
@compnerd: This is a request from @jimingham. If we add more instances of `GetCPPLanguageRuntime` it might be a bigger deal, kind of like what is happening with the ObjC case today.

jimingham:
Right, the equivalent from ObjC is used many times, and so the Get function is more clearly motivated.  But if we're going to use the Get pattern for ObjC, which I presume @xiaobai will get to at some point, then we should also use it for CPP - which is what he got to first...

dcci (approved):
LGTM.

PR #1662, opened 2019-06-05
Cherry picking changes for SystemV ABI and SwiftASTContext::CanPassInRegisters.

plotfi:
These are for build czar duties. @xiaobai @compnerd.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -5135,6 +5135,11 @@ bool SwiftASTContext::IsVoidType(void *type) {
   return type == GetASTContext()->TheEmptyTupleType.getPointer();
 }
 
+bool SwiftASTContext::CanPassInRegisters(const CompilerType &type) {
+  abort();
+  return false;
+}
+
 bool SwiftASTContext::IsGenericType(const CompilerType &compiler_type) {
   if (!compiler_type.IsValid())
     return false;
--- a/include/lldb/Symbol/ClangASTContext.h
+++ b/include/lldb/Symbol/ClangASTContext.h
@@ -638,6 +638,8 @@ public:
 
   bool IsVoidType(lldb::opaque_compiler_type_t type) override;
 
+  bool CanPassInRegisters(const CompilerType &type) override;
+
   bool SupportsLanguage(lldb::LanguageType language) override;
 
   static bool GetCXXClassName(const CompilerType &type,
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -453,6 +453,8 @@ public:
 
   bool IsVoidType(void *type) override;
 
+  bool CanPassInRegisters(const CompilerType &type) override;
+
   static bool IsGenericType(const CompilerType &compiler_type);
 
   bool IsErrorType(const CompilerType &compiler_type);
--- a/include/lldb/Symbol/TypeSystem.h
+++ b/include/lldb/Symbol/TypeSystem.h
@@ -193,6 +193,8 @@ public:
 
   virtual bool IsVoidType(lldb::opaque_compiler_type_t type) = 0;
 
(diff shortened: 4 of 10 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @dcci @adrian-prantl

source/Symbol/SwiftASTContext.cpp
```diff
@@ -5135,6 +5135,11 @@ bool SwiftASTContext::IsVoidType(void *type) {
   return type == GetASTContext()->TheEmptyTupleType.getPointer();
 }
 
+bool SwiftASTContext::CanPassInRegisters(const CompilerType &type) {
+  abort();
```
dcci, line 5139: Is abort what we really want here? Maybe use `unreachable` instead?

dcci (changes requested):

source/Symbol/SwiftASTContext.cpp
compnerd: I think that this is reachable.  This would be invoked if you hit a frame that returns a value and you do `fin`.  That should pop the frame and attempt to extract the value of the returned value, which may have been SRoA'ed - and if the frame is a Swift frame, this function would be responsible for evaluating that.  This would also mean that we avoid duplicating the SRoA logic in LLDB, remove all the code that deals with whether types can be flattened, and ensure that we see the same behaviour as the compiler always for parameter passing.

I was hoping that this will fail the tests.

dcci: I can see this argument happening for code that's compiled with `-O`, but is this happening also at `-Onone`? [genuine question, I haven't looked at the pipeline nor codegen]

plotfi: @dcci The failing Swift Test on OS X Platform appears to be:

```
Assertion failed: (m_exe_ctx.GetTargetPtr() == NULL), function CheckRequirements, file /Users/plotfi/Projects/Swift/lldb/source/Interpreter/CommandObject.cpp, line 147.
0  lldb                     0x000000010eaeb99c llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 60
1  lldb                     0x000000010eaebf59 PrintStackTraceSignalHandler(void*) + 25
2  lldb                     0x000000010eae8f96 llvm::sys::RunSignalHandlers() + 118
3  lldb                     0x000000010eaf2992 SignalHandler(int) + 210
4  libsystem_platform.dylib 0x00007fff7a111b5d _sigtramp + 29
5  LLDB                     0x000000011c77bf66 cmark_strbuf__initbuf + 111365
6  libsystem_c.dylib        0x00007fff79fd16a6 abort + 127
7  libsystem_c.dylib        0x00007fff79f9a20d basename_r + 0
8  LLDB                     0x0000000110a47d6a lldb_private::CommandObject::CheckRequirements(lldb_private::CommandReturnObject&) + 106
9  LLDB                     0x0000000110a4c00e lldb_private::CommandObjectParsed::Execute(char const*, lldb_private::CommandReturnObject&) + 894
```

Code is at  source/Interpreter/CommandObject.cpp:147 

```
bool CommandObject::CheckRequirements(CommandReturnObject &result) {
#ifdef LLDB_CONFIGURATION_DEBUG
  // Nothing should be stored in m_exe_ctx between running commands as
  // m_exe_ctx has shared pointers to the target, process, thread and frame and
  // we don't want any CommandObject instances to keep any of these objects
  // around longer than for a single command. Every command should call
  // CommandObject::Cleanup() after it has completed
  assert(m_exe_ctx.GetTargetPtr() == NULL);
  assert(m_exe_ctx.GetProcessPtr() == NULL);
  assert(m_exe_ctx.GetThreadPtr() == NULL);
  assert(m_exe_ctx.GetFramePtr() == NULL);
#endif
```

I might need some help figuring out where this CommandObject::Cleanup() call needs to be invoked.

dcci: @JDevlieghere knows and might be of help here.

PR #1666, opened 2019-06-10
[LanguageRuntime] Introduce LLVM-style casts

bulbazord:
This is a cherry-pick of rL362884 from llvm.org and an accompanying patch to support swift.

cc @compnerd @dcci @JDevlieghere


compnerd:
@swift-ci please test

PR #1667, opened 2019-06-10
[5.1][Swift] Add a test for getBitMaskForNoPayloadElements().

dcci:
The fix is in the swift compiler [pull request #25298]

<rdar://problem/51211938>


dcci:
Please test with the following PR:
apple/swift#25333

@swift-ci please test

PR #1669, opened 2019-06-11
Fix a use-after free in triple handling.

adrian-prantl:
This was caught by the Ubuntu 18.04 bots.

<rdar://problem/51454103>


adrian-prantl:
@swift-ci test and merge

PR #1670, opened 2019-06-11
 Fix a crash in option parsing.

adrian-prantl:
The call to getopt_long didn't handle the case where the *last* option
had an argument missing.

<rdar://problem/51231882>

Differential Revision: https://reviews.llvm.org/D63110

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@363101 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 2abc98b56446972e7460fdc69c54270c3a3d8edf)

### DIFF
--- a/lit/Driver/Inputs/process_attach_pid.in
+++ b/lit/Driver/Inputs/process_attach_pid.in
@@ -1,0 +1,1 @@
+process attach --pid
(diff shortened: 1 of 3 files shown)


adrian-prantl:
@swift-ci test and merge

PR #1671, opened 2019-06-11
When reading ObjC class table, use new SPI if it is avail

jasonmolenda:
When reading ObjC class table, use new SPI if it is avail

In the latest OS betas, the objc runtime has a special interface
for the debugger, class_getNameRaw(), instead of the existing
class_getName(), which will return class names in their raw, unmangled
(in the case of swift) form.  When lldb can access the unmangled
names of classes, it won't need to fetch them out of the inferior
process after we run our "get the objc class table" expression.

If the new interface is absent (debugging a process on an older
target), lldb will fall back to class_getName and reading any class
names that it got back in demangled form, at a bit of a performance
cost on the first expression.

on llvm.org in r363103.

<rdar://problem/50688054>


jasonmolenda:
@swift-ci test and merge

compnerd:
This broke the windows build! https://dev.azure.com/compnerd/windows-swift/_build/results?buildId=3116
CC: @adrian-prantl @dcci @JDevlieghere

compnerd:
This was also reverted upstream in SVN r363115

PR #1685, opened 2019-06-13
Cherry picking changes for SystemV ABI and SwiftASTContext::CanPassInRegisters.

plotfi:
These are for build czar duties. @xiaobai @compnerd.

These are the same patches as PR #1662


compnerd:
@swift-ci please test

PR #1688, opened 2019-06-14
SymbolFile: adjust for SVN r363381

compnerd:
SymbolFile is now stored as a reference rather than a pointer.  Adjust
accordingly.


compnerd:

fredriss:
This seems trivial enough that I'm going to merge it without asking for tests (I suppose they won't pass on this branch).

PR #1693, opened 2019-06-17
[SwiftExpressionParser] Don't get an unbound version of a bound generic.

dcci:
There's no need, and it breaks type resolution for `self` in the
expression parser. Not only removing the hack fixes a crash but
now it also does the right thing WRT evaluating members, computed
properties et similia.

<rdar://problem/30933988>
<rdar://problem/51405362>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/member_lookup_crash/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/member_lookup_crash/main.swift
@@ -10,7 +10,8 @@ extension Measurement where UnitType == UnitAngle {
   }
 
   func f() {
-    return //%self.expect('p self.radians', substrs=["property 'radians' requires the types 'UnitType' and 'UnitAngle' be equivalent"], error=True)
+    return //%self.expect('p self.radians', substrs=["(CGFloat) $R0 = 1.745"])
+           //%self.expect('p self', substrs=["Measurement<UnitAngle>"])
   }
 }
 
--- a/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
@@ -572,17 +572,6 @@ AddRequiredAliases(Block *block, lldb::StackFrameSP &stack_frame_sp,
     imported_self_type = CompilerType(imported_self_type.GetTypeSystem(),
                                       object_type.getPointer());
 
-  // If the type of 'self' is a bound generic type, get the unbound version.
-  bool is_generic = imported_self_type_flags.AllSet(lldb::eTypeIsSwift |
-                                                    lldb::eTypeIsGeneric);
-  bool is_bound =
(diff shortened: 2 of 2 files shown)


dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

packages/Python/lldbsuite/test/lang/swift/member_lookup_crash/main.swift
```diff
@@ -10,7 +10,8 @@ extension Measurement where UnitType == UnitAngle {
   }
 
   func f() {
-    return //%self.expect('p self.radians', substrs=["property 'radians' requires the types 'UnitType' and 'UnitAngle' be equivalent"], error=True)
+    return //%self.expect('p self.radians', substrs=["(CGFloat) $R0 = 1.745"])
+           //%self.expect('p self', substrs=["Measurement<UnitAngle>"])
```
adrian-prantl, line 14: Do you think it would be possible to extract the interesting bits from the type and remove the dependency on Foundation? It would make it much clearer what is being tested for.

packages/Python/lldbsuite/test/lang/swift/member_lookup_crash/main.swift
```diff
@@ -10,7 +10,8 @@ extension Measurement where UnitType == UnitAngle {
   }
 
   func f() {
```
adrian-prantl, line 12: Can you rename this test to something more meaningful than "crash"?

packages/Python/lldbsuite/test/lang/swift/member_lookup_crash/main.swift
dcci: Maybe, but this is testcase reported by somebody internally so I would love to keep it as is.
[we can add another one]

packages/Python/lldbsuite/test/lang/swift/member_lookup_crash/main.swift
dcci: Sure, the test contained the word "crash" to begin with but I agree it can be renamed to something more useful :)

packages/Python/lldbsuite/test/lang/swift/member_lookup_crash/main.swift
adrian-prantl: If you want to keep it, you could duplicate the test into
1. an end-to-end test inside the foundation subdirectory
2. a coverage test that faithfully reproduces the bug fixed by this commit without any dependencies

dcci:
Self-contained test case for you:

```
protocol Tinky {}

struct Winky : Tinky {
  var x : Int
}

extension Patatino where T == Winky {
  var baciotto : Int {
    return 0
  }

  func f() {
    return
  }
}

struct Patatino<T> where T : Tinky {
  let x : T
}

let pat = Patatino<Winky>(x: Winky(x: 23))
pat.f()
```

PR #1694, opened 2019-06-18
[Swift] Add a testcase that doesn't depend on Foundation.

dcci:
While I'm here, rename the other test to something more useful.


dcci:
@swift-ci test and merge

dcci:
@swift-ci test linux platform

PR #1696, opened 2019-06-18
 Only perform the zero-sized type shortcut in ValueObject::Update for …

adrian-prantl:
…Swift.

In C++ even empty structs have a size of at least 1.

rdar://problem/51815632


adrian-prantl:
@swift-ci test and merge

PR #1698, opened 2019-06-18
[5.1][SwiftExpressionParser] Don't get an unbound version of a bound generic.

dcci:


dcci:
@swift-ci test and merge

PR #1702, opened 2019-06-19
Merge conflict in HashedNameToDIE.h

weliveindetail:
I think my resolution is ok, but anyway..


source/Plugins/SymbolFile/DWARF/HashedNameToDIE.h
```diff
@@ -53,16 +53,24 @@ class DWARFMappedHash {
...
-    DIEInfo(dw_offset_t c, dw_offset_t o, dw_tag_t t, uint32_t f, uint32_t h);
+    /// Any flags for this DIEInfo
+    uint32_t type_flags = 0;
 
-    explicit operator DIERef() const { return die_ref; }
+    /// A 32 bit hash of the fully qualified name
+    uint32_t qualified_name_hash = 0;
+
+    // 64 bit string table offset
+    uint64_t strp;
```
weliveindetail: `strp` is not upstream. It's also unused in swift-lldb as far as I understand. It was added with the initial "Add Swift support to LLDB" https://github.com/apple/swift-lldb/commit/84ef5bf74617

Upstream code was touched in https://reviews.llvm.org/D63399. It mentions a series of patches to come.

source/Plugins/SymbolFile/DWARF/HashedNameToDIE.cpp
```diff
@@ -117,13 +117,9 @@ const char *DWARFMappedHash::GetAtomTypeName(uint16_t atom) {
...
-DWARFMappedHash::DIEInfo::DIEInfo()
-    : tag(0), type_flags(0), qualified_name_hash(0) {}
-
-DWARFMappedHash::DIEInfo::DIEInfo(dw_offset_t c, dw_offset_t o, dw_tag_t t,
-                                  uint32_t f, uint32_t h)
-    : die_ref(DIERef::Section::DebugInfo, c, o), tag(t), type_flags(f),
-      qualified_name_hash(h) {}
+DWARFMappedHash::DIEInfo::DIEInfo(dw_offset_t o, dw_tag_t t, uint32_t f,
+                                  uint32_t h)
+    : die_offset(o), tag(t), type_flags(f), qualified_name_hash(h) {}
```
weliveindetail, line 124: Interestingly, there was no merge conflict here although the original https://github.com/apple/swift-lldb/commit/84ef5bf74617 also added initialization for `strp`.

JDevlieghere: Maybe it was already mis-merged in the past?

source/Plugins/SymbolFile/DWARF/HashedNameToDIE.h
JDevlieghere: So we should just remove it then?

source/Plugins/SymbolFile/DWARF/HashedNameToDIE.cpp
weliveindetail: Yes, this one: https://github.com/apple/swift-lldb/commit/9b5219050f16

It looks like mainline did initialize `strp` before:
```
$ git blame 9b5219050f^ HashedNameToDIE.cpp | grep -w strp
4534090caff (Todd Fiala       2016-09-07 16:29:46 -0700 118)     return "strp";
4534090caff (Todd Fiala       2016-09-07 16:29:46 -0700 125)       type_flags(0), qualified_name_hash(0), strp(UINT64_MAX) {}
4534090caff (Todd Fiala       2016-09-07 16:29:46 -0700 130)       strp(s) {}
```

Upstream didn't and this is what was taken:
```
$ git blame 9b5219050f^2 HashedNameToDIE.cpp | grep -w strp
$ git blame 9b5219050f HashedNameToDIE.cpp | grep -w strp
4534090caff (Todd Fiala       2016-09-07 16:29:46 -0700 117)     return "strp";
```

source/Plugins/SymbolFile/DWARF/HashedNameToDIE.h
weliveindetail: Well, it looks like we lost functionality here, but as I don't know what it does, I cannot say whether or not we need it. I only see a unreasonable merge conflict.

source/Plugins/SymbolFile/DWARF/HashedNameToDIE.cpp
dcci: But is this actually used [other than being initialized? Otherwise we can just remove it.

weliveindetail: Well, it was used in the original commit https://github.com/apple/swift-lldb/blob/84ef5bf746/source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp#L2145, but it isn't used anymore (and neither was it used at the point of the merge in question).

So, I guess it's fine to remove the field and get this through. Thanks

PR #1705, opened 2019-06-24
Merge remote-tracking branch 'llvm/master' into upstream-with-swift

weliveindetail:
It looks like after https://github.com/apple/swift-lldb/commit/cde4416d690b lands in swift-lldb, we need to implement `GetBreakpointExceptionPrecondition()` also for `SwiftLanguageRuntime`, just as it was done upstream for `ObjCLanguageRuntime` in https://reviews.llvm.org/rG7f9c9f226423.


bulbazord:
Yeah, we're going to definitely need to implement it for SwiftLanguageRuntime. Let's get this merged in and we can sort it out afterwards.

PR #1706, opened 2019-06-24
Set DYLD_LIBRARY_PATH to test with the just-built stdlib

JDevlieghere:
With Swift in the OS, we need to ensure that the test suite runs with
the standard library we just built, instead of the one form the
operating system. We do this by telling dyld to pick the one from our
build directory by setting the `DYLD_LIBRARY_PATH`. Swift takes the same
approach in their test suite.


JDevlieghere:
@swift-ci please test

adrian-prantl:
Does this not work at all at the moment, or only for some tests?
I wonder if it would be better to put this into the lit/lit-lldb-init file we are passing to every lldb instance.

JDevlieghere:
> I wonder if it would be better to put this into the lit/lit-lldb-init file we are passing to every lldb instance.

That's exactly what this patch does :-) Maybe I'm misunderstanding the question?

adrian-prantl (approved):
Sorry, I see it now!

dcci (approved):
This is correct. I'll let you click the merge button.

PR #1708, opened 2019-06-24
[Reproducer] Extend reproducer test case.

JDevlieghere:
Check that we can remove the binary and move the reproducer.


JDevlieghere:

PR #1709, opened 2019-06-24
[NFC] Replace a plugin header with a non-plugin header

bulbazord:
git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@363338 91177308-0d34-0410-b5e6-96231b3b80d8

cc @JDevlieghere @dcci


bulbazord:
@swift-ci please test

PR #1711, opened 2019-06-24
[Target] Decouple ObjCLanguageRuntime from LanguageRuntime

bulbazord:
git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@364098 91177308-0d34-0410-b5e6-96231b3b80d8

This PR also has a patch to account for needed changes to swift. The cherry-picked patch has also landed in upstream-with-swift, but the patch to adjust for swift hasn't. It should propagate from stable into upstream-with-swift.

cc @weliveindetail @dcci @JDevlieghere


bulbazord:
@swift-ci please test

PR #1719, opened 2019-06-25
[Target][NFC] Simplify ThreadPlanStepOut constructor

bulbazord:
The constructor tries to figure out if it's stepping out of a frame with swift
code because it wants to ensure any potential error is preserved. This commit
simplifies the process of determining the language of the frame by asking the
frame to try to determine it.

This should be equivalent, because StackFrame::GuessLanguage does what the
constructor was doing before and more.

cc @dcci @JDevlieghere


bulbazord:
@swift-ci please test

PR #1720, opened 2019-06-25
[Target] Clean up ThreadPlanStepInRange::DefaultShouldStopHere{Callback,Impl}

bulbazord:
The following 3 commits do the following:

1) Clear up some confusing logic in DefaultShouldStopHereImpl
2) Remove commented out (dead) code.
3) Hoist a method from SwiftLanguageRuntime into LanguageRuntime to generalize some behavior in ThreadPlanStepInRange.

cc @JDevlieghere @dcci @compnerd @adrian-prantl

### DIFF
--- a/source/Target/ThreadPlanStepInRange.cpp
+++ b/source/Target/ThreadPlanStepInRange.cpp
@@ -12,9 +12,9 @@
 #include "lldb/Core/Module.h"
 #include "lldb/Symbol/Function.h"
 #include "lldb/Symbol/Symbol.h"
+#include "lldb/Target/LanguageRuntime.h"
 #include "lldb/Target/Process.h"
 #include "lldb/Target/RegisterContext.h"
-#include "lldb/Target/SwiftLanguageRuntime.h"
 #include "lldb/Target/SectionLoadList.h"
 #include "lldb/Target/Target.h"
 #include "lldb/Target/Thread.h"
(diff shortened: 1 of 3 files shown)


bulbazord:
@swift-ci please test

source/Target/ThreadPlanStepInRange.cpp
```diff
@@ -532,41 +532,31 @@ bool ThreadPlanStepInRange::DefaultShouldStopHereImpl(Flags &flags,
...
-                bkpt_sp->SetThreadID(GetThread().GetID());
-                m_step_in_deep_bps.push_back(bkpt_sp->GetID());
-              }
-            }
+    auto frame_language = frame->GuessLanguage();
+    if (auto *runtime =
+            GetThread().GetProcess()->GetLanguageRuntime(frame_language)) {
+      std::vector<Address> interesting_addresses;
+      runtime->FindFunctionPointersInCall(*frame, interesting_addresses);
+      if (interesting_addresses.size() > 0) {
```
compnerd: Might be nicer to spell this `!interesting_addresses.empty()`

source/Target/ThreadPlanStepInRange.cpp
```diff
@@ -532,41 +532,31 @@ bool ThreadPlanStepInRange::DefaultShouldStopHereImpl(Flags &flags,
...
+        // Run through the addresses we found, make sure they have debug info,
+        // and if so set breakpoints on all these addresses.
+        for (size_t i = 0; i < interesting_addresses.size(); i++) {
```
compnerd: Is there a reason to not prefer:

```
for (const auto & address : interesting_address) {
```

source/Target/ThreadPlanStepInRange.cpp
```diff
@@ -532,41 +532,31 @@ bool ThreadPlanStepInRange::DefaultShouldStopHereImpl(Flags &flags,
...
+          LineEntry line_entry;
+          if (interesting_addresses[i].CalculateSymbolContextLineEntry(
+                  line_entry)) {
+            // It has debug information, use it:
+            const bool internal = true;
+            const bool hardware = false;
+            BreakpointSP bkpt_sp = GetTarget().CreateBreakpoint(
```
compnerd: Could probably hoist the `GetTarget()` and `GetThread().GetID()` calls into variables outside of the loop.

bulbazord: They're called exactly once so it's not a big deal either way imo

bulbazord:
@swift-ci please test

PR #1726, opened 2019-06-25
[LanguageRuntime] Hoist SwiftLanguageRuntime::IsSymbolARuntimeThunk

bulbazord:
There's nothing swift-specific about thunks. There was a static function
in LanguageRuntime to determine if a symbol is a runtime thunk, but this just
creates a circular dependency between LanguageRuntime and SwiftLanguageRuntime.
I think if we hoist the one in SwiftLanguageRuntime into LanguageRuntime and
make it a virtual method, we should have a better, more general implementation.

cc @dcci @JDevlieghere @compnerd @adrian-prantl

### DIFF
--- a/include/lldb/Target/LanguageRuntime.h
+++ b/include/lldb/Target/LanguageRuntime.h
@@ -189,7 +189,10 @@ public:
     return false;
   }
 
-  static bool IsSymbolAnyRuntimeThunk(Symbol &symbol);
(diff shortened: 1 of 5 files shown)


bulbazord:
@swift-ci please test

compnerd (approved):

bulbazord:
@swift-ci please test

PR #1729, opened 2019-06-25
 Surface Swift deserialization problems to the user.

adrian-prantl:
Error messages such as .swiftmodules being produced by an older
compiler are common, but were previously not visible to and LLDB user
unless they also turned on the types log. This patch changes that by
surfacing the errors on the LLDB error stream, but only once per
module.

<rdar://problem/51686950>

### DIFF
--- a/include/lldb/Target/Process.h
+++ b/include/lldb/Target/Process.h
@@ -536,7 +536,7 @@ public:
   //------------------------------------------------------------------
   /// Process warning types.
   //------------------------------------------------------------------
-  enum Warnings { eWarningsOptimization = 1, eWarningsCantLoadSwift };
+  enum Warnings { eWarningsOptimization = 1, eWarningsSwiftImport };
 
   typedef Range<lldb::addr_t, lldb::addr_t> LoadRange;
   // We use a read/write lock to allow on or more clients to access the process
@@ -1590,19 +1590,15 @@ public:
   void PrintWarningOptimization(const SymbolContext &sc);
 
   //------------------------------------------------------------------
-  /// Print a user-visible warning about a module having Swift settings
-  /// incompatible with the current system
-  ///
-  /// Prints a async warning message to the user one time per Process for a
-  /// Module
-  /// whose Swift AST sections couldn't be loaded because they aren't buildable
-  /// on
-  /// the current machine.
+  /// Prints a async warning message to the user one time per Process
+  /// for a Module whose Swift AST sections couldn't be loaded because
+  /// they aren't buildable on the current machine.
   ///
   /// @param [in] module
   ///     The affected Module.
   //------------------------------------------------------------------
-  void PrintWarningCantLoadSwift(const Module &module);
+  void PrintWarningCantLoadSwiftModule(const Module &module,
+                                       std::string details);
 
   virtual bool GetProcessInfo(ProcessInstanceInfo &info);
 
--- a/packages/Python/lldbsuite/test/lang/swift/deserialization_failure/TestSwiftDeserializationFailure.py
+++ b/packages/Python/lldbsuite/test/lang/swift/deserialization_failure/TestSwiftDeserializationFailure.py
@@ -25,7 +25,8 @@ class TestSwiftDeserializationFailure(TestBase):
         dynamic_bkpt = target.BreakpointCreateByName('dynamicTypes')
         generic_bkpt = target.BreakpointCreateByName('genericTypes')
         lldbutil.continue_to_breakpoint(process, static_bkpt)
-        self.expect("fr var hello", substrs=["(String)", "world"])
+        self.expect("fr var i", substrs=["23"])
+        self.expect("fr var s", substrs=["(String)", "world"])
 
         # We should not be able to resolve the types defined in the module.
         lldbutil.continue_to_breakpoint(process, dynamic_bkpt)
@@ -50,5 +51,6 @@ class TestSwiftDeserializationFailure(TestBase):
         self.prepare()
         with open(self.getBuildArtifact("a.swiftmodule"), 'w') as mod:
             mod.write('I am damaged.\n')
+
         target, process, _, _ = lldbutil.run_to_name_breakpoint(self, 'main')
         self.run_tests(target, process)
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1335,22 +1335,24 @@ static bool DeserializeCompilerFlags(swift::CompilerInvocation &invocation,
   return false;
 }
 
-static void printASTValidationInfo(
+static void printASTValidationError(
     llvm::raw_ostream &errs,
     const swift::serialization::ValidationInfo &ast_info,
     const swift::serialization::ExtendedValidationInfo &ext_ast_info,
-    const Module &module, StringRef module_buf, bool invalid_name,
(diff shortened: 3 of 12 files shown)


adrian-prantl:
@swift-ci test

dcci:
This requires more mental bandwidth than I have available today. I'll take a look at it tomorrow.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1335,22 +1335,23 @@ static bool DeserializeCompilerFlags(swift::CompilerInvocation &invocation,
...
-  }
+  errs << "AST validation error";
+  if (!invalid_name)
+    errs << " in \"" << ast_info.name << '"';
+  errs << ": ";
+  if (invalid_size)
+    errs << "The serialized module is corrupted.";
+  else if (invalid_name)
+    errs << "The serialized module has an invalid name.";
+  else
```
jimingham, line 1354: Even if the layers above you determined that the error is invalid_name or invalid_size, do you know the getImportFailureString - if it is not empty - doesn't have additional relevant info?  Maybe print it unconditionally if it is non-empty.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1369,15 +1370,21 @@ static void printASTValidationInfo(
...
+  for (const std::string &message : m_module_import_warnings)
+    process.PrintWarningCantLoadSwift(module, message.c_str());
+}
+
 /// Retrieve the serialized AST data blobs and initialize the compiler
 /// invocation with the concatenated search paths from the blobs.
 /// \returns true if an error was encountered.
 static bool DeserializeAllCompilerFlags(SwiftASTContext &swift_ast,
-                                        Module &module,
+                                        Process *process, Module &module,
```
jimingham: You added "process" to the API here, but you don't seem to use it anywhere?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1705,12 +1713,12 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
     bool got_serialized_options;
     llvm::SmallString<0> error;
     llvm::raw_svector_ostream errs(error);
-
-    if (DeserializeAllCompilerFlags(*swift_ast_sp, module, m_description, errs,
+    Process *process = target ? target->GetProcessSP().get() : nullptr;
+    if (DeserializeAllCompilerFlags(*swift_ast_sp, process, module,
```
jimingham: I doesn't look to me like DeserializeAllCompilerFlags actually uses the process, so maybe you don't need to do this?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3990,7 +3993,7 @@ void SwiftASTContext::ValidateSectionModules(
     SourceModule module_info;
     module_info.path.push_back(ConstString(module_name));
     if (!GetModule(module_info, error))
-      module.ReportWarning("unable to load swift module \"%s\" (%s)",
+      module.ReportWarning("ZZZZ unable to load swift module \"%s\" (%s)",
```
jimingham: You probably intended to remove this ZZZZ, right?

jimingham (commented):
This makes sense to me in general.  You added a Process * argument to DeserializeAllCompilerFlags, and have to add changes to pass the process into this call, but I don't see anywhere in DeserializeAllCompilerFlags where you actually use it.

include/lldb/Target/Process.h
```diff
@@ -1590,19 +1590,14 @@ class Process : public std::enable_shared_from_this<Process>,
...
-  /// on
+  /// Prints a async warning message to the user one time per Process for a Module
+  /// whose Swift AST sections couldn't be loaded because they aren't buildable on
   /// the current machine.
   ///
   /// @param [in] module
   ///     The affected Module.
   //------------------------------------------------------------------
-  void PrintWarningCantLoadSwift(const Module &module);
+  void PrintWarningCantLoadSwift(const Module &module, const char *details);
```
dcci: Should this just be a `std::string` or a `StringRef` ?

dcci: Can we rename this function to something more generic? Something like `PrintWarningCantLoadModule`

packages/Python/lldbsuite/test/lang/swift/deserialization_failure/Inputs/main.swift
```diff
@@ -9,3 +9,4 @@ let hello = "world"
 staticTypes(23, hello)
 dynamicTypes(D())
 genericTypes(hello)
+
```
dcci: Unneeded?

packages/Python/lldbsuite/test/lang/swift/deserialization_failure/TestSwiftDeserializationFailure.py
```diff
@@ -50,5 +51,6 @@ def test_damaged_module(self):
         self.prepare()
         with open(self.getBuildArtifact("a.swiftmodule"), 'w') as mod:
             mod.write('I am damaged.\n')
+
```
dcci, line 54: You can probably remove this one?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1369,15 +1370,21 @@ static void printASTValidationInfo(
     LLDB_LOG(log, "  -- {0}", ExtraOpt);
 }
 
+void SwiftASTContext::DiagnoseWarnings(Process &process, Module &module) const {
```
dcci, line 1374: Process seems unused here?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1369,15 +1370,21 @@ static void printASTValidationInfo(
...
+  for (const std::string &message : m_module_import_warnings)
+    process.PrintWarningCantLoadSwift(module, message.c_str());
```
dcci: If you make that function taking a string you don't need the `.c_str()` conversion.

dcci:
Thank you very much for taking a look at this! Some comments inline.

include/lldb/Target/Process.h
adrian-prantl: This is feeding into `Process::PrintWarning(uint64_t warning_type, const void *repeat_key, const char *fmt, ...)` so a StringRef is not an option because the NUL isn't guaranteed. A std::string would work though.

adrian-prantl: done.

source/Symbol/SwiftASTContext.cpp
adrian-prantl: That's intentional, if the size or the name is invalid, we don't even parse, and the error will be, "The module file is malformed in some way." and since we know the "some way" we explicitly be more specific. I'll add a comment.

source/Symbol/SwiftASTContext.cpp
adrian-prantl: thanks, that's a holdover from an earlier attempt.

adrian-prantl: Sorry.. it's unused in DeserializeCompilerFlags. It's very much in use here :-)

adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

PR #1731, opened 2019-06-26
Update test for newer host os.

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #1732, opened 2019-06-26
Make test pass on swift-5.1-branch.

adrian-prantl:
The underlying patch is to avoid a crash, so the exact error message
is not substantial.


adrian-prantl:
@swift-ci test and merge

PR #1736, opened 2019-06-26
[x86-64] Use `//` for integer division in the target definition.

dcci:
This forces integer division and works with python 2 and python 3.

<rdar://problem/52073911>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@364465 91177308-0d34-0410-b5e6-96231b3b80d8


dcci:
@swift-ci test and merge

PR #1742, opened 2019-06-27
Reproducer: Add test for bridging header

JDevlieghere:


JDevlieghere:

PR #1745, opened 2019-06-27
 Avoid a crash when the ClangImporter couldn't be initialized.

adrian-prantl:
This needs more cleanup and a better architecture, most notably
the property getter should never return an empty string, but that
needs to happen upstream and not on swift-5.1-branch.

<rdar://problem/52222566>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -819,16 +819,20 @@ static bool IsDeviceSupport(const char *path) {
 }
 } // namespace
 
+static std::string GetClangModulesCacheProperty() {
+  llvm::SmallString<128> path;
+  auto props = ModuleList::GetGlobalModuleListProperties();
+  props.GetClangModulesCachePath().GetPath(path);
+  return path.str();
+}
+
 SwiftASTContext::SwiftASTContext(std::string description, llvm::Triple triple,
                                  Target *target)
     : TypeSystem(TypeSystem::eKindSwift),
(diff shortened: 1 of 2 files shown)


adrian-prantl:
@swift-ci test and merge

adrian-prantl:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3131,21 +3135,39 @@ swift::ASTContext *SwiftASTContext::GetASTContext() {
...
+    // Even though it is initialized to the default Clang location at startup a
+    // user could have overwritten it with an empty path.
+    if (moduleCachePath.empty()) {
+      llvm::SmallString<0> path;
+      std::error_code ec =
+          llvm::sys::fs::createUniqueDirectory("ModuleCache", path);
+      if (!ec)
+        moduleCachePath = path.str();
+      else
+        moduleCachePath = "/tmp/lldb-ModuleCache";
```
dcci, line 3167: is there a way we can avoid hardcoding the module cache path when we end up fixing this upstream? I think it's OK for the branch, but it would be great if all this knowledge didn't leak in lldb.

adrian-prantl: I don't know what you mean by hardcoding. We are

1. Asking for the user-specified path
2. Asking Clang to compute the default location
3. Aking for a unique temporary directory
4. use /tmp/lldb-ModuleCache

in that order.

JDevlieghere:
@swift-ci test and merge

PR #1748, opened 2019-06-27
[Swift] Add a test check to make sure we resolve metatypes correctly.

dcci:
<rdar://problem/45600654>


dcci:
@swift-ci test and merge

PR #1751, opened 2019-06-28
Use std::string::c_str() instead of data() wherever possible.

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

include/lldb/Target/SwiftLanguageRuntime.h
```diff
@@ -326,9 +326,9 @@ class SwiftLanguageRuntime : public LanguageRuntime {
   /// Determines wether \c variable is the "self" object.
   static bool IsSelf(Variable &variable);
 
-  void AddToLibraryNegativeCache(const char *library_name);
+  void AddToLibraryNegativeCache(std::string library_name);
```
JDevlieghere, line 329: Why not a StringRef?

adrian-prantl: Because the negative cache is a std::unordered_set<std::string>. I suppose an llvm::StringMap would be better but I have to limit the scope of this patch somewhere.

dcci: I think we should just replace the negative cache and the positive cache with a single one taking an `Optional<>` [also for MangledName -> Type]

dcci (approved):

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -3386,12 +3386,12 @@ Status SwiftLanguageRuntime::SwiftExceptionPrecondition::ConfigurePrecondition(
...
 }
 
-void SwiftLanguageRuntime::AddToLibraryNegativeCache(const char *library_name) {
+void SwiftLanguageRuntime::AddToLibraryNegativeCache(std::string library_name) {
   std::lock_guard<std::mutex> locker(m_negative_cache_mutex);
   m_library_negative_cache.insert(library_name);
 }
 
-bool SwiftLanguageRuntime::IsInLibraryNegativeCache(const char *library_name) {
+bool SwiftLanguageRuntime::IsInLibraryNegativeCache(std::string library_name) {
```
JDevlieghere, line 3394: Const ref?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -3386,12 +3386,12 @@ Status SwiftLanguageRuntime::SwiftExceptionPrecondition::ConfigurePrecondition(
   return error;
 }
 
-void SwiftLanguageRuntime::AddToLibraryNegativeCache(const char *library_name) {
+void SwiftLanguageRuntime::AddToLibraryNegativeCache(std::string library_name) {
```
JDevlieghere, line 3389: Const ref?

JDevlieghere (changes requested):

include/lldb/Target/SwiftLanguageRuntime.h
JDevlieghere: Fair enough. In that case it could still take a const ref, right?

PR #1755, opened 2019-06-28
[SwiftArray] Pass an execution scope when getting the size of the array.

dcci:
For non-fixed layout types, we have to ask the runtime what's the
size of the type (and go through mirrors). This shows up with resilient
types, but probably it was a more general problem.
Thanks to Adrian for help to initially diagnose this!

<rdar://problem/51146107>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/array_tuple_resilient/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/array_tuple_resilient/Makefile
@@ -1,0 +1,3 @@
+LEVEL = ../../../make
+SWIFT_SOURCES := main.swift
+include $(LEVEL)/Makefile.rules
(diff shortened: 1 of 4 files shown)


dcci:
@swift-ci test

adrian-prantl (commented):
Line SwiftArray.cpp:197 seems to have the exact same bug.

dcci:
@adrian-prantl thanks for the catch. Updated.

dcci:
I'll audit all the `getBitSize()` next, but this should be good for now.

dcci:
@swift-ci test

adrian-prantl:
Great! GetBitStride is also a likely candidate, if we use it at all.

adrian-prantl (approved):

dcci:
@swift-ci test and merge

PR #1761, opened 2019-07-01
[SwiftASTContext] Fix printing for Obj-C enum.

dcci:
The value is not an increasing number [starting from zero] as
it happens for the pure no-payload swift-enum, they can have
arbitrary values.

<rdar://problem/51678061>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/enum_objc/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/enum_objc/main.swift
@@ -1,0 +1,18 @@
+import Foundation
+
+func test()
+{
+    let a : NSNumber = 0.0
+    let b : NSNumber = 1.0
+    let ascending = a.compare(b)
+    let descending = b.compare(a)
+    let same = a.compare(a)
+    return //%self.expect('frame var -d run-target -- ascending', substrs=['orderedAscending'])
+           //%self.expect('frame var -d run-target -- descending', substrs=['orderedDescending'])
+           //%self.expect('frame var -d run-target -- same', substrs=['orderedSame'])
+           //%self.expect('expr -d run-target -- ascending', substrs=['orderedAscending'])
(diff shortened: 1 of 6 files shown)


dcci:
@swift-ci test

packages/Python/lldbsuite/test/lang/swift/enum_objc/main.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
```
adrian-prantl: Can you just define the enum in a modularized or bridging header file and import that? Otherwise this depends on the implementation of Foundation and this makes it harder to understand what the test is testing and less fragile.

dcci: I've never done this, do we have other examples in tree?

source/Symbol/SwiftASTContext.cpp
```diff
@@ -390,6 +390,9 @@ class SwiftCStyleEnumDescriptor : public SwiftEnumDescriptor {
     m_nopayload_elems_bitmask =
         enum_impl_strategy.getBitMaskForNoPayloadElements();
 
+    if (enum_decl->isObjC())
```
adrian-prantl, line 393: Does this not apply to C enums on Linux?

dcci: So? I think this is the intended behaviour.

dcci:
Addressed Adrian's comments.

dcci:
@swift-ci test

adrian-prantl (approved):
Thanks! This is much better!

source/Symbol/SwiftASTContext.cpp
jckarter: AFAIK, they're still called `@objc` enums on Linux, even though they're technically C enums.

jckarter (approved):
LGTM, thanks Davide.

PR #1763, opened 2019-07-02
Convert SwiftLanguageRuntime::m_bridged_synthetics_map to an

adrian-prantl:
llvm::DenseMap and explicitly replace const char* with ConstString
where this is actually required.

NFC

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -4070,25 +4070,25 @@ ConstString SwiftASTContext::GetMangledTypeName(swift::TypeBase *type_base) {
   swift::Mangle::ASTMangler mangler(true);
   std::string s = mangler.mangleTypeForDebugger(swift_type, nullptr);
   if (s.empty())
-    return ConstString();
+    return {};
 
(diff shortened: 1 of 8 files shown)


adrian-prantl:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -4110,69 +4110,68 @@ swift::Type convertSILFunctionTypesToASTFunctionTypes(swift::Type t) {
...
   if (found_type) {
     LOG_PRINTF(LIBLLDB_LOG_TYPES, "(\"%s\") -- found in the positive cache",
-               mangled_typename);
+               mangled_cstr);
     assert(&found_type->getASTContext() == ast_ctx);
     return {found_type};
   }
 
-  if (m_negative_type_cache.Lookup(mangled_name.GetCString())) {
+  if (m_negative_type_cache.Lookup(mangled_cstr)) {
```
dcci, line 4147: We have a positive and a negative cache. Should we have a single cache instead that takes an `Optional<>` ?

PR #1767, opened 2019-07-02
Generalize ValueObject::IsRuntimeSupportValue

bulbazord:
The first commit in this PR is a cherry-pick from llvm.org (rL364845). The second is a cherry-pick from upstream-with-swift which adjusts for swift's use case.


bulbazord:
@swift-ci please test

PR #1768, opened 2019-07-02
[Core] Remove unused SwiftASTContextLock

bulbazord:
If I understand how SwiftASTContextLock is used, I think that this instance is unused. I could be wrong as I'm not entirely confident in my understanding.

cc @dcci @JDevlieghere @adrian-prantl

### DIFF
--- a/source/Core/ValueObjectDynamicValue.cpp
+++ b/source/Core/ValueObjectDynamicValue.cpp
@@ -147,8 +147,6 @@ bool ValueObjectDynamicValue::UpdateValue() {
     m_data.SetAddressByteSize(target->GetArchitecture().GetAddressByteSize());
   }
  
-  auto swift_scratch_ctx_lock = SwiftASTContextLock(&exe_ctx);
-
   // First make sure our Type and/or Address haven't changed:
   Process *process = exe_ctx.GetProcessPtr();
   if (!process)


bulbazord:
@swift-ci please test

adrian-prantl (changes requested):
`GetDynamicTypeAndAddress` performs dynamic type resolution in the scratch context. We need to lock it to ensure it doesn't get destroyed before we use the type that is returned. If anything the lock needs to be pushed into the caller of this function, but it can't just be removed. Can you explain how you arrived at the conclusion that this is unnecessary?

bulbazord:
> `GetDynamicTypeAndAddress` performs dynamic type resolution in the scratch context. We need to lock it to ensure it doesn't get destroyed before we use the type that is returned. If anything the lock needs to be pushed into the caller of this function, but it can't just be removed. Can you explain how you arrived at the conclusion that this is unnecessary?

Thanks for explaining what's happening. I completely misunderstood the SwiftASTContext lifecycle. I wasn't aware that SwiftASTContexts could be destroyed abruptly, which I obviously would have found out if I kept digging into Swift more. Apologies for the noise.

adrian-prantl:
> I wasn't aware that SwiftASTContexts could be destroyed abruptly

Note that only the scratch context (=SwiftASTContextForExpressions) has this property. That's a big  difference between swift-lldb and how llvm.org lldb handles ClangASTContext. This is not a great system by any means, but being able to tear down the scratch context allows us to recover from fatal errors in the Swift compiler.

PR #1769, opened 2019-07-02
[SwiftLanguage] Add a formatter for StringIndex.

dcci:
First step towards fixing <rdar://problem/31977859>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/string_index/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/string_index/main.swift
@@ -1,0 +1,8 @@
+let name = "patatino"
+let index1 = name.firstIndex(of: "t")
+let index_none = name.firstIndex(of: "Q")
+let long_name = "sed avarissimi hominis cupiditati satisfacere posse"
+let index2 = long_name.firstIndex(of: "h")
+print() //%self.expect('frame var -d run-target -- index1', substrs=['(String.Index?) index1 = 2'])
+        //%self.expect('frame var -d run-target -- index_none', substrs=['(String.Index?) index_none = nil'])
+        //%self.expect('frame var -d run-target -- index2', substrs=['(String.Index?) index2 = 15'])
--- a/source/Plugins/Language/Swift/SwiftFormatters.cpp
+++ b/source/Plugins/Language/Swift/SwiftFormatters.cpp
@@ -68,6 +68,29 @@ bool lldb_private::formatters::swift::SwiftSharedString_SummaryProvider(
       StringPrinter::ReadStringAndDumpToStreamOptions());
 }
 
+// String's Index has the following layout:
+// (mirrored from StringIndex.swift in the standard library.
+//
+// ┌──────────┬───────────────────┬─────────╥────────────────┬──────────┐
+// │ b63:b16  │      b15:b14      │   b13   ║     b12:b8     │  b6:b0   │
+// ├──────────┼───────────────────┼─────────╫────────────────┼──────────┤
+// │ position │ transcoded offset │ aligned ║ grapheme cache │ reserved │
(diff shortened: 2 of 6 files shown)


dcci:
@lorentey / @milseman / @airspeedswift can you please take a look? Thanks!

dcci:
@swift-ci test

source/Plugins/Language/Swift/SwiftFormatters.cpp
```diff
@@ -68,6 +68,29 @@ bool lldb_private::formatters::swift::SwiftSharedString_SummaryProvider(
...
 }
 
+// String's Index has the following layout:
+// (mirrored from StringIndex.swift in the standard library.
+//
+// ┌──────────┬───────────────────┬─────────╥────────────────┬──────────┐
+// │ b63:b16  │      b15:b14      │   b13   ║     b12:b8     │  b6:b0   │
+// ├──────────┼───────────────────┼─────────╫────────────────┼──────────┤
+// │ position │ transcoded offset │ aligned ║ grapheme cache │ reserved │
+// └──────────┴───────────────────┴─────────╨────────────────┴──────────┘
```
milseman, line 78: Is there any way around hardcoding this again? This is changing even now: https://github.com/apple/swift/pull/25948

`position` and `transcoded offset` are baked into the ABI, so if you only use those, that's less sketchy

packages/Python/lldbsuite/test/lang/swift/string_index/main.swift
```diff
@@ -0,0 +1,8 @@
+let name = "patatino"
+let index1 = name.firstIndex(of: "t")
+let index_none = name.firstIndex(of: "Q")
+let long_name = "sed avarissimi hominis cupiditati satisfacere posse"
+let index2 = long_name.firstIndex(of: "h")
+print() //%self.expect('frame var -d run-target -- index1', substrs=['(String.Index?) index1 = 2'])
+        //%self.expect('frame var -d run-target -- index_none', substrs=['(String.Index?) index_none = nil'])
+        //%self.expect('frame var -d run-target -- index2', substrs=['(String.Index?) index2 = 15'])
```
milseman, line 8: What about lazily-bridged NSStrings?

dcci: Do you have a specific example in mind? `firstIndex` doesn't seem to be defined on `NSString`(s)

source/Plugins/Language/Swift/SwiftFormatters.cpp
dcci: Here I only use `offset`, which is part of the ABI, so I guess it should be fine?. I can maybe just point to the structure in the file instead of inlining it here. What do you think?

packages/Python/lldbsuite/test/lang/swift/string_index/main.swift
milseman: Oh, do you have a test for Substring formatting? Do you want to be printing out "position"?

dcci: This is just the first part, it's a formatter for StringIndex. I'll build the Substring formatter on top of this.

milseman: Right, that's what we were talking about. We don't want to display the position, as it's encoding-specific and depends on whether the String was native or lazily bridged.

dcci: so you're saying that the formatter for `StringIndex` shouldn't exist? That sounds a bit awkward. If people call `firstIndex` and want to see the result, should we show the raw bits? That is a little weird..

dcci:
Closing this after discussing with Michael in person.

PR #1771, opened 2019-07-03
R49458356 dont link against debugsymbols framework

jasonmolenda:
Don't link against the DebugSymbols private framework; try to dlopen
    + dlsym the two functions we need from there at runtime.
    
    I'm not maintaining a negative cache if DebugSymbols is absent, so
    we'll try to dlopen() it on every call to
    LocateMacOSXFilesUsingDebugSymbols but this file is only built on
    mac and iOS type systems, so there's a slight perf impact running
    lldb on an iOS type system.
    
    I store the function pointer results in two global variables without
    any locking; two threads calling into LocateMacOSXFilesUsingDebugSymbols
    for the first time will both try to set these fptrs, but they'll be
    setting them to the same value, so I'm not too worried.
    
    I didn't see where in the cmake build configurations we link against
    DebugSymbols, but I removed the dependency from the xcode project
    file.
    
    <rdar://problem/49458356>


jasonmolenda:
@swift-ci please test

PR #1772, opened 2019-07-03
[Target] Centralize creation of scratch SwiftASTContexts

bulbazord:
I don't think it makes sense for ValueObject to know about SwiftASTContexts, so
we should centralize the creation logic into Target.

I know there are still references to SwiftASTContext in parts of ValueObject and its subclasses. I plan to try to factor those out  sometime later.

### DIFF
--- a/source/Target/Target.cpp
+++ b/source/Target/Target.cpp
@@ -2534,6 +2534,14 @@ SwiftASTContextReader Target::GetScratchSwiftASTContext(
   return SwiftASTContextReader(GetSwiftScratchContextLock(), swift_ast_ctx);
 }
 
+SwiftASTContextReader Target::GetScratchSwiftASTContext(Status &error,
+                                                        ValueObject &valobj,
+                                                        bool create_on_demand) {
+  ExecutionContext ctx = valobj.GetExecutionContextRef().Lock(false);
+  auto *exe_scope = ctx.GetBestExecutionContextScope();
+  return GetScratchSwiftASTContext(error, *exe_scope);
+}
+
 static SharedMutex *
 GetSwiftScratchContextMutex(const ExecutionContext *exe_ctx) {
   if (!exe_ctx)
--- a/include/lldb/Core/ValueObject.h
+++ b/include/lldb/Core/ValueObject.h
@@ -9,7 +9,6 @@
 #ifndef liblldb_ValueObject_h_
 #define liblldb_ValueObject_h_
 
-#include "lldb/Core/SwiftASTContextReader.h"
 #include "lldb/Core/Value.h"
 #include "lldb/Symbol/CompilerType.h"
 #include "lldb/Symbol/Type.h"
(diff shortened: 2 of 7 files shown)


bulbazord:
@swift-ci please test

bulbazord:
friendly ping @JDevlieghere @dcci @adrian-prantl

compnerd:
This makes sense to me.  Not sure if @adrian-prantl, @JDevlieghere, or @dcci have any objections.

source/Target/Target.cpp
```diff
@@ -2538,6 +2538,14 @@ SwiftASTContextReader Target::GetScratchSwiftASTContext(
   return SwiftASTContextReader(GetSwiftScratchContextLock(), swift_ast_ctx);
 }
 
+SwiftASTContextReader Target::GetScratchSwiftASTContext(ValueObject &valobj,
+                                                        bool create_on_demand) {
+  Status error;
```
JDevlieghere: I don't like this error getting swallowed.

source/Target/Target.cpp
```diff
@@ -2538,6 +2538,14 @@ SwiftASTContextReader Target::GetScratchSwiftASTContext(
...
+  ExecutionContext ctx = valobj.GetExecutionContextRef().Lock(false);
+  auto *exe_scope = ctx.GetBestExecutionContextScope();
+  return GetScratchSwiftASTContext(error, *exe_scope);
```
JDevlieghere, line 2542: Can we make this return an llvm::Expected?

JDevlieghere (commented):
I'm not an expert on the functionality, but I'd prefer for us to improve error handling while we're around.

source/Target/Target.cpp
bulbazord: Yeah I can do that :)

bulbazord: Okay so I just looked into it. Looks like We always return a SwiftASTContextReader no matter what, so returning an llvm::Expected doesn't make much sense here. What I did find is that `GetScratchSwiftASTContext` will sometimes fail to build a scratch SwiftASTContext and instead return the project-wide scratch SwiftASTContext. Even if we fail to build a SwiftASTContext, we still return a SwiftASTContextReader.

dcci: I have no strong preferences, so I'm happy with this once Jonas feels his comments have been addressed.
Maybe longer term we can enforce that the `SwiftASTContextReader` returned is always valid [and if it's not, assert], given we expect it in such state..

bulbazord:
@JDevlieghere Mind taking another look when you get the chance?

source/Target/Target.cpp
JDevlieghere: Do we ever need to differentiate between the two? This code is saying that the error doesn't matter. Is that true for this function, or is that true always? In the latter case, maybe we should remove the Status as the first argument?

bulbazord: From the looks of it, `Target::GetScratchSwiftASTContext` will return a Status and the callsites generally use them. It looks like error handling is just not taken into account with the ValueObject version. So even though it's true just for this function, I think that we can probably do a better job and improve error handling here by making the ValueObject version take a Status and then have the clients actually use that Status for error handling.

bulbazord:
@swift-ci please test

PR #1773, opened 2019-07-03
[Core] Remove redundant check in ValueObjectVariable::UpdateValue

bulbazord:
No need to check variable->GetType twice!


bulbazord:

PR #1778, opened 2019-07-03
[Symbol] Add DeclVendor::FindTypes

bulbazord:
[Symbol] Add DeclVendor::FindTypes

Summary:
Following up on the plan I outlined in D63622, we can remove the
dependence on clang in all the places where we only want to find the
types from the DeclVendor. This means that currently DeclVendor depends
on clang, but centralizing the dependency makes it easier to refactor
cleanly.

Differential Revision: https://reviews.llvm.org/D63853

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@364962 91177308-0d34-0410-b5e6-96231b3b80d8

This also contains a patch on top to account for swift changes.

### DIFF
--- a/include/lldb/Symbol/DeclVendor.h
+++ b/include/lldb/Symbol/DeclVendor.h
@@ -53,6 +53,19 @@ public:
                              uint32_t max_matches,
                              std::vector<clang::NamedDecl *> &decls) = 0;
 
+  /// Look up the types that the DeclVendor currently knows about matching a
+  /// given name.
+  ///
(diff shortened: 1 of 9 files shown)


bulbazord:
@swift-ci please test

dcci:
@xiaobai Hey, we noticed this broke one of the bots which still does Xcodebuild. As we need to unblock it to release a toolchain (cc: @shahmishal) I"m going to revert this and follow-up with you offline to figure out the next steps.

PR #1780, opened 2019-07-04
Plugins: permit building on Windows ARM64

compnerd:
Rather than relying on `sizeof(void *)` to determine the architecture,
use the `CMAKE_SYSTEM_PROCESSOR` variable.  This should allow us to
build for Windows and cross-compile.  Without this, we would attempt to
build the x64 plugin on ARM64 which would fail due to the `CONTEXT` type
being defined for ARM64 rather than `x64`.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@365155 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 74acd9b072c2c454690af165fbe2038348e94083)


compnerd:
@swift-ci please test

PR #1783, opened 2019-07-08
Revert "[Symbol] Add DeclVendor::FindTypes"

dcci:
This breaks the package bots. I'll follow up with @xiaobai to get this sorted out.


dcci:
@swift-ci test and merge

PR #1786, opened 2019-07-11
Make upstream-with-swift compilable again.

dcci:


dcci:
Can you please take a look before I merge this?

cmake/modules/LLDBStandalone.cmake
```diff
@@ -50,33 +50,26 @@ if (CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
...
+  if(CMAKE_CROSSCOMPILING)
+    set(LLVM_NATIVE_BUILD "${LLDB_PATH_TO_LLVM_BUILD}/NATIVE")
+    if (NOT EXISTS "${LLVM_NATIVE_BUILD}")
+      message(FATAL_ERROR
+        "Attempting to cross-compile LLDB standalone but no native LLVM build
+        found. Please cross-compile LLVM as well.")
+    endif()
+
+    if (CMAKE_HOST_SYSTEM_NAME MATCHES "Windows")
+      set(HOST_EXECUTABLE_SUFFIX ".exe")
```
JDevlieghere, line 62: I know you're just merging this, but why do we need this? Can't we use CMAKE_EXECUTABLE_SUFFIX? Or will that contain the cross compilation target's suffix?

dcci: This code is upstream, so if anything, that needs to change first?

bulbazord: I think this was my code originally. Is CMAKE_EXECUTABLE_SUFFIX determiend by CMAKE_HOST_SYSTEM_NAME or CMAKE_SYSTEM_NAME? That's where I was unsure.

But I agree with Davide, needs to change upstream before it makes its way here.

JDevlieghere: Yup, just wanted the raise the issue as I was looking at the code. Looks like it might be needed after all: https://gitlab.kitware.com/cmake/cmake/issues/17553

dcci:
If there are no additional comments, I'm going ahead and merging this so we can make progress on `upstream-with-swift`.

JDevlieghere (approved):

PR #1789, opened 2019-07-11
[SwiftExpressionParser] Remove a bunch of hacks in the handling of `s…

dcci:
…elf`.

Presumably there was a time where these were needed, but this is not true
anymore. I'm mainly motivated by the fact that every time I stop in
the expression parser hitting all this dead code I lose focus, but
this is generally good for readability and maintanability in general.

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
@@ -525,18 +525,9 @@ AddRequiredAliases(Block *block, lldb::StackFrameSP &stack_frame_sp,
       self_type = valobj_sp->GetCompilerType();
   }
 
-  if (!self_type.IsValid()) {
-    if (Type *type = self_var_sp->GetType()) {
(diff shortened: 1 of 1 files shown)


dcci:
@swift-ci test and merge

source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp
```diff
@@ -550,52 +541,11 @@ AddRequiredAliases(Block *block, lldb::StackFrameSP &stack_frame_sp,
       swift_runtime->DoArchetypeBindingForType(*stack_frame,
                                                imported_self_type);
```
compnerd, line 542: Since we overwrite `imported_self_type`, might be a good thing to assert that the archtetype binding succeeded (`assert(imported_type_type.isValid());`)

compnerd (approved):

JDevlieghere:
@swift-ci test and merge

slavapestov (approved):

PR #1790, opened 2019-07-12
[SwiftASTManipulator] Stop overriding access control.

dcci:
Defer this work to the compiler.

Also fixes <rdar://problem/40955478>


dcci:
@swift-ci test

jrose-apple (changes requested):
You can't disable access control in the REPL or Playgrounds, though.

dcci:
Fair, let me see what breaks if I bring that back.

PR #1793, opened 2019-07-15
[Swift/Optional] Fix the formatter to print the right value.

dcci:
There's a peculiarity in the way lldb represents `Int/UInt` types
in Swift. They don't have summaries, but values. This historical
choice forces us to switch between GetSummary() and GetValue()
in the formatters. Maybe we should revisit this choice because it
caused several issues over the last couple of years.

<rdar://problem/37835998>

### DIFF
--- a/lit/SwiftREPL/Optional.test
+++ b/lit/SwiftREPL/Optional.test
@@ -21,5 +21,18 @@ let y : Optional<Test> = Test(a: 1, b: Patatino.first)
 // CHECK-NEXT: b = first
 // CHECK-NEXT: }
 
+class A {
+  var x : Int
+  init(_ x : Int) {
+    self.x = x
+  }
+}
+
+var q : A? = A(23)
+
+// CHECK-NEXT: {{q}}: A? = 0x{{.*}} {
+// CHECK-NEXT:   x = 23
+// CHECK-NEXT: }
+
 let tinky : UInt8? = 250
 // CHECK-NEXT: {{tinky}}: UInt8? = 250
--- a/lit/SwiftREPL/Optional.test
+++ b/lit/SwiftREPL/Optional.test
@@ -20,3 +20,6 @@ let y : Optional<Test> = Test(a: 1, b: Patatino.first)
 // CHECK-NEXT: a = 1
 // CHECK-NEXT: b = first
 // CHECK-NEXT: }
+
+let tinky : UInt8? = 250
+// CHECK-NEXT: {{tinky}}: UInt8? = 250
--- a/source/Plugins/Language/Swift/SwiftOptional.cpp
+++ b/source/Plugins/Language/Swift/SwiftOptional.cpp
@@ -126,10 +126,13 @@ SwiftOptional_SummaryProvider_Impl(ValueObject &valobj, Stream &stream,
     return true;
   }
 
-  const char *value_summary = some->GetSummaryAsCString();
+  const char *summary = some->GetSummaryAsCString();
+  const char *value = some->GetValueAsCString();
 
-  if (value_summary)
-    stream.Printf("%s", value_summary);
+  if (summary)
+    stream.Printf("%s", summary);
+  else if (value)
+    stream.Printf("%s", value);
   else if (lldb_private::DataVisualization::ShouldPrintAsOneLiner(*some)) {
     TypeSummaryImpl::Flags oneliner_flags;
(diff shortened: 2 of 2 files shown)


dcci:
@swift-ci test

dcci:
@jimingham can you give a thorough look? Thanks!

jimingham (approved):
This seems fine as a short term fix.  

I wonder if a better fix might be to have the value providing summaries might also provide a summary?  The real point to the value providing Int's is you can do SBValue.GetValueAsUnsigned() and have that work for a swift Uint.  Having a summary string wouldn't get in the way of this - it is fine for ValueObjects to have both a value and a summary.  Then you wouldn't have to switch on code like this.

But if you want to do this for 5.1 this targeted fix is probably a better change.

dcci:
@jimingham I originally thought about that. Having value providing summaries also providing a summary seems a lot of work [basically, as much work as treating integers for what they really are in swift], and some amount of duplication in the logic. I'd rather go for the original plan we discussed.

That said, I wouldn't be opposed if somebody else decides to pick up the slack and implement the solution you suggest, I'm just advocating in favor of what I would implement [will?] when I have the time :)

fredriss:
Can one of you tell me what GetValueAsCString() actually does? What will it return for things that are not Swift integers?

jimingham:
For most swift struct values GetValueAsCString returns an empty string.  For classes it may return the pointer value for the class, I don't remember.  For swift Int's - because they have a "value providing summary" - GetValueAsCString will return the integer value as a string.

dcci:
> Can one of you tell me what GetValueAsCString() actually does? What will it return for things that are not Swift integers?

`nullptr`, for basically everything that's not a swift integer [and maybe float, but I haven't checked in a while]

fredriss:
> For most swift struct values GetValueAsCString returns an empty string. For classes it may return the pointer value for the class, I don't remember. For swift Int's - because they have a "value providing summary" - GetValueAsCString will return the integer value as a string.

A few things to check then:
- The patch check for a null pointer, is that what you mean by empty string?
- Does this mean an optional of a class will be printed differently now?

Basically I'd like to be sure this causes no functional change except for Integers.

dcci:
I checked earlier today and I'm fairly confident the return value in case 1) is a nullptr and not an empty string.
I also have reasons to believe that this doesn't change printing for an optional of a class, but I have to double-check.

jimingham:
For structs, GetValue is null.
For classes, GetValue returns the pointer value.

jimingham:
So I think Fred is right, this would cause us to print the pointer value for a class that had no summary whereas before we would have printed nothing.

dcci:
The test that I modified has coverage for structs and enums. `lit/SwiftREPL/Optional.test`
I think it's a fair point to test also other aggregates to make sure this doesn't break anything, so I'm going to add coverage for them before merging [actually, I'll check if we have it already and add in case we don't]

jimingham:
Note, however, since the Optional is set to print children, then we still print the address of a class.  Here's what you get for frame var today:

(tryme.Bar?) myBar = 0x0000000100303950 {
  b = 10
}

But in this case I'm pretty sure the pointer value and the rest are the child printing.

Class and struct optional printing is a little inconsistent anyway.  If I have:

```
struct Foo {
  let a = 10
}

class Bar {
  let b = 10
}

func doIt () {
 let myFoo : Foo? = Foo()
 let myRealFoo = myFoo!
 let myBar : Bar? = Bar()
 let myRealBar = myBar!
print(myFoo!, myBar!)
}

doIt()
```

Then in lldb I see:

```
(lldb) frame var
(tryme.Foo?) myFoo = some {
  a = 10
}
(tryme.Foo) myRealFoo = {
  a = 10
}
(tryme.Bar?) myBar = 0x0000000100303950 {
  b = 10
}
(tryme.Bar) myRealBar = 0x0000000100303950 {
  b = 10
}
```

Note that the filled optional of a struct says "some" but the filled optional of a class doesn't.

dcci:
I think this is something interesting to clean up regardless. Both structs (and classes, and enums, and ${whatever}) should print `some`. I personally never noticed this inconsistency, but I read the code and I realized it wouldn't be that hard to fix. It's unlikely it'll get to 5.1, though.

dcci:
Added a test for `class` optionals.

dcci:
@swift-ci test and merge

PR #1794, opened 2019-07-16
[Expression] Replace plugin header with non-plugin header

bulbazord:


bulbazord:

PR #1801, opened 2019-07-17
[cmake] Only set LLVM_CODESIGN_IDENTITY for repl_swift on Darwin

lanza:
This variable being set was causing cross compilations from X -> Darwin
to attempt to codesign using xcrun. Simply opt out unless we have a
usable xcrun.

### DIFF
--- a/tools/repl/swift/CMakeLists.txt
+++ b/tools/repl/swift/CMakeLists.txt
@@ -10,8 +10,12 @@ if( CMAKE_SYSTEM_NAME MATCHES "Linux" )
   set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib/swift/linux:${CMAKE_INSTALL_RPATH}")
 endif()
 
-# Override locally, so the repl is ad-hoc signed.
-set(LLVM_CODESIGNING_IDENTITY "-")
+# Only set LLVM_CODESIGNING_IDENTITY for building on Apple hosts for Apple
+# targets
+if (CMAKE_HOST_APPLE AND APPLE)
+  # Override locally, so the repl is ad-hoc signed.
+  set(LLVM_CODESIGNING_IDENTITY "-")
+endif()
 
 # Requires system-provided Swift libs.
 set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14.4)


lanza:
@xiaobai @JDevlieghere @weliveindetail

weliveindetail:
Sounds reasonable. So, it will not code-sign at all when cross-compiling?
Maybe this could use a short comment, as the cross-compilation case is not so obvious here.

lanza:
> Sounds reasonable. So, it will not code-sign at all when cross-compiling?
> Maybe this could use a short comment, as the cross-compilation case is not so obvious here.

Done!

lanza:
This build isn't going to work according to the results from my previous pull request. Python is broken while configuring llvm.

tools/repl/swift/CMakeLists.txt
```diff
@@ -10,8 +10,13 @@ if( CMAKE_SYSTEM_NAME MATCHES "Linux" )
   set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib/swift/linux:${CMAKE_INSTALL_RPATH}")
 endif()
 
-# Override locally, so the repl is ad-hoc signed.
-set(LLVM_CODESIGNING_IDENTITY "-")
+# LLVM_CODESIGNING_IDENTITY being set makes cmake attempt to invoke xcrun when
+# crosscopmiling to Darwin. So if we can't find xcrun then opt out of attempting
+# to do so.
+if (CMAKE_XCRUN)
```
JDevlieghere: I don't really like that we depend on an implementation detail of code signing (that it'll go through xcrun) to disable this. Is there something better we can check? Or even better, can we fix `llvm_codesign` in LLVM to take cross compiling into account?

lanza: `CMAKE_HOST_APPLE AND APPLE`? I can't say I know how the Apple code signing works to solve this, just that it's not required elsewhere.

JDevlieghere: What about using `CMAKE_HOST_APPLE` in https://github.com/llvm/llvm-project/blob/master/llvm/cmake/modules/AddLLVM.cmake#L1685 ?

weliveindetail: +1 but upstream maybe better fixed after the 9.0 branch cut

weliveindetail (approved):
The latest version LGTM. I like the `llvm_codesign` proposal too. I will have to strip down some of it anyway soon, which became unused with the LLDB install changes.

JDevlieghere:
FWIW I'm not on board with this. Now that the branch has been cut, can we fix `llvm_codesign`?

lanza:
Yea sure, we can check `CMAKE_HOST_APPLE` along with the `if(APPLE)` check in `llvm_codesign`.

PR #1802, opened 2019-07-17
[REPL] Add coverage for property wrappers in top-level code.

dcci:
They shouldn't crash the REPL, but emit a sensible error.

<rdar://problem/52655421>


dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

PR #1803, opened 2019-07-18
[cmake] adjust NATIVE build for Swift

lanza:


lanza:
@xiaobai

bulbazord:
I don't like the way this patch looks but I understand it's necessary for cross-compilation to work correctly and there aren't many other options right now. Merging.

PR #1805, opened 2019-07-18
[cmake] Add find_package(SWIFT) that was removed during a rebase

lanza:
A rebase of a change to LLDBStandalone.cmake lost the
find_package(SWIFT) invocation. Bring it back here.


lanza:
@xiaobai @JDevlieghere

PR #1806, opened 2019-07-18
Fall back to dyld's _dyld_start when no LC_MAIN / main() func can be found

jasonmolenda:
Fall back to dyld's _dyld_start when no LC_MAIN / main() func can be found

The new DriverKit user-land kernel drivers in macOS 10.15 / Catalina
do not have a main() function or an LC_MAIN load command.  lldb uses
the address of main() as the return address for inferior function
calls; it puts a breakpoint on main, runs the inferior function call,
and when the main() breakpoint is hit, lldb knows unambiguously that
the inferior function call ran to completion - no other function calls
main.

This change hoists the logic for finding the "entry address" from
ThreadPlanCallFunction to Target.  It changes the logic to first
try to get the entry address from the main executable module,
but if that module does not have one, it will iterate through all
modules looking for an entry address.

The patch also adds code to ObjectFileMachO to use dyld's
_dyld_start function as an entry address.

<rdar://problem/52343958>

Differential Revision: https://reviews.llvm.org/D64897

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@366493 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 48a57bc8555aa4cf70bd95c39b18b6fee9e58c79)

### DIFF
--- a/include/lldb/Target/Target.h
+++ b/include/lldb/Target/Target.h
@@ -1269,6 +1269,24 @@ public:
 
   lldb::addr_t GetPersistentSymbol(ConstString name);
 
+  /// This method will return the address of the starting function for
+  /// this binary, e.g. main() or its equivalent.  This can be used as
+  /// an address of a function that is not called once a binary has 
+  /// started running - e.g. as a return address for inferior function
+  /// calls that are unambiguous completion of the function call, not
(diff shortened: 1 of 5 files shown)


jasonmolenda:
@swift-ci please test

PR #1807, opened 2019-07-18
Add support to ProcessMachCore::DoLoadCore to handle an EFI UUID str.

jasonmolenda:
Add support to ProcessMachCore::DoLoadCore to handle an EFI UUID str.
    
    If a core file has an EFI version string which includes a UUID
    (similar to what it returns for the kdp KDP_KERNELVERSION packet)
    in the LC_IDENT or LC_NOTE 'kern ver str' load command.  In that
    case, we should try to find the binary and dSYM for the UUID
    listed.  The dSYM may have python code which knows how to relocate
    the binary to the correct address in lldb's target section load
    list and loads other ancillary binaries.
    
    The test case is a little involved,
    
    1. it compiles an inferior hello world apple (a.out),
    2. it compiles a program which can create a corefile manually
       with a specific binary's UUID encoded in it,
    3. it gets the UUID of the a.out binary,
    4. it creates a shell script, dsym-for-uuid.sh, which will
       return the full path to the a.out + a.out.dSYM when called
       with teh correct UUID,
    5. it sets the LLDB_APPLE_DSYMFORUUID_EXECUTABLE env var before
       creating the lldb target, to point to this dsym-for-uuid.sh,
    6. runs the create-corefile binary we compiled in step #2,
    7. loads the corefile from step #6 into lldb,
    8. verifies that lldb loaded a.out by reading the LC_NOTE
       load command from the corefile, calling dsym-for-uuid.sh with
       that UUID, got back the path to a.out and loaded it.
    
    whew!
    
    <rdar://problem/47562911>

### DIFF
--- a/packages/Python/lldbsuite/test/macosx/lc-note/kern-ver-str/Makefile
+++ b/packages/Python/lldbsuite/test/macosx/lc-note/kern-ver-str/Makefile
@@ -1,0 +1,15 @@
+LEVEL = ../../../make
+
+MAKE_DSYM := NO
+
+C_SOURCES := main.c
+
+all: a.out create-empty-corefile
+
+create-empty-corefile:
+	$(MAKE) VPATH=$(VPATH) -f $(SRCDIR)/create-empty-corefile.mk
+
+clean::
+	$(MAKE) -f create-empty-corefile.mk clean
+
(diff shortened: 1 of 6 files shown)


jasonmolenda:
@swift-ci please test

PR #1815, opened 2019-07-20
build: allow the user to specify `llvm-tblgen`

compnerd:
This follows the same pattern as Clang and permits the user to specify
the tablegen to use via `-DLLVM_TABLEGEN=`.  This allows for
cross-compiling LLDB for a foreign target (e.g. Windows ARM64 on Windows
X64).  The LLVM dependency for LLDB in that case must be a Windows ARM64
build which cannot cross-compile llvm-tblgen due to the way that Visual
Studio works.  Instead, permit the user to have a separate tablegen
build which can be used during the build.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@366639 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit e69fd7651e7a5c51c7ea64246dc31fcac1dc0b94)


compnerd:

PR #1819, opened 2019-07-22
Upstream with swift

plotfi:
Merging llvm.org lldb with upstream-with-swift. Resolved merge conflict with breakpoint_set_exception_typename/exception-typename (table/.td file fixed).


bulbazord:
cc @JDevlieghere @dcci

PR #1821, opened 2019-07-23
Upstream with swift

plotfi:
Merge with upstream llvm


plotfi:
@JDevlieghere @xiaobai

PR #1827, opened 2019-07-25
Cherry pick Fix CreateFunctionTemplateSpecialization to prevent dangling poiner to stack memory

shafik:
In ClangASTContext::CreateFunctionTemplateSpecializationInfo a TemplateArgumentList is allocated on the stack but is treated as if it is persistent in subsequent calls. When we exit the function func_decl will still point to the stack allocated memory. We will use TemplateArgumentList::CreateCopy instead which will allocate memory out of the DeclContext.

Differential Revision: https://reviews.llvm.org/D64777

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@366365 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit ce6bbd355f9b9e999b8e2681a55b1f758d28de56)

### DIFF
--- a/packages/Python/lldbsuite/test/expression_command/function_template_specialization_temp_args/Makefile
+++ b/packages/Python/lldbsuite/test/expression_command/function_template_specialization_temp_args/Makefile
@@ -1,0 +1,5 @@
+LEVEL = ../../make
+
+CXX_SOURCES := main.cpp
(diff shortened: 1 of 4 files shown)


shafik:
@swift-ci test

PR #1829, opened 2019-07-25
test: only pass `-sdk` on Darwin targets

compnerd:
On non-Darwin targets, the `-sdk` parameter may or may not contain the
sysroot which is needed to enable the use of `-sdk` to allow
cross-compilation.  This SDK contains the swift modules and the swift
runtime registrar.  Setting the SDKROOT to `/` on Linux requires that
the Swift SDK contents are installed to `/`.


compnerd:
@swift-ci please test

JDevlieghere (approved):
Sounds reasonable to me.

PR #1836, opened 2019-07-29
[Reproducers] Add support for Swift modules

JDevlieghere:
This add support for Swift modules and a corresponding test case. For
this to work we need to add a callback in Swift's dependency collector,
to notify LLDB's FileCollector. This will ensure the modulemap and
header end up in the reproducer.

### DIFF
--- a/lit/Reproducer/Swift/TestModule.test
+++ b/lit/Reproducer/Swift/TestModule.test
@@ -1,0 +1,37 @@
+# UNSUPPORTED: system-windows, system-freebsd
+
+# This tests replaying a Swift reproducer with bridging.
+
+# Start clean.
+# RUN: rm -rf %t.build && mkdir %t.build && cd %t.build
+# RUN: cp %S/Inputs/Module.swift %t.build/Module.swift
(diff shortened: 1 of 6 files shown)


lit/Reproducer/Swift/TestModule.test
```diff
@@ -0,0 +1,37 @@
...
+# RUN:          -module-cache-path %t.build/cache %t.build/Module.swift \
+# RUN:          -I. \
+# RUN:          -module-name main -o %t.out
+
+# Change to run directory.
+# RUN: rm -rf %t.repro
+# RUN: rm -rf %t.run && mkdir %t.run && cd %t.run
+
+# Capture the reproducer.
+# RUN: %lldb -x -b -s %S/Inputs/Module.in \
```
adrian-prantl, line 21: `%lldb` sets a module cache path already, right?

adrian-prantl (approved):
Changes seem to make sense.

lit/Reproducer/Swift/TestModule.test
JDevlieghere: Yep, I do believe so.

PR #1838, opened 2019-07-30

jrose-apple:
This feature deserves more tests, but this gets the "don't crash" part in place, and helped shake out early bugs on the compiler side. Continuing from #1414; I think it's ready to go now!

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/ReadMe.md
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/ReadMe.md
@@ -1,0 +1,11 @@
+These tests concern the experimental Swift feature `@_implementationOnly import`.\* Unlike a normal import, an `@_implementationOnly` import is guaranteed not to be part of a library's API or ABI, which means that client apps don't have to know about the import at all to use the library. However, the import *can* be used in the *implementation* of the library, so it's important for LLDB to handle the cases where the `@_implementationOnly` import is and isn't available when debugging.
+
+In a scenario with a client app, a library, and an implementation-only import that the library uses, there are three situations we care about:
+
+1. The library has no debug info, or just line tables. In this case, the implementation-only import is irrelevant; someone debugging the app can only use the public parts of the library.
+
+2. The library has debug info and the implementation-only import is available. In this case, the debugger should make everything available like it would with a normal import.
+
+3. The library has debug info, but the implementation-only import is not available (for whatever reason). In this case LLDB may still have to deal with internal parts of the library even though some types will not be available. (This is the least important case, but it'd be good to not crash or lie to users.)
+
+\* Hopefully someone will remember to edit this ReadMe before the feature goes public!
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/Makefile
@@ -1,0 +1,21 @@
+LEVEL = ../../../../make
+
+SWIFT_SOURCES=main.swift
+SWIFTFLAGS_EXTRAS= -I.
+LD_EXTRAS=-L.
+
+all: SomeLibraryCore.swiftmodule SomeLibrary.swiftmodule a.out
+
+include $(LEVEL)/Makefile.rules
+
+SomeLibraryCore.swiftmodule: SomeLibraryCore.swift
+	$(MAKE) BASENAME=$(shell basename $@ .swiftmodule) \
+		VPATH=$(SRCDIR) -f $(SRCDIR)/../dylib.mk all
+
+SomeLibrary.swiftmodule: SomeLibrary.swift SomeLibraryCore.swiftmodule
+	$(MAKE) BASENAME=$(shell basename $@ .swiftmodule) \
+		VPATH=$(SRCDIR) -f $(SRCDIR)/../dylib.mk all
+
+clean::
+	rm -rf *.o *.dylib *.so *.dll *.dSYM *.swiftdoc *.swiftmodule *.swiftinterface
+
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
@@ -1,0 +1,89 @@
+# TestLibraryIndirect.py
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2019 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ------------------------------------------------------------------------------
+"""
+Test `@_implementationOnly import` behind some indirection in a library used by the main executable
+"""
+import commands
+import lldb
+from lldbsuite.test.lldbtest import *
+from lldbsuite.test.decorators import *
+import lldbsuite.test.lldbutil as lldbutil
+import os
+import os.path
+import time
+import unittest2
+
+class TestLibraryIndirect(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+
+    def launch_info(self):
+        info = lldb.SBLaunchInfo([])
+
+        if self.getPlatform() == "freebsd" or self.getPlatform() == "linux":
+            # LD_LIBRARY_PATH must be set so the shared libraries are found on
+            # startup
+            library_path = os.environ.get("LD_LIBRARY_PATH", "")
+            if library_path:
+                library_path += ":"
+            library_path += self.getBuildDir()
+
+            info.SetEnvironmentEntries(["LD_LIBRARY_PATH=" + library_path], True)
+
+        return info
+
+    @swiftTest
+    def test_implementation_only_import_library(self):
+        """Test `@_implementationOnly import` behind some indirection in a library used by the main executable
+        
+        See the ReadMe.md in the parent directory for more information.
+        """
+        self.build()
+        def cleanup():
+            lldbutil.execute_command("make cleanup")
+        self.addTearDownHook(cleanup)
+        lldbutil.run_to_source_breakpoint(self, "break here", lldb.SBFileSpec("main.swift"), self.launch_info())
+
+        # This test is deliberately checking what the user will see, rather than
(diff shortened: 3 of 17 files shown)


jrose-apple:
@swift-ci Please test

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/ReadMe.md
```diff
@@ -0,0 +1,11 @@
+These tests concern the experimental Swift feature `@_implementationOnly import`.\* Unlike a normal import, an `@_implementationOnly` import is guaranteed not to be part of a library's API or ABI, which means that client apps don't have to know about the import at all to use the library. However, the import *can* be used in the *implementation* of the library, so it's important for LLDB to handle the cases where the `@_implementationOnly` import is and isn't available when debugging.
```
adrian-prantl, line 1: An even better location for this text is as a docstring in the test function (or class) in the .py file. IIRC, the test driver knows about them, too.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/Makefile
```diff
@@ -0,0 +1,27 @@
+LEVEL = ../../../../make
+
+EXE=a.out
+SWIFT_SOURCES=main.swift
+SWIFTFLAGS_EXTRAS= -I$(shell pwd)
```
adrian-prantl: Does `-I. ` also work?

adrian-prantl: Otherwise `Makefile.rules` defines `$(BUILDDIR)` as `$(shell pwd)`.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
```diff
@@ -0,0 +1,70 @@
...
+import os
+import os.path
+import time
+import unittest2
+
+class TestLibraryIndirect(TestBase):
+
+    mydir = TestBase.compute_mydir(__file__)
+    
+    @swiftTest
```
adrian-prantl, line 44: @JDevlieghere do we still use this decorator for anything?

JDevlieghere: Yep, it ensures we skip everything but macOS/Linux and no i386.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
```diff
@@ -0,0 +1,70 @@
...
+        
+        See the ReadMe.md in the parent directory for more information.
+        """
+        self.build()
+        def cleanup():
+            lldbutil.execute_command("make cleanup")
+        self.addTearDownHook(cleanup)
+        lldbutil.run_to_source_breakpoint(self, "break here", lldb.SBFileSpec("main.swift"))
+
+        self.expect("fr var", substrs=[
```
adrian-prantl, line 58: It may be less fragile to use self.frame().FindVariable("container") and then walk the value object tree programmatically instead of expecting the string output to match up perfectly. `TestObjCIvarDiscovery.py`  seems to have an example of doing that. There is also `lldbutil.check_variable()`.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_resilient/Makefile
```diff
@@ -0,0 +1,27 @@
+LEVEL = ../../../../make
+
+EXE=a.out
+SWIFT_SOURCES=main.swift
+SWIFTFLAGS_EXTRAS= -I$(shell pwd)
```
adrian-prantl: ditto

adrian-prantl:
You can pass a dictionary with environment variables to make in `self.build()`. (For example `TestCppValueCast.py`). Is it feasibly to reuse source files and test functions more?

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_resilient/TestLibraryResilient.py
```diff
@@ -0,0 +1,67 @@
+# TestLibrray.py
```
dcci: Small typo here.

dcci:
hmm, looks like this revealed a bug

```
14:01:22 AssertionError: False is not True : 'fr var' returns expected result, got 'value = <could not resolve type>
14:01:22 (main.ContainsTwoInts) container = (other = 10)
14:01:22 (Int) simple = 1'
```

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/ReadMe.md
jrose-apple: This is one level up in a folder of four separate tests. I don't want to duplicate it in several places, but I mentioned it in the docstrings for the tests.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
jrose-apple: I think you commented that last time, too. :-) I don't want that output; I want to see how the system accounts for the missing fields in how it's displayed to the user.

jrose-apple:
Huh, it worked locally. I guess I need to rebase and rebuild.

jrose-apple:
> You can pass a dictionary with environment variables to make in self.build(). (For example TestCppValueCast.py). Is it feasibly to reuse source files and test functions more?

I can combine the two MainExecutable ones for sure. The Library ones are testing different things, though, even if they share a lot.

jrose-apple:
Hm. The macOS tests aren't failing locally for me, even after I rebase. Also, any idea what's up with the Linux failures? It sounds like the breakpoint's not being hit at all, but I'm not sure why that would be.

adrian-prantl:
> The macOS tests aren't failing locally for me, even after I rebase. 

Wild guess but: Maybe there is nondeterminism in the order in which modules are imported (something like as rdar://problem/50667488)?

> Also, any idea what's up with the Linux failures?

Maybe it's because 

`return // break here`

doesn't result in code that would suffice as a breakpoint location on Linux. Try

```
func stop() {}
stop() // break here
```

which is safer. If that isn't it, perhaps the Makefile somehow causes the Linux version to drop debug info?

jrose-apple:
Okay, I think I've figured out the difference in results: it's based on whether the *main* executable is using a dSYM or not. I'm not sure *why* that would affect the results, though; any idea?

adrian-prantl:
Perhaps it's the .swiftinterface files being copied into the dSYM that aren't found in the DWARF case?

jrose-apple:
@swift-ci Please test

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/Makefile
jrose-apple: Oops, I forgot this one. I'll get it in the next rebase.

jrose-apple:
@swift-ci Please test macOS

jrose-apple:
Still stymied as to why this isn't working on Linux.

jrose-apple:
@swift-ci Please test Linux

jrose-apple:
@swift-ci Please test

jrose-apple:
Still needs some re-review about some shortcuts I took.

jrose-apple:
@swift-ci Please test

jrose-apple:
This PR seems to be cursed that only one platform can pass at a time.

@swift-ci Please test macOS

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
```diff
@@ -0,0 +1,147 @@
...
+            lldbutil.execute_command("make cleanup")
+        self.addTearDownHook(cleanup)
+        lldbutil.run_to_source_breakpoint(self, "break here", lldb.SBFileSpec("main.swift"), self.launch_info())
+
+        # FIXME: This particular test config is producing different results on
+        # different machines, but it's also the least important configuration
+        # (trying to debug something built against a library without that
+        # library present, implementation-only or not, and that library doesn't
+        # even have library evolution support enabled. Just make sure we don't
+        # crash.
```
jrose-apple, line 87: @adrian-prantl @dcci Is this acceptable? I don't like it but I've already sunk way too much time into these few tests.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_resilient/TestLibraryResilient.py
```diff
@@ -0,0 +1,88 @@
...
+            if library_path:
+                library_path += ":"
+            library_path += self.getBuildDir()
+
+            info.SetEnvironmentEntries(["LD_LIBRARY_PATH=" + library_path], True)
+
+        return info
+
+    @swiftTest
+    @expectedFailureOS(no_match(["macosx"])) # Requires Remote Mirrors support
```
jrose-apple, line 45: @adrian-prantl @dcci Got a better idea for this condition?

adrian-prantl: That condition LGTM.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/main_executable/TestMainExecutable.py
adrian-prantl: They produce different results, but they are all different kinds of failures?
We *are* checking for the existence of any error, so I guess that is okay.

jrose-apple: They are failing in different ways and it's really rather bizarre to me. I could file a Radar or Jira if you want it explicitly tracked somehow, but as noted it's not the most interesting configuration and I don't expect anyone to spend time tracking it down in practice.

adrian-prantl: As long as we are testing for the fact *that they fail* that is good enough at the moment.

jrose-apple:
Fixed the typo Davide spotted.

@swift-ci Please test

PR #1839, opened 2019-07-31
[SwiftLanguageRuntime] TypeDecoder wants a canonical name.

dcci:
Otherwise mangledName -> typeRef conversion fails, we can't retrieve
the type, and we can't print it.

SR-11226.


dcci:
@swift-ci test and merge

PR #1841, opened 2019-07-31
Use llvm::Expected when getting TypeSystems

compnerd:
Summary:
This commit achieves the following:
- Functions used to return a `TypeSystem *` return an
  `llvm::Expected<TypeSystem *>` now. This means that the result of a call
  is always checked, forcing clients to move more carefully.
- `TypeSystemMap::GetTypeSystemForLanguage` will either return an Error or a
  non-null pointer to a TypeSystem.

Reviewers: JDevlieghere, davide, compnerd

Subscribers: jdoerfert, lldb-commits

Differential Revision: https://reviews.llvm.org/D65122

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@367360 91177308-0d34-0410-b5e6-96231b3b80d8

### DIFF
--- a/include/lldb/Core/Module.h
+++ b/include/lldb/Core/Module.h
@@ -812,24 +812,8 @@ public:
 
   bool GetIsDynamicLinkEditor();
 
-  // This function must be called immediately after construction of the Module
-  // in the cases where the AST is to be shared.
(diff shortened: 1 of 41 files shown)


compnerd:
CC: @dcci @JDevlieghere @xiaobai

PR #1842, opened 2019-08-01
Revert "Implement ComputeSwiftResourceDirectory work on all platforms"

weliveindetail:
This reverts commit c7d5112a49d01cdbf740e52330b07d161f90e4dc to see if it fixes PR testing issues on Linux.


weliveindetail:
@swift-ci please test

dcci (commented):
[cc:ing the folks at Facebook who authored the change]
Can you please elaborate on what you think the issue is and why this revert would fix? 
This would be the minimum amount needed to justify the revert, and it's also useful for commit archeology.

dcci:
+ Slava who hit this issue on the bots a certain number of times.

weliveindetail:
@dcci Can we run this through swift-PR-Linux somehow?

weliveindetail:
And: as the initial comment states, this is mainly a test whether this change introduced the issue.

weliveindetail:
> Can you please elaborate on what you think the issue is and why this revert would fix?

In incremental builds LLDB's copy of the Swift resource directory doesn't get updated. This is either a feature or a bug of `add_custom_command_target()`. There were concerns about it in the review that were not acted upon:
`https://github.com/apple/swift-lldb/pull/1436/commits/c7d5112a49d01cdbf740e52330b07d161f90e4dc#diff-301dd16d95cf5bb591982af5b0879431R189`

The same mechanism exists for Clang resources but seems less critical as they change rarely.
Apparently, with this change the original location of the resource directory (or any other one) is not a fallback option anymore. Thus, builds fail in case of stale Swift resources from previous builds in PR testing now.

Possible fixes:
* run before each incremental build: `$ rm -rf build/buildbot_linux/ldb-linux-x86_64/lib/lldb`
* merge this PR
* copy the resources directories unconditionally

adrian-prantl:
@dcci see <rdar://problem/53766821>

dcci:
@compnerd @xiaobai @gmittert can we go ahead and merge this?

bulbazord:
If the CMake is the only issue, could we just revert the CMake portion? This provides windows functionality that we need. We can clean up the portion of code that looks in `lib/lldb/swift` since that won't be used anymore, but the implementation works so that if it's not there it will look elsewhere instead of failing immediately.

weliveindetail:
@xiaobai Sure, feel free to find a more appropriate fix. I think you made the right observation in the review already. Otherwise, let's merge this PR.

bulbazord (changes requested):
#1846

dcci:
Superseded by #1846

PR #1845, opened 2019-08-02
[Formatters] Temporarily disable libc++ std::function formatter due to performance issue

shafik:
Summary: We have been seeing increased reports of performance issue around large project and formatting std::function variables especially in functions signatures in back traces. There are some possible fixes but exploring those fixes may take time and it is better to temporarily disable the formatter due to its impact and re-enable it once we have a fix.

Differential Revision: https://reviews.llvm.org/D65666

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@367701 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit dd13fa23ee620c1d3b197389d902f0494a278889)


JDevlieghere (approved):

shafik:
@swift-ci test

PR #1846, opened 2019-08-02
[CMake] Always copy clang and swift resource directories when rebuilding

bulbazord:
Currently the clang and swift resource dirs are only copied during configure
time. This could be a problem for iterative builds of lldb, where the clang and/or swift
resource dir being built against could have changed. In order to make sure that
we have the most up-to-date clang and swift resource directories, we copy them
whenever we rebuild liblldb.

### DIFF
--- a/source/API/CMakeLists.txt
+++ b/source/API/CMakeLists.txt
@@ -194,7 +194,7 @@ endif()
 
 # Copy the clang resource directory.
 add_custom_command_target(
-    unused_var
+    copy-clang-resource-dir
     COMMAND "${CMAKE_COMMAND}" "-E" "copy_directory" "${CLANG_RESOURCE_PATH}" "${lib_dir}/lldb/clang/"
     OUTPUT "${lib_dir}/lldb/clang/"
     VERBATIM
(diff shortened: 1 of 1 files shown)


bulbazord:
cc @dcci @weliveindetail @JDevlieghere @compnerd @kusmour

JDevlieghere (approved):
Thanks!

bulbazord:
@swift-ci please test

dcci (approved):
LGTM

source/API/CMakeLists.txt
```diff
@@ -194,21 +194,23 @@ endif()
...
 add_custom_command_target(
-    unused_var
+    copy-swift-resource-dir
     COMMAND "${CMAKE_COMMAND}" "-E" "copy_directory" "${SWIFT_RESOURCE_PATH}" "${lib_dir}/lldb/swift/"
     OUTPUT "${lib_dir}/lldb/swift/"
     VERBATIM
     ALL
     DEPENDS ${SWIFT_RESOURCE_PATH})
 
+add_dependencies(liblldb copy-clang-resource-dir copy-swift-resource-dir)
```
adrian-prantl: I assume that liblldb and LLDB.framework are mutually exclusive, i.e., the resource dir isn't copied twice when building for macOS now?

bulbazord: It will be copied twice, but to different locations. Once to the framework location and once to `lib/lldb/swift`.

bulbazord:
@swift-ci please test

weliveindetail (approved):
If it works, go ahead.

For future readers of the history it may be interesting whether or not there was a particular reason to keep `add_custom_command_target()` (which has an incredibly complex implementation in Swift) and not use something much simpler instead, like https://github.com/apple/swift-lldb/blob/32fd43845af7/cmake/modules/LLDBFramework.cmake#L117

PR #1847, opened 2019-08-05
[CMake] Towards unified build support for swift-lldb

weliveindetail:
In-tree builds aka "unified builds" would allow LLDB to avoid Swift's build-script. With this patch unified builds should start working. Test configurations need some more efforts, but from here we could iterate.

```
> git clone https://github.com/apple/swift.git
> ./swift/utils/update-checkout --scheme=master-next
> cmake -B build/unified -C /path/to/lldb/cmake/caches/Apple-unified-macOS.cmake llvm
> ninja check-lldb
```


weliveindetail:
AFAIK we don't have unified build bots. CI tests will only validate that this doesn't break current builds.

weliveindetail:
With apple/swift#26490
@swift-ci please test

bulbazord (approved):
LGTM.

This is also something that we want. The only thing that currently doesn't work is the install rules for the swift components, which I have been working on.

weliveindetail:
Linux PR test failure is unrelated

weliveindetail:
> This is also something that we want. The only thing that currently doesn't work is the install rules for the swift components, which I have been working on.

Is the test suite working for you on macOS?

bulbazord:
> > This is also something that we want. The only thing that currently doesn't work is the install rules for the swift components, which I have been working on.
> 
> Is the test suite working for you on macOS?

Haven't tried to run the test suite yet unfortunately.

weliveindetail:
Had a compile error in my latest build. Looks like I missed at least one dependency. Will revisit that tomorrow. The approach was: build and fix what breaks. Maybe it needs a bit of time, until I went through all possible ninja schedules.

weliveindetail:
Apparently, PR testing is not working for `upstream-with-swift`.

weliveindetail:
@xiaobai 
> The only thing that currently doesn't work is the install rules for the swift components, which I have been working on.

Hmm how bad they don't use `llvm_distribution_add_targets`
What's your approach?

bulbazord:
> @xiaobai
> 
> > The only thing that currently doesn't work is the install rules for the swift components, which I have been working on.
> 
> Hmm how bad they don't use `llvm_distribution_add_targets`
> What's your approach?

My approach is to create targets for each of the swift components so that we can add the components to `LLVM_DISTRIBUTION_COMPONENTS`. The work is nearly done. I just finished working on https://github.com/apple/swift/pull/25118 (reviews needed!). After that you should be able to use a combination of `SWIFT_INSTALL_COMPONENTS` and `LLVM_DISTRIBUTION_COMPONENTS`. After that, I plan on making it so you can drop `SWIFT_INSTALL_COMPONENTS` during a unified build.

weliveindetail:
* depend on `swift-syntax-generated-headers` as we do on `clang-tablegen-targets`
* set `SWIFT_BUILD_RUNTIME_WITH_HOST_COMPILER` to avoid dependency to just-built Clang in Swift

With this LLDB builds reliably now (given a checkout that builds at all).

weliveindetail:
Rebased

PR #1849, opened 2019-08-06
[NameLookup:ASTScope] Disable ASTScopeLookup

davidungar:
ASTScope lookup does not work let with lldb, so fall back to context-based lookup.
rdar://53971116
Requires https://github.com/apple/swift/pull/26511

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -837,6 +837,9 @@ SwiftASTContext::SwiftASTContext(std::string description, llvm::Triple triple,
     : TypeSystem(TypeSystem::eKindSwift),
       m_compiler_invocation_ap(new swift::CompilerInvocation()),
       m_description(description) {
+  // rdar://53971116
+  m_compiler_invocation_ap->disableASTScopeLookup();
+
   // Set the clang modules cache path.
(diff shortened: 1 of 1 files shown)


davidungar:
@swift-ci please test

dcci (commented):
David, can you please elaborate what you mean here with does not work yet with lldb?
Do you see tests failing? Also, we generally don't end up putting rdar in comments but a more descriptive explanation of why things are disabled, so that external people can actually understand what's going on.

davidungar:
@dcci Thanks for your comment; sorry I merged before seeing it. I am getting ready to land the compiler with ASTScope lookup enabled by default. But the problem is that I'm getting source ranges from lldb that the ASTScope system cannot handle, causing test failures for the compiler PR. So, I am disabling ASTScope lookup when the compiler is invoked from lldb. At present, since ASTScope is disabled by default, this PR won't change anything. But when I land ASTScope lookups in the compiler, this PR will allow lldb to work with the compiler. The radar is a note to myself to go back and change lldb so that ASTScope can be enabled by default for it, too.

Let me know if you need to revert this, or if the explanation is unclear.

Thanks, Davide

dcci:
@davidungar I don't see any urgence in reverting this, in particular under the assumption that you're going back at it. I'm actually extremely curious about what's the failure mode, and which tests are failing. Also, do you need any help debugging/diagnosing? Please let me know.

PR #1850, opened 2019-08-06
[SymbolFile] Update after SymbolVendor API change

smeenai:
r368001 removed the passthrough buttons from SymbolVendor, so use
SymbolFile directly instead.


smeenai:
CC @compnerd @xiaobai @dcci

smeenai:
Looks like r368001 hasn't made its way to upstream-with-swift yet, so we may wanna wait for that merge.

PR #1851, opened 2019-08-06
 [SwiftASTContext] Change GetByteStride() to return an Optional.

dcci:
Before, we didn't have a way to notify whether the function
errored out or not. NFCI, for now.

<rdar://problem/54004017>

### DIFF
--- a/include/lldb/Symbol/ClangASTContext.h
+++ b/include/lldb/Symbol/ClangASTContext.h
@@ -760,7 +760,8 @@ public:
   GetBitSize(lldb::opaque_compiler_type_t type,
              ExecutionContextScope *exe_scope) override;
 
-  uint64_t GetByteStride(lldb::opaque_compiler_type_t type) override;
+  llvm::Optional<uint64_t>
+  GetByteStride(lldb::opaque_compiler_type_t type) override;
 
   lldb::Encoding GetEncoding(lldb::opaque_compiler_type_t type,
                              uint64_t &count) override;
--- a/source/Plugins/Language/Swift/SwiftArray.cpp
+++ b/source/Plugins/Language/Swift/SwiftArray.cpp
@@ -68,7 +68,7 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
     : m_metadata_ptr(LLDB_INVALID_ADDRESS),
       m_reserved_word(LLDB_INVALID_ADDRESS), m_size(0), m_capacity(0),
       m_first_elem_ptr(LLDB_INVALID_ADDRESS), m_elem_type(elem_type),
-      m_element_size(0), m_element_stride(elem_type.GetByteStride()),
+      m_element_size(0), m_element_stride(0),
       m_exe_ctx_ref(valobj.GetExecutionContextRef()) {
   if (native_ptr == LLDB_INVALID_ADDRESS)
     return;
@@ -84,9 +84,13 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
   ProcessSP process_sp(m_exe_ctx_ref.GetProcessSP());
   if (!process_sp)
     return;
+  auto element_stride = m_elem_type.GetByteStride();
   auto opt_size = elem_type.GetByteSize(process_sp.get());
   if (opt_size)
     m_element_size = *opt_size;
+  auto opt_stride = elem_type.GetByteStride();
+  if (opt_stride)
+    m_element_stride = *opt_stride;
   size_t ptr_size = process_sp->GetAddressByteSize();
   Status error;
   lldb::addr_t next_read = native_ptr;
@@ -194,7 +198,7 @@ SwiftArraySliceBufferHandler::GetElementAtIndex(size_t idx) {
 SwiftArraySliceBufferHandler::SwiftArraySliceBufferHandler(
     ValueObject &valobj, CompilerType elem_type)
     : m_size(0), m_first_elem_ptr(LLDB_INVALID_ADDRESS), m_elem_type(elem_type),
-      m_element_size(0), m_element_stride(elem_type.GetByteStride()),
+      m_element_size(0), m_element_stride(0),
       m_exe_ctx_ref(valobj.GetExecutionContextRef()), m_native_buffer(false),
       m_start_index(0) {
   static ConstString g_start("subscriptBaseAddress");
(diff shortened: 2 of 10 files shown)


dcci:
+ @lorentey as he wrote the formatters original [he didn't write this bug though :)]

dcci:
@swift-ci test

source/Plugins/Language/Swift/SwiftArray.cpp
```diff
@@ -68,7 +68,7 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
     : m_metadata_ptr(LLDB_INVALID_ADDRESS),
       m_reserved_word(LLDB_INVALID_ADDRESS), m_size(0), m_capacity(0),
       m_first_elem_ptr(LLDB_INVALID_ADDRESS), m_elem_type(elem_type),
-      m_element_size(0), m_element_stride(elem_type.GetByteStride()),
+      m_element_size(0), m_element_stride(0),
```
vedantk, line 71: I'm not sure about this. Above, in GetElementAtIndex, I see: `child_location = m_first_elem_ptr + idx * m_element_stride`

Perhaps it would be safer to remove the m_element_stride member and add a helper that returns Optional<size_t>. Or to store the stride as an Optional.

dcci: That would be a functional change, which I tried explicitly to avoid in this first pass.
I agree with you though. Probably we should just error out if we can't find a stride? I can do that as a follow-up.

dcci: I also think the same problem is true for size, which follows the same exact pattern.

vedantk: Sweet, no objections then.

source/Plugins/Language/Swift/SwiftArray.cpp
```diff
@@ -84,9 +84,13 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
   ProcessSP process_sp(m_exe_ctx_ref.GetProcessSP());
   if (!process_sp)
     return;
+  auto element_stride = m_elem_type.GetByteStride();
```
dcci, line 87: this is redundant and should be removed.

include/lldb/Symbol/ClangASTContext.h
```diff
@@ -760,7 +760,8 @@ class ClangASTContext : public TypeSystem {
   GetBitSize(lldb::opaque_compiler_type_t type,
              ExecutionContextScope *exe_scope) override;
 
-  uint64_t GetByteStride(lldb::opaque_compiler_type_t type) override;
+  llvm::Optional<uint64_t>
+  GetByteStride(lldb::opaque_compiler_type_t type) override;
```
adrian-prantl, line 764: Did you make this change in llvm.org, too? If not, please do so.

source/Plugins/Language/Swift/SwiftArray.cpp
```diff
@@ -84,9 +84,13 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
...
   auto opt_size = elem_type.GetByteSize(process_sp.get());
```
adrian-prantl, line 88: Personally, I'd have written this as 
```
if (Optional<uint64_t> stride = elem_type.GetByteStride())
  m_element_stride = *stride;
```
since the type is not obvious from the name `GetByteStride`

source/Plugins/Language/Swift/SwiftArray.cpp
```diff
@@ -194,7 +198,7 @@ SwiftArraySliceBufferHandler::GetElementAtIndex(size_t idx) {
 SwiftArraySliceBufferHandler::SwiftArraySliceBufferHandler(
     ValueObject &valobj, CompilerType elem_type)
     : m_size(0), m_first_elem_ptr(LLDB_INVALID_ADDRESS), m_elem_type(elem_type),
-      m_element_size(0), m_element_stride(elem_type.GetByteStride()),
+      m_element_size(0), m_element_stride(0),
```
adrian-prantl, line 201: Shouldn't m_element_stride be `Optional<uint64_t>`, too?

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -442,8 +432,15 @@ NativeHashedStorageHandler::NativeHashedStorageHandler(
   if (!m_process)
     return;
 
+  auto key_stride = key_type.GetByteStride();
```
adrian-prantl, line 435: same here... (type not obvious)

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -442,8 +432,15 @@ NativeHashedStorageHandler::NativeHashedStorageHandler(
...
+  if (key_stride) {
+    m_key_stride = *key_stride;
```
adrian-prantl, line 437: and here... (m_key_stride should be optional)

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -458,7 +455,10 @@ NativeHashedStorageHandler::NativeHashedStorageHandler(
       m_element_type = swift_ast->CreateTupleType(tuple_elements);
       auto *swift_type = reinterpret_cast<::swift::TypeBase *>(
           m_element_type.GetCanonicalType().GetOpaqueQualType());
-      m_key_stride_padded = m_element_type.GetByteStride() - m_value_stride;
+      auto element_stride = m_element_type.GetByteStride();
```
adrian-prantl, line 458: type not obvious

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -514,9 +514,13 @@ static bool GetObjectDescription_ObjectCopy(SwiftLanguageRuntime *runtime,
     static_type = runtime->DoArchetypeBindingForType(*frame_sp, static_type);
   }
 
+  auto stride = 0;
+  auto opt_stride = static_type.GetByteStride();
```
adrian-prantl, line 518: and here

include/lldb/Symbol/ClangASTContext.h
dcci: This API is only on github.

PR #1852, opened 2019-08-06
[ExpressionParser] Update for upstream API change

smeenai:
r368082 changed the Wrap argument to be an enum instead of a boolean.


smeenai:
CC @compnerd @xiaobai @dcci @JDevlieghere 

Note that r368082 isn't in upstream-with-swift yet.

PR #1856, opened 2019-08-08
[SwiftLanguageRuntime] Fix resilient types in hashed containers.

dcci:
The formatters for arrays and hash table called GetByteStride(),
which was only looking at the type _statically_, asking IRGen
about the properties.

This doesn't work for non-fixed-size types, e.g. resilient ones.
This patch implements the support to ask to the Runtime (i.e.
Remote mirrors about the properties).

### DIFF
--- a/lit/SwiftREPL/ResilientArray.test
+++ b/lit/SwiftREPL/ResilientArray.test
@@ -1,0 +1,9 @@
+// RUN: %lldb --repl < %s | FileCheck %s
+
+import Foundation
+
+let x : [URL] = [URL(string: "https://github.com")!, URL(string: "https://apple.com")!]
+// CHECK: {{x}}: [URL] = 2 values {
+// CHECK-NEXT:  [0] = "https://github.com"
+// CHECK-NEXT:  [1] = "https://apple.com"
+// CHECK-NEXT: }
--- a/lit/SwiftREPL/ResilientArray.test
+++ b/lit/SwiftREPL/ResilientArray.test
@@ -1,3 +1,4 @@
+// REQUIRES: system-darwin
 // RUN: %lldb --repl < %s | FileCheck %s
 
 import Foundation
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -6144,11 +6144,28 @@ SwiftASTContext::GetByteStride(lldb::opaque_compiler_type_t type,
                                ExecutionContextScope *exe_scope) {
   if (!type)
     return {};
+  swift::CanType swift_can_type(GetCanonicalSwiftType(type));
+  if (swift_can_type->hasTypeParameter()) {
+    if (!exe_scope)
+      return {};
+    ExecutionContext exe_ctx;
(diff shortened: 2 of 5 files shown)


dcci:
WIP! Don't merge! I need to write tests!

source/Symbol/SwiftASTContext.cpp
```diff
@@ -6049,11 +6049,28 @@ SwiftASTContext::GetByteStride(lldb::opaque_compiler_type_t type,
                                ExecutionContextScope *exe_scope) {
   if (!type)
     return {};
+  swift::CanType swift_can_type(GetCanonicalSwiftType(type));
```
adrian-prantl, line 6147: Can you factor out the code from `SwiftASTContext::GetBitSize()` into a helper function and use it here, too?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2591,6 +2591,13 @@ llvm::Optional<uint64_t> SwiftLanguageRuntime::GetBitSize(CompilerType type) {
   return type_info->getSize() * 8;
 }
 
+llvm::Optional<uint64_t> SwiftLanguageRuntime::GetByteStride(CompilerType type) {
+  auto *type_info = GetTypeInfo(type);
+  if (!type_info)
```
JDevlieghere:
```
if (auto *type_info = GetTypeInfo(type))
   return type_info->getStride();
return {};
```

source/Symbol/SwiftASTContext.cpp
dcci: It's not entirely obvious how to do it. I'll follow-up offline.

source/Target/SwiftLanguageRuntime.cpp
dcci: Done.

dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

dcci:
This is <rdar://problem/54062894>

lit/SwiftREPL/ResilientArray.test
```diff
@@ -0,0 +1,10 @@
+// REQUIRES: system-darwin
+// RUN: %lldb --repl < %s | FileCheck %s
+
+import Foundation
+
+let x : [URL] = [URL(string: "https://github.com")!, URL(string: "https://apple.com")!]
```
adrian-prantl, line 6: I think we've had this discussion before, but this is not the best way to test this change. It would be better to have a self-contained test that doesn't depend on Foundation and can run on all platforms, by also including the definition of the resilient type in the test.

source/Symbol/SwiftASTContext.cpp
adrian-prantl: I will give it a try!

lit/SwiftREPL/ResilientArray.test
dcci: Yes. Fair enough. I'll add a new test. Admittedly I should've done this here. Apologies.

PR #1857, opened 2019-08-08
[5.1][SwiftLanguageRuntime] Fix resilient types in hashed containers. 

dcci:
The formatters for arrays and hash table called GetByteStride(),
which was only looking at the type _statically_, asking IRGen
about the properties.

This doesn't work for non-fixed-size types, e.g. resilient ones.
This patch implements the support to ask to the Runtime (i.e.
Remote mirrors about the properties).

### DIFF
--- a/source/Plugins/Language/Swift/SwiftArray.cpp
+++ b/source/Plugins/Language/Swift/SwiftArray.cpp
@@ -68,7 +68,7 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
     : m_metadata_ptr(LLDB_INVALID_ADDRESS),
       m_reserved_word(LLDB_INVALID_ADDRESS), m_size(0), m_capacity(0),
       m_first_elem_ptr(LLDB_INVALID_ADDRESS), m_elem_type(elem_type),
-      m_element_size(0), m_element_stride(elem_type.GetByteStride()),
+      m_element_size(0), m_element_stride(0),
       m_exe_ctx_ref(valobj.GetExecutionContextRef()) {
   if (native_ptr == LLDB_INVALID_ADDRESS)
     return;
@@ -84,9 +84,13 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
   ProcessSP process_sp(m_exe_ctx_ref.GetProcessSP());
   if (!process_sp)
     return;
+  auto element_stride = m_elem_type.GetByteStride();
   auto opt_size = elem_type.GetByteSize(process_sp.get());
   if (opt_size)
     m_element_size = *opt_size;
+  auto opt_stride = elem_type.GetByteStride();
+  if (opt_stride)
+    m_element_stride = *opt_stride;
   size_t ptr_size = process_sp->GetAddressByteSize();
   Status error;
   lldb::addr_t next_read = native_ptr;
@@ -194,7 +198,7 @@ SwiftArraySliceBufferHandler::GetElementAtIndex(size_t idx) {
 SwiftArraySliceBufferHandler::SwiftArraySliceBufferHandler(
     ValueObject &valobj, CompilerType elem_type)
     : m_size(0), m_first_elem_ptr(LLDB_INVALID_ADDRESS), m_elem_type(elem_type),
-      m_element_size(0), m_element_stride(elem_type.GetByteStride()),
+      m_element_size(0), m_element_stride(0),
       m_exe_ctx_ref(valobj.GetExecutionContextRef()), m_native_buffer(false),
       m_start_index(0) {
   static ConstString g_start("subscriptBaseAddress");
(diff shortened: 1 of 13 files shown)


dcci:
@swift-ci test

dcci:
Note for the reviewers: this is the 5.1 version of the patch, where I just cherry-picked the minimum amount of changes needed. If you have comment about style, I would recommend check the `stable` version of this code too.

dcci:
```
20:44:47 
/Users/buildnode/jenkins/workspace/swift-lldb-PR-osx/lldb/source/Symbol/SwiftASTContext.cpp:6071:45: error: no member named 'Get' in 'lldb_private::SwiftLanguageRuntime'
20:44:47   if (auto *runtime = SwiftLanguageRuntime::Get(*exe_scope->CalculateProcess()))
20:44:47                       ~~~~~~~~~~~~~~~~~~~~~~^
20:44:47 1 error generated.
```

Looks like I need to adapt this for 5.1, but the general idea is there.

adrian-prantl (approved):
LGTM (in the sense of that it is a safe change for swift-5.1-branch).

source/Plugins/Language/Swift/SwiftArray.cpp
```diff
@@ -87,6 +87,9 @@ SwiftArrayNativeBufferHandler::SwiftArrayNativeBufferHandler(
   auto opt_size = elem_type.GetByteSize(process_sp.get());
   if (opt_size)
     m_element_size = *opt_size;
+  auto opt_stride = elem_type.GetByteStride(process_sp.get());
+  if (opt_stride)
+    m_element_stride = *opt_stride;
```
jimingham, line 92: If you failed to get a byte stride, don't you want to error out of the array formatter?  It seems very unlikely to work in that case.

source/Plugins/Language/Swift/SwiftArray.cpp
```diff
@@ -211,6 +214,10 @@ SwiftArraySliceBufferHandler::SwiftArraySliceBufferHandler(
   if (opt_size)
     m_element_size = *opt_size;
 
+  auto opt_stride = elem_type.GetByteStride(process_sp.get());
+  if (opt_stride)
```
jimingham, line 218: Same comment as above.

source/Plugins/Language/Swift/SwiftHashedContainer.cpp
```diff
@@ -442,8 +432,15 @@ NativeHashedStorageHandler::NativeHashedStorageHandler(
...
+  auto key_stride = key_type.GetByteStride(m_process);
+  if (key_stride) {
+    m_key_stride = *key_stride;
+    m_key_stride_padded = *key_stride;
+  }
+
   if (value_type) {
-    m_value_stride = value_type.GetByteStride();
+    auto value_type_stride = value_type.GetByteStride(m_process);
+    m_value_stride = value_type_stride ? *value_type_stride : 0;
```
jimingham, line 443: Ditto.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -6044,14 +6044,33 @@ SwiftASTContext::GetBitSize(lldb::opaque_compiler_type_t type,
...
+    ExecutionContext exe_ctx;
+    exe_scope->CalculateExecutionContext(exe_ctx);
+    auto swift_scratch_ctx_lock = SwiftASTContextLock(&exe_ctx);
+    CompilerType bound_type = BindAllArchetypes({this, type}, exe_scope);
+    // Note thay the bound type may be in a different AST context.
+    return bound_type.GetByteStride(exe_scope);
   }
-  return 0;
+
+  const swift::irgen::FixedTypeInfo *fixed_type_info =
```
jimingham, line 6064: This looks like the cheapest of the three ways of getting the byte size.  Shouldn't we try it first?

jimingham (approved):
Except for merging over the Get function, this is fine.  I'm approving this provisional to that of course...

I added some general review comments.  Particularly now that you are returning an Optional for GetByteStride, you can tell the difference between failing to get a byte stride and getting a 0 byte stride - whatever that means.  So it would be better to error out in the Formatters if you fail to get a byte stride.  At present you continue on with an artificial 0 value, which will almost sure mean we print junk for the elements.

I think all these could be addressed on stable rather than here, they're just details.  But I'm just putting them here because I didn't see a PR for review on stable...

dcci:
@swift-ci test

source/Symbol/SwiftASTContext.cpp
dcci: We always need to bind the archetypes, otherwise we'll go to IRGen and assert miserably.
Also, I'm not sure what's cheaper between going to IRGen or asking mirrors. Maybe the latter, but I haven't profiled either.

source/Plugins/Language/Swift/SwiftArray.cpp
dcci: *Nod*. I'll make this change on stable.

dcci:
I tried to abstract and Adrian tried too. 
The result is not really satisfactory ...
https://github.com/apple/swift-lldb/pull/1859

source/Symbol/SwiftASTContext.cpp
jimingham: Okay.

dcci:
This is rdar://problem/54062894

PR #1858, opened 2019-08-08
Add -Werror and -Wdocumentation

weliveindetail:
```
[CMake] Add option LLDB_ENABLE_WERROR
[CMake] Build LLDB.framework with -Wdocumentation
```
rdar://problem/47921376


JDevlieghere:
Should upstream have `-Wdocumentation` as well?

weliveindetail:
> Should upstream have `-Wdocumentation` as well?

Yes, why not. Committed upstream with: https://reviews.llvm.org/rG2c72b4175756
Removed it from the PR and rebased on latest upstream-with-swift.

weliveindetail:
For more background on why ``-Werror`` is not upstream, please see https://reviews.llvm.org/D65939

PR #1859, opened 2019-08-08
Abstract common functionality from GetBitSize and GetByteStride.

adrian-prantl:
Here's my attempt at factoring out the common pattern in these functions. Let me know what you think!


adrian-prantl:
@swift-ci test

dcci:
I thought about using lambdas + function_ref<> but... is this the right tradeoff?
Is it true that this shares more code, but it also adds more code, and it's much harder to read than before.

adrian-prantl:
frankly, I'm not sure either. Are there more functions that share this pattern?

dcci:
Not I'm aware of.

PR #1862, opened 2019-08-08
[DWARF/Swift] All the mangled names are now in DWARF.

dcci:
No need to do an AST lookup.

<rdar://problem/48277264>


dcci:
@swift-ci test and merge

source/Plugins/SymbolFile/DWARF/DWARFASTParserSwift.cpp
```diff
@@ -120,14 +120,6 @@ lldb::TypeSP DWARFASTParserSwift::ParseTypeFromDWARF(const SymbolContext &sc,
     }
     if (SwiftLanguageRuntime::IsSwiftMangledName(name.GetCString()))
       mangled_name = name;
-    else {
-      const char *type_name_cstr = name.GetCString();
```
adrian-prantl, line 124: I have actually seen instances on types without mangled names in recently generated DWARF, however they were always forward declarations that only existed to serve as the DeclContext of a function. So I still think this is safe.

PR #1863, opened 2019-08-08
[5.1][SwiftLanguageRuntime] TypeDecoder wants a canonical name.

dcci:
Otherwise mangledName -> typeRef conversion fails, we can't retrieve
the type, and we can't print it.

SR-11226.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/optional_url/main.swift
+++ b/packages/Python/lldbsuite/test/lang/swift/optional_url/main.swift
@@ -1,0 +1,4 @@
+import Foundation
+
+let u : URL? = URL(string: "https://github.com")
+print(u) //%self.expect('po u', substrs=['https://github.com'])
--- a/source/Target/SwiftLanguageRuntime.cpp
+++ b/source/Target/SwiftLanguageRuntime.cpp
@@ -2558,7 +2558,8 @@ lldb::addr_t SwiftLanguageRuntime::FixupAddress(lldb::addr_t addr,
 const swift::reflection::TypeInfo *
 SwiftLanguageRuntime::GetTypeInfo(CompilerType type) {
   swift::CanType swift_can_type(GetCanonicalSwiftType(type));
-  ConstString mangled_name(type.GetMangledTypeName());
+  CompilerType can_type(swift_can_type);
+  ConstString mangled_name(can_type.GetMangledTypeName());
   StringRef mangled_no_prefix =
       swift::Demangle::dropSwiftManglingPrefix(mangled_name.GetStringRef());
   swift::Demangle::Demangler Dem;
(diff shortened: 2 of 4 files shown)


dcci:
@adrian-prantl @jimingham what do you think about this going in 5.1 ?

packages/Python/lldbsuite/test/lang/swift/optional_url/main.swift
```diff
@@ -0,0 +1,4 @@
+import Foundation
+
+let u : URL? = URL(string: "https://github.com")
+print(u) //%self.expect('po u', substrs=['https://github.com'])
```
adrian-prantl, line 4: As usual, a test without foundation would be better.

adrian-prantl: Does the test also work with frame variable?

source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -2558,7 +2558,8 @@ lldb::addr_t SwiftLanguageRuntime::FixupAddress(lldb::addr_t addr,
 const swift::reflection::TypeInfo *
 SwiftLanguageRuntime::GetTypeInfo(CompilerType type) {
   swift::CanType swift_can_type(GetCanonicalSwiftType(type));
-  ConstString mangled_name(type.GetMangledTypeName());
+  CompilerType can_type(swift_can_type);
+  ConstString mangled_name(can_type.GetMangledTypeName());
```
adrian-prantl, line 2562: This seems safe to me.

packages/Python/lldbsuite/test/lang/swift/optional_url/main.swift
dcci: No, the formatter is broken for other reasons. But this at least fixes `po`.

adrian-prantl: Could you please file a bug for that, too?

jimingham:
This restriction doesn't appear anywhere in TypeDecoder.h.  If it is a requirement (I'm not doubting that you know that, but...) then it should be stated somewhere.  And also it seems to me if it is going to require that the API should take a CanType and do the couple of lines of demangling itself.  Either that or if you can tell a CanType from the demangling node it would give an error if you didn't give it one.  The API on the swift side seems like a little trap waiting to bite unsuspecting developers.

OTOH, that doesn't seem like a change for 5.1.  If this is required then this seems a safe enough way to do it.

adrian-prantl:
It's an API for mangled names and presumably (implicitly) does not accept names mangled with the special debugger mangling. That said, it would be better to fix the API to also accept debugger-mangled (= sugared) names.

dcci:
@swift-ci test

packages/Python/lldbsuite/test/lang/swift/optional_url/main.swift
dcci: There's already a rdar. <rdar://problem/51780502>

PR #1864, opened 2019-08-08
DWARFmporterDelegate: Fix type lookup in DWARF (= no .dSYM) projects.

adrian-prantl:
The SymbolFileDWARF change has also been made on llvm.org as r368345.


adrian-prantl:
@swift-ci test and merge

PR #1869, opened 2019-08-09
DWARFImporterDelegate: Support enums.

adrian-prantl:
Also includes a tiny improvement to SwiftOptionSetSummaryProvider to
avoid printing the zero-case of an option set.


adrian-prantl:
@swift-ci test and merge

PR #1871, opened 2019-08-09
[SwiftASTContext] Only a variant of CreateTupleType is used.

dcci:
Remove the other.


dcci:
@swift-ci test and merge

PR #1872, opened 2019-08-09
[Dematerializer] Remove unneded DoArchetypeBinding().

dcci:
The (de)materializer works on variables that have been already
materialized, and fully realized. We don't need to perform archetype
binding, because the materializer did it already.


dcci:
@swift-ci test

bulbazord (approved):
I've been wanting to get rid of this for a while. Thank you!

adrian-prantl (commented):
Nice. As discussed offline, we think this is redundant since GetByteStride & friends do this anyway.

PR #1873, opened 2019-08-09
Document DWARFImporter's inability to import Objective-C Protocol types.

adrian-prantl:
Protocols are not emitted into DWARF by Clang. We may be able to
salvage them as dynamic types using the AppleObjectiveCDeclVendor.

rdar://problem/49233932


adrian-prantl:
@swift-ci test and merge

PR #1885, opened 2019-08-14
Fix merge conflicts when merging upstream-with-swift into stable

Teemperor:
rdar://54281876


dcci:
it's the other way around. `stable` -> `upstream-with-swift`.

PR #1894, opened 2019-08-16
Re-remove call to DoArchetypeBindingForType

bulbazord:


dcci:
When was this reintroduced? Please remove also the header reference.

dcci (approved):

bulbazord:
Unsure. The header is already gone, which is how I caught this in the first place.

dcci:
anyway, thanks!

PR #1900, opened 2019-08-17
[Core] Change the underlying implementation of SharedMutex

bulbazord:
I'm making this change for several reasons:
- llvm::sys::RWMutexImpl is effectively gone except in certain situations. Using
  it directly is no longer feasible. Instead use llvm::sys::SmartRWMutex. This
  is better than using a std::shared_timed_mutex directly because older versions
  of macOS don't expose this symbol.
- There is a race condition with try_lock. We need to guard any use of
  `m_readers`.

Also, I think swift-lldb/upstream-with-swift should build after this, so I will trigger a build.

### DIFF
--- a/include/lldb/Core/SwiftASTContextReader.h
+++ b/include/lldb/Core/SwiftASTContextReader.h
@@ -21,23 +21,33 @@ class SwiftASTContext;
 class ExecutionContext;
 class ExecutionContextRef;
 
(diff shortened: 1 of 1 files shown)


bulbazord:
@swift-ci please test

JDevlieghere (approved):

PR #1903, opened 2019-08-20
[Symbol] Remove swift from CompilerType

bulbazord:
I removed clang from CompilerType, so removing swift seems like the natural
progression.

### DIFF
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -739,6 +739,8 @@ public:
                                 lldb::StackFrameWP &stack_frame_wp,
                                 swift::SourceFile *source_file, Status &error);
 
+  static CompilerType GetCompilerType(swift::Type swift_type);
+
 protected:
   /// This map uses the string value of ConstStrings as the key, and the TypeBase
   /// * as the value. Since the ConstString strings are uniqued, we can use
--- a/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
@@ -749,7 +749,7 @@ void SwiftASTManipulator::FindVariableDeclarations(
     auto type = var_decl->getDeclContext()->mapTypeIntoContext(
         var_decl->getInterfaceType());
     persistent_info.m_name = name;
-    persistent_info.m_type = {type.getPointer()};
+    persistent_info.m_type = SwiftASTContext::GetCompilerType(type);
     persistent_info.m_decl = var_decl;
 
     m_variables.push_back(persistent_info);
@@ -815,7 +815,7 @@ void SwiftASTManipulator::InsertResult(
     SwiftASTManipulator::ResultLocationInfo &result_info) {
   swift::ASTContext &ast_context = m_source_file.getASTContext();
 
-  CompilerType return_ast_type(result_type.getPointer());
+  CompilerType return_ast_type = SwiftASTContext::GetCompilerType(result_type);
 
   result_var->overwriteAccess(swift::AccessLevel::Public);
   result_var->overwriteSetterAccess(swift::AccessLevel::Public);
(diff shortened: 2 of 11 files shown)


bulbazord:
@swift-ci please test

bulbazord:
Hmm, looks like the swift-lldb merge is ahead of the swift-llvm merge. LLDB is using `llvm::Regex::isValid()`, but the patch adding that hasn't yet made it into swift-llvm.

adrian-prantl (changes requested):
What's the motivation behind this patch?

Could you achieve the same by perhaps having a 
`class SwiftCompilerType : public CompilerType` that does provide this constructor?

adrian-prantl:
Can you post a link to the review were you removed clang from CompilerType? I'd like to read up on that to better understand what's going on.

bulbazord:
> What's the motivation behind this patch?
> 
> Could you achieve the same by perhaps having a
> `class SwiftCompilerType : public CompilerType` that does provide this constructor?

My motivation is more decoupling of swift from non-plugin libraries (lldbSymbol in this case). I think that we could probably have a SwiftCompilerType. If that's what we went with, I'd like for it to live in the SwiftExpressionParser plugin to avoid adding more swift-specific bits to non-plugin libraries.

> Can you post a link to the review were you removed clang from CompilerType? I'd like to read up on that to better understand what's going on.

Review was here: https://reviews.llvm.org/D66102

adrian-prantl:
Thanks, I posted the same question on the llvm review. Whatever interface we decide on there should probably also be used here. That said, I don't think this necessarily needs to hold up this patch.

bulbazord:
Sounds good. Whatever happens on llvm.org, I'll mirror here for swift. I think landing this in the meantime should be okay (assuming I can get a build on swift-ci).

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
```diff
@@ -814,7 +816,8 @@ void SwiftASTManipulator::InsertResult(
     SwiftASTManipulator::ResultLocationInfo &result_info) {
   swift::ASTContext &ast_context = m_source_file.getASTContext();
 
-  CompilerType return_ast_type(result_type.getPointer());
+  CompilerType return_ast_type(SwiftASTContext::GetSwiftASTContext(&ast_context),
+                               result_type.getPointer());
```
adrian-prantl: This is dangerous, because it is not necessarily equivalent to the original code. Without having read the context, there is no guarantee that `ast_context` is equivalent to `result_type->getASTContext()`. There are many concurrent SwiftASTContexts floating around.

adrian-prantl (changes requested):

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
bulbazord: I see. In that case, we should almost always be asking the swift::Type for its ASTContext when constructing a CompilerType then?

bulbazord: If that's the case, then a wrapper or something would be a great idea. It would cut down on boilerplate significantly.

adrian-prantl:
Yeah, I feel like the wrapper makes much more sense in the Swift case than in the Clang case.

bulbazord:
@swift-ci please test

adrian-prantl (approved):
Ah, so the wrapper already was there?

bulbazord:
I just added the wrapper actually. Not too bad to do it, even if a bit repetitive.

bulbazord:
@swift-ci please test

bulbazord:
@swift-ci please test

bulbazord:
@swift-ci please test

bulbazord:
@swift-ci please test

bulbazord:
@swift-ci please test

dcci (approved):

include/lldb/Symbol/SwiftASTContext.h
```diff
@@ -725,6 +725,8 @@ class SwiftASTContext : public TypeSystem {
                                 lldb::StackFrameWP &stack_frame_wp,
                                 swift::SourceFile *source_file, Status &error);
 
+  static CompilerType GetSwiftCompilerType(swift::Type swift_type);
```
adrian-prantl: isn't the "Swift" in the name redundant?
Should it be `SwiftASTContext::GetCompilerType(...)`?

bulbazord: Yeah, it is somewhat redundant. I need to update this PR, I can rename it when I do that.

bulbazord:
@swift-ci please test

bulbazord:
@swift-ci please test

bulbazord:
@swift-ci please test

bulbazord:
@swift-ci please test

PR #1907, opened 2019-08-22
DWARFImporterDelegate: Adopt the new FindFiles API

adrian-prantl:
By using the new LLDB FindFiles API that llows for fuzzy CompilerContext matching, we can unify the two lookup implementations, dramatically improve the performance by only looking types up once, and look for types inside of unknown submodules.
    
rdar://problem/49233932
rdar://problem/54471165


adrian-prantl:
@swift-ci test and merge

PR #1908, opened 2019-08-22
[Misc] Replace std::once_flag with llvm::once_flag.

dcci:
The latter is what's used upstream and apparently some platforms
have bugs in the std implementation.
This reduces the diff with llvm.org


dcci:
@swift-ci test and merge

PR #1910, opened 2019-08-22
[Host/macox] Remove an unneeded #ifdef static_analyzer.

dcci:
Reduces the difference with llvm.org/lldb.


dcci:
@swift-ci test and merge

PR #1911, opened 2019-08-22
[SBThread] lldb_private functions are not instrumented.

dcci:
This was a mismerge from llvm.org. This commit gets rid of
an unneeded difference between github and llvm.org


dcci:
@swift-ci test and merge

JDevlieghere:
Thanks!

PR #1912, opened 2019-08-22
[Scalar] Remove a now unnecessary check.

dcci:
Reduces the difference between llvm.org and github.

<rdar://problem/35898017>


dcci:
@swift-ci test and merge

PR #1913, opened 2019-08-23
[ReproducerInstrumentation] Remove stale Recorder::Log.

dcci:
This function doesn't exist anymore on llvm.org. My understanding
is that https://reviews.llvm.org/D60984 removed it, but maybe
was mismerged to github.

### DIFF
--- a/source/Utility/ReproducerInstrumentation.cpp
+++ b/source/Utility/ReproducerInstrumentation.cpp
@@ -119,13 +119,4 @@ Recorder::~Recorder() {
   UpdateBoundary();
 }
 
-void Recorder::Log(unsigned id) {
-#ifndef LLDB_REPRO_INSTR_TRACE
-  LLDB_LOG(GetLogIfAllCategoriesSet(LIBLLDB_LOG_API), "Recording {0}: {1}", id,
-           m_pretty_func);
-#else
-  llvm::errs() << "Recording " << id << ": " << m_pretty_func << "\n";
-#endif
-}
-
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci test and merge

source/Utility/ReproducerInstrumentation.cpp
```diff
@@ -119,13 +119,4 @@ Recorder::~Recorder() {
   UpdateBoundary();
 }
 
-void Recorder::Log(unsigned id) {
```
fredriss, line 122: How did this compile in the first place? Is there a matching declaration in a header that should be removed too?

dcci:
@swift-ci test test osx platform

dcci:
@swift-ci test test macos platform

source/Utility/ReproducerInstrumentation.cpp
dcci: Yes. There's a matching call in the header that needs to be removed.

JDevlieghere: Yeah we should update the PR to remove that as well.

dcci:
Removed the decl.

dcci:
@swift-ci test and merge

JDevlieghere (approved):

dcci:
@swift-ci test test macos platform

dcci:
@swift-ci test macos

dcci:
@swift-ci test macos platform

PR #1914, opened 2019-08-23
[CMake] Pass the correct CMake directory for Swift

JDevlieghere:
Swift sets its module directory through `SWIFT_CMAKE_DIR`. We are
currently incorrectly using `Swift_DIR` instead. This is different than
LLVM, which is special because its modules live in the `LLVM_DIR`.


bulbazord (approved):
Thank you!

PR #1915, opened 2019-08-23
[Plugins/Target] Remove differences from upstream.

dcci:
-> Sort headers correctly.
-> Remove an unneded shared_from_this() call.


dcci:
@swift-ci test and merge

PR #1922, opened 2019-08-27
[Swift] Update tests now that we print the swiftified type.

dcci:
<rdar://problem/54730382>


dcci:
@swift-ci test and merge

PR #1923, opened 2019-08-27
Comment the complex interaction between Swift and DWARFimporterDelegate.

adrian-prantl:
<rdar://problem/49233932>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -3188,7 +3188,7 @@ class SwiftDWARFImporterDelegate : public swift::DWARFImporterDelegate {
     }
     return nullptr;
   }
-  
+
   static CompilerContextKind
   GetCompilerContextKind(llvm::Optional<swift::ClangTypeKind> kind) {
     if (!kind)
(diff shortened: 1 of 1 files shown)


adrian-prantl:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3218,6 +3218,60 @@ class SwiftDWARFImporterDelegate : public swift::DWARFImporterDelegate {
...
+  /// There are two primary ways that this delegate method is called:
+  ///
+  ///    1. When resolving a type from a mangled name. In this case \p
+  ///       kind will be known, but the owning module of a Clang type
+  ///       in a mangled name is always __ObjC or __C.
+  ///
+  ///    2. When resolving a type from a serialized module
+  ///       cross reference. In this case \c kind will be unspecified,
+  ///       but the (top-level) module that the type is defined in
+  ///       will be known.
```
jrose-apple, line 3232: I'm not sure how much we want to trust this module, though. ObjC lets you move things between modules, and I consider it a bug that Swift is *this* picky about it when we're talking about a member.

PR #1925, opened 2019-08-27
[CommandObjectExpression] Remove swift-only option.

dcci:
This is probably unused, only enabled under DEBUG, and in
a central place [so it causes merge conflicts]. Another difference
between llvm.org and github gone.


dcci:
@swift-ci test and merge

PR #1931, opened 2019-08-28
Future-proofing: Added check for Python 3.8 and higher

tachoknight:
where platform.linux_distribution() is no longer available and causes the build to fail. Python 3.8 is scheduled to be released this year per https://www.python.org/dev/peps/pep-0569/.


dcci:
Fair dinkum.

dcci:
@swift-ci test and merge

PR #1932, opened 2019-08-28
[InstrumentationRuntime] Remove duplicated plugin implementation.

dcci:
One is enough.


dcci:
I'm going ahead and try to test and merge this, but your review is of course appreciated.

dcci:
@swift-ci test and merge

vedantk:
If the idea is to move this upstream, this is fine, but I don’t think the plugin itself is duplicated.

dcci:
I'm confused by your reply. The plugin is already upstream, it just has a different name.
`UBSan` VS `UndefinedBehaviour`. Downstream had both, and one was unused.

vedantk:
Yikes. Now this makes sense. Thank you.

dcci:
No worries :) Thank you!

PR #1933, opened 2019-08-28
[ClangASTSource] Remove divergence between upstream and swift-lldb.

dcci:


dcci:
@swift-ci test

shafik:
This looks ok to me I am looking in Decl and DeclBase and they both have asserts on `hasExternalLexicalStorage`. I don't see another reason why this would be bad. Let's see what Raphael has to say.

Teemperor (approved):
Seems like this was added during the initial 60k lines "add swift support" commit ( 84ef5bf74617b9506b6dbec436d9501910f94bfa ), so there is not really any information on why this was changed in the first place. But this check seems redundant to me, so LGTM.

PR #1935, opened 2019-08-29
Remove eLanguageFlag which is not used anywhere

Teemperor:


Teemperor:
@swift-ci test

Teemperor:
Apparently we put changes into stable....

dcci:
looks like it's actually used :|

PR #1936, opened 2019-08-29
Remove eLanguageFlag which is not used anywhere

Teemperor:


Teemperor:
@swift-ci test

Teemperor:
Seems like it's only unused in upstream-with-swift and the thing was introduced in some swift-only clang refactoring in 5f3aa16203d308773108c4d03ed4d4ec8e4a47c2

dcci:
We should see whether we can get rid of that.

Teemperor:
Yeah, it seems that was done to kind of allow some form of unified setting thing between Clang and Swift. Working on it.

PR #1937, opened 2019-08-29
[ClangExpressionParser] Remove code that's not upstream.

dcci:
Checked with Adrian and this probably can go.


dcci:
@swift-ci test and merge

PR #1941, opened 2019-08-29
[Swift] Switch capture extension to the request model

CodaFi:
To support apple/swift#26932


CodaFi:
apple/swift#26932
@swift-ci Please test

PR #1949, opened 2019-09-02
[ClangExpressionParser] Revert duplicated file creation logic 

Teemperor:
We already create a temporary file directly below this code,
so this seems redundant.

PR #1953, opened 2019-09-03
[ExpressionSourceCode] Move Swift-only function SaveExpressionTextToTempFile to the Swift source.

Teemperor:


Teemperor:
@swift-ci test

Teemperor:
(the first commit is from #1949 where we remove the use of that function from the ClangExpressionParser)

adrian-prantl (approved):
That's a good idea!

dcci (approved):
LGTM

PR #1959, opened 2019-09-03
[CMake] Copy swift resource dir only when building standalone

bulbazord:
When building with a unified setup (lldb and swift are a part of the same build)
lldb should be able to use the swift libdir from the swift build.


compnerd (approved):

compnerd:
@swift-ci please test

PR #1964, opened 2019-09-04
DWARFImporter: Add a test for inheriting form an Objective-C class.

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

PR #1972, opened 2019-09-05
[Reproducer] Support Swift textual interfaces

JDevlieghere:
Support was already there for capturing Swift's textual interfaces. This
adds a test to make sure that they're used by the reproducer.

### DIFF
--- a/lit/Reproducer/Swift/TestSwiftInterface.test
+++ b/lit/Reproducer/Swift/TestSwiftInterface.test
@@ -1,0 +1,59 @@
+# Test that reproducers can deal with .swiftinterface files.
+
+# REQUIRES: system-darwin
+
+# Start clean.
+# RUN: rm -rf %t.build && mkdir %t.build && cd %t.build
+# RUN: rm -rf %t.lib && mkdir %t.lib
+# RUN: cp %S/Inputs/A.swift %t.build/AA.swift
+# RUN: cp %S/Inputs/Main.swift %t.build/Main.swift
+
+# RUN: %target-swiftc -module-name AA \
+# RUN:          -module-cache-path %t.build/cache \
+# RUN:          -enable-library-evolution \
+# RUN:          -emit-module-interface-path %t.lib/AA.swiftinterface \
+# RUN:          -emit-library \
+# RUN:          -o %t.lib/libAA%target-shared-library-suffix \
+# RUN:          %t.build/AA.swift
+
+# Regenerate the Swift module but with a different type.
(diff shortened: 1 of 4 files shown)


JDevlieghere:
@adrian-prantl Can you please confirm that this is sufficient to check that the .swiftinterfaces are used and/or if there's a better way? They're definitely in the reproducer, but the goal of the test is to ensure that they're also used by Swift.

adrian-prantl:
@nathawes recently added `settings set symbols.swift-module-loading-mode prefer-serialized` to force one loader over the other, but this may actually be the opposite of what you want to test if you want to test that the expected loader is chosen automatically. Not sure.

adrian-prantl:
What you want is logging that tells us through which swift::ModuleLoader the module was imported?

lit/Reproducer/Swift/TestSwiftInterface.test
```diff
@@ -0,0 +1,53 @@
...
+# Start clean.
+# RUN: rm -rf %t.build && mkdir %t.build && cd %t.build
+# RUN: rm -rf %t.lib && mkdir %t.lib
+# RUN: cp %S/Inputs/A.swift %t.build/AA.swift
+# RUN: cp %S/Inputs/Main.swift %t.build/Main.swift
+
+# RUN: %target-swiftc -module-name AA \
+# RUN:          -module-cache-path %t.build/cache \
+# RUN:          -enable-library-evolution \
+# RUN:          -emit-parseable-module-interface-path %t.lib/AA.swiftinterface \
```
harlanhaskins: Nit: Can you use `-emit-module-interface-path`? It's the final name of this flag.

lit/Reproducer/Swift/TestSwiftInterface.test
```diff
@@ -0,0 +1,53 @@
...
+# RUN:          -emit-library \
+# RUN:          -o %t.lib/libAA%target-shared-library-suffix \
```
harlanhaskins, line 16: This can be `%t.lib/%target-library-name(AA)`

harlanhaskins (approved):

harlanhaskins:
Small nits, but this test looks good

lit/Reproducer/Swift/TestSwiftInterface.test
JDevlieghere: I believe that's a Swift thing that's not available in LLDB. That said, maybe we should have it too :-)

nathawes (approved):
LGTM!

JDevlieghere:
@nathawes: Thanks a lot for the trick with the typename!

PR #1973, opened 2019-09-05
Backport fix for detecting illegal command options

Teemperor:
Fixes rdar://53280240 (swift-5.1 branch lldb doesn't filter out illegal command options)

Original test didn't have a test, so I back ported part of the test that is upstream.


Teemperor:
@swift-ci test

fredriss:
No need to pull this into 5.1. We're close to rebasing master on a newer upstream version which has the fix.

PR #1974, opened 2019-09-09
Long timeouts for the MacOSX SystemRuntime plugins under ASAN; else q…

JDevlieghere:
…uick.

In April via r357829, Adrian unified timeouts across lldb and set the
default value high so that we wouldn't get timeouts on ASAN bots that
were running under load.

The library that the MacOSX SystemRuntime has functions that need
to take a lock, and if that lock is held already, those functions
will never complete; we're seeing the 15 second timeout being hit
with inferiors that are doing a lot of enqueuing and dequeuing of
libdispatch work items causing this deadlocking behavior.

This patch reverts to a very short timeout for these SystemRuntime
function calls, given the behavior of this library that they are 
calling into.  When lldb is built with AddressSanitizer enabled,
they will use the default 15 second timeout.

tl;dr: this reverts to the previous timeouts for these SystemRuntime
inf func calls.

<rdar://problem/54538149> 

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@371280 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 99d2a3485de3849e6508b64937864fffe39c606c)

### DIFF
--- a/source/Plugins/SystemRuntime/MacOSX/AppleGetItemInfoHandler.cpp
+++ b/source/Plugins/SystemRuntime/MacOSX/AppleGetItemInfoHandler.cpp
@@ -333,6 +333,11 @@ AppleGetItemInfoHandler::GetItemInfo(Thread &thread, uint64_t item,
   options.SetUnwindOnError(true);
   options.SetIgnoreBreakpoints(true);
   options.SetStopOthers(true);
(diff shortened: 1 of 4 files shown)


JDevlieghere:

PR #1975, opened 2019-09-09
Fixing build break with upstream lldb (SwiftASTContext::GetFloatTypeSemantics)

plotfi:
SwiftASTContext::GetFloatTypeSemantics not implemented was causing build break.

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -6151,6 +6151,11 @@ bool SwiftASTContext::IsFixedSize(CompilerType compiler_type) {
   return false;
 }
 
+const llvm::fltSemantics &
+SwiftASTContext::GetFloatTypeSemantics(size_t byte_size) {
+  llvm_unreachable("SwiftASTContext::GetFloatTypeSemantics not implemented.");
+}
+
 llvm::Optional<uint64_t>
 SwiftASTContext::GetBitSize(lldb::opaque_compiler_type_t type,
                             ExecutionContextScope *exe_scope) {
--- a/include/lldb/Symbol/SwiftASTContext.h
+++ b/include/lldb/Symbol/SwiftASTContext.h
@@ -547,6 +547,8 @@ public:
 
   // Exploring the type
 
+  const llvm::fltSemantics &GetFloatTypeSemantics(size_t byte_size) override;
+
   llvm::Optional<uint64_t>
   GetBitSize(lldb::opaque_compiler_type_t type,
              ExecutionContextScope *exe_scope) override;


source/Symbol/SwiftASTContext.cpp
```diff
@@ -6152,6 +6152,19 @@ bool SwiftASTContext::IsFixedSize(CompilerType compiler_type) {
   return false;
 }
 
+const llvm::fltSemantics &
+SwiftASTContext::GetFloatTypeSemantics(size_t byte_size) {
+  const size_t bit_size = byte_size * 8;
+  switch (bit_size) {
+    default:
+      return llvm::APFloatBase::Bogus();
```
compnerd: This really should be an assertion failure.  There is no support in the frontend to emit the FP semantics.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -6152,6 +6152,19 @@ bool SwiftASTContext::IsFixedSize(CompilerType compiler_type) {
...
+    case 32: return llvm::APFloat::IEEEsingle();
+    case 64: return llvm::APFloat::IEEEdouble();
+    case 80: return llvm::APFloat::x87DoubleExtended();
+  }
+  return llvm::APFloatBase::Bogus();
```
compnerd: `llvm_unreachable`?  This is a covered switch return.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -6152,6 +6152,17 @@ bool SwiftASTContext::IsFixedSize(CompilerType compiler_type) {
   return false;
 }
 
+const llvm::fltSemantics &
+SwiftASTContext::GetFloatTypeSemantics(size_t byte_size) {
+  const size_t bit_size = byte_size * 8;
```
jrose-apple: This is highly suspect; IIRC Float80 is stored as 16 bytes. @stephentyrone?

compnerd: @jrose-apple, IIRC, the FP80 is backed by Builtin.FPIEEE80. so, the size should be correct?

compnerd:
@swift-ci please test

source/Symbol/SwiftASTContext.cpp
stephentyrone: @compnerd Depends quite a bit on how `byte_size` is derived in the caller. Where is this function used?

plotfi: @stephentyrone As far as I can tell, GetFloatTypeSemantics is only used in https://github.com/llvm/llvm-project/blob/master/lldb/source/Core/DumpDataExtractor.cpp

edit: But DumpDataExtractor() is used in a lot of places. This was added in llvm-svn r371258. ie https://reviews.llvm.org/D67239

compnerd: https://github.com/llvm/llvm-project/blob/master/lldb/source/Core/DumpDataExtractor.cpp#L573-L574

The backing for floating point should be 10-bytes (FP80) on x87 and 8-bytes or 16-bytes for single and double precision.

compnerd: Also, https://github.com/llvm/llvm-project/blob/master/lldb/source/Core/DumpDataExtractor.cpp#L561-L562 - we will never really come down this path.  Why?!  We should just move forward with this for now I think.  This is effectively dead code, but has reasonable behaviour.  Whenever someone comes around to adding support for this in Swift, we should be able to catch this.

CC: @xiaobai @dcci

plotfi: @JDevlieghere

JDevlieghere: If this is indeed dead code can we add an llvm_unreachable instead?

compnerd: @JDevlieghere - it is only dead for the time being (in some sense a bug in lldb).  I don't expect this to be dead forever, this should be something that should get implemented at some point - we would want to be able to render `Float` and `Double` in the Swift context properly.

plotfi: @JDevlieghere done

JDevlieghere (approved):

source/Symbol/SwiftASTContext.cpp
JDevlieghere: @compnerd Of course and when that time comes we should implement it correctly. I prefer the unreachable because it conveys the right signal.

PR #1977, opened 2019-09-09
[WIP/Don't touch] Use mirrors and type reconstruction for classes.

dcci:


source/Target/SwiftLanguageRuntime.cpp
```diff
@@ -1543,35 +1543,32 @@ bool SwiftLanguageRuntime::GetDynamicTypeAndAddress_Class(
...
-    }
-    return false;
-  }
+  Status error;
+
+  // FIXME: Use an API to retrieve the prefix instead of hardcoding it.
+  ConstString mangled_with_prefix("$s" + mangled);
+
+  SwiftASTContext *swift_ast_ctx = llvm::dyn_cast_or_null<SwiftASTContext>(
+      in_value.GetCompilerType().GetTypeSystem());
```
adrian-prantl, line 1566: Obviously this isn't meant to be production-ready yet, but: This should be the scratch context, not in_value's module context.

dcci: Indeed.

PR #1979, opened 2019-09-10
Porting 'Replace `lldb_private::CleanUp` by `llvm::scope_exit`' to swift-lldb.

plotfi:
llvm svn version r371474 changed CleanUp to scope_exit. This makes it so
the swift specific bits here build.

(cherry picked from commit 9c20e413a40c36033f096de94933ebdc2c042220)

### DIFF
--- a/source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
@@ -31,7 +31,7 @@
 #include "lldb/Target/Target.h"
 #include "lldb/Target/Thread.h"
 #include "lldb/Utility/AnsiTerminal.h"
-#include "lldb/Utility/CleanUp.h"
+#include "llvm/ADT/ScopeExit.h"
 
 #include "llvm/Support/raw_ostream.h"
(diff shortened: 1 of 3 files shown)


source/Plugins/ExpressionParser/Swift/SwiftREPL.cpp
```diff
@@ -213,10 +213,11 @@ lldb::REPLSP SwiftREPL::CreateInstanceFromDebugger(Status &err,
...
   // Destroy the process and the event handler thread after a fatal error.
-  CleanUp cleanup{[&] {
-    process_sp->Destroy(/*force_kill=*/false);
-    debugger.StopEventHandlerThread();
-  }};
+  auto cleanup =
+    llvm::make_scope_exit([&]() {
+        process_sp->Destroy(/*force_kill=*/false);
+        debugger.StopEventHandlerThread();
+        });
```
JDevlieghere: Is this clang-formatted?

plotfi: It is now. (I didn't have clang-format.py on the devbox I pushed this from :-( )

PR #1980, opened 2019-09-10
Remove the -static-stdlib from swift invocations when running

jasonmolenda:
the testsuite against and ios device.

Add builder_darwin.py to come up with the correct triple when
building swift programs, using the version number from the
current SDK.


jasonmolenda:
@swift-ci test

vedantk:
👍

Looks like there's some kind of bad merge in clang :/. Mishal never took away my merge permissions, ping me if this needs to be force-landed.

jasonmolenda:
thanks!  upstream-with-swift doesn't require tests to run before merging, so I can merge it.  I should do at least a mac native testsuite run on my desktop before I do though.

PR #1983, opened 2019-09-11
 Reverting because this was committed twice and is causing another build break.

plotfi:


plotfi:
@compnerd @vedantk @dcci @JDevlieghere

compnerd (approved):

PR #1985, opened 2019-09-12
[SwiftREPL] Re-enable ExclusivityREPL.

dcci:
I really can't reproduce the race locally. I'll try to capture a log from the bots when this fails again.


dcci:
@swift-ci test and merge

stephentyrone (approved):

PR #1989, opened 2019-09-16
regenerate static swig bindings

marcrasi:
These are checked-in swig-generated files. We have made some tensorflow branch specific changes to the swig interface, so these files must be regenerated.

These checked-in files are only important on machines that try to build lldb without swig installed, which is why the staleness has not affected us.

If you experience a merge conflict on these files, simply regenerate them.

For future reference, regeneration instructions:
1. Build a full toolchain.
2. cd to the lldb repo
3. `cp ../build/buildbot_linux/lldb-linux-x86_64/scripts/lldb.py ../build/buildbot_linux/lldb-linux-x86_64/scripts/LLDBWrapPython.cpp scripts/Python/static-binding/`


bgogul:
@marcrasi should we cherry-pick this for 0.5?

dan-zheng (approved):
LGTM if you've verified that `import lldb` works!

marcrasi:
No need to cherry-pick this. It should not affect the build results, just make it easier to build on machines that don't have swig installed.

I wouldn't expect this change to affect whether `import lldb` works (I don't expect it to fix or break it). I'll build locally and make sure that `import lldb` still works for me though.

PR #1991, opened 2019-09-16
Install a TypeChecker in the ASTContext

CodaFi:
Having the type checker requests available means that we actually do need a type checker to be installed here (for now).  Once the lazy resolver disappears, this should be re-evaluated.


CodaFi:
@swift-ci please test

CodaFi:
⛵️

PR #1993, opened 2019-09-17
[Swift] Infer the canonical type of the condition in breakpoints.

dcci:
lldb checks that the type of the condition is `Swift.Bool`.
Parentheses in the condition are carried in the type, e.g.
((cond)) will have type ((Swift.Bool)). Always desugar the
type as sugar doesn't matter for this check.

Fixes <rdar://problem/55372511>

### DIFF
--- a/lit/Swift/Inputs/main.swift
+++ b/lit/Swift/Inputs/main.swift
@@ -1,0 +1,3 @@
+func main() { return }
+
+_ = main()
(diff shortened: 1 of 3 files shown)


dcci:
@swift-ci test

JDevlieghere (approved):

dcci:
@swift-ci test linux

adrian-prantl (commented):
lgtm

PR #1995, opened 2019-09-17
[5.1][Swift] Infer the canonical type of the condition in breakpoints.

dcci:
lldb checks that the type of the condition is `Swift.Bool`.
Parentheses in the condition are carried in the type, e.g.
((cond)) will have type ((Swift.Bool)). Always desugar the
type as sugar doesn't matter for this check.

Fixes <rdar://problem/55372511>

### DIFF
--- a/lit/Swift/cond-breakpoint.test
+++ b/lit/Swift/cond-breakpoint.test
@@ -1,0 +1,10 @@
+# RUN: rm -rf %t && mkdir %t && cd %t
+# RUN: %target-swiftc -g %S/Inputs/main.swift -o a.out
+# RUN: %lldb a.out -b -s %s 2>&1 | FileCheck %s
+
+# CHECK-NOT: Stopped due to an error evaluating condition of breakpoint
+
+# The parentheses surrounding the expression matter.
+br s -n main -c '((nil == nil))'
+r
+c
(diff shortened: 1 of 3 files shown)


dcci:
@swift-ci test

lit/Swift/cond-breakpoint.test
```diff
@@ -0,0 +1,10 @@
+# RUN: rm -rf %t && mkdir %t && cd %t
+# RUN: %target-swiftc -g %S/Inputs/main.swift -o a.out
+# RUN: %lldb a.out -b -s %s 2>&1 | FileCheck %s
+
+# CHECK-NOT: Stopped due to an error evaluating condition of breakpoint
```
adrian-prantl, line 5: Can you add a positive CHECK, too? /bin/true would also pass this test :-)

adrian-prantl (approved):
Looks safe to me!

lit/Swift/cond-breakpoint.test
dcci: Do you have ideas on what to check exactly?

JDevlieghere (approved):
LGTM

lit/Swift/cond-breakpoint.test
JDevlieghere: You could match the stop at the breakpoint.

PR #1998, opened 2019-09-18
[SwiftInterface] Reenable these two tests on linux.

dcci:
Turns out the reason why they're failing was unrelated to
the nature of the tests themselves.

<rdar://problem/53312825>


dcci:
@nathawes let's give these another shot.

dcci:
@swift-ci test and merge

nathawes (approved):

PR #2000, opened 2019-09-18
If commands modules fails to load, use subprocess module

ematejska:
The commands module has been removed in Python3. Use subprocess instead if it fails to load as recommended by https://docs.python.org/2/library/commands.html.

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
@@ -12,7 +12,7 @@
 """
 Test `@_implementationOnly import` behind some indirection in a library used by the main executable
 """
-try
+try:
     import commands
 except ImportError:    
     import subprocess
--- a/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
+++ b/packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
@@ -12,14 +12,10 @@
 """
 Test `@_implementationOnly import` behind some indirection in a library used by the main executable
 """
-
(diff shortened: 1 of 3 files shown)


marcrasi:
Is this something we could do as an upstream patch instead? It seems like it would be useful there too.

ematejska:
Marc, good point. I'll try there first then.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
```diff
@@ -12,7 +12,14 @@
 """
 Test `@_implementationOnly import` behind some indirection in a library used by the main executable
 """
-import commands
+
+#SWIFT_TENSORFLOW: Since tensorflow uses Python 3 and commands is deprected
```
dan-zheng: Minor: our "`tensorflow`-specific change" marker is called `SWIFT_ENABLE_TENSORFLOW`.
```suggestion
# SWIFT_ENABLE_TENSORFLOW: Since tensorflow uses Python 3 and commands is deprecated
```

This won't be relevant if you submit the patch upstream. 🙂

bgogul:
> Marc, good point. I'll try there first then.

Actually, can we merge it into tensorflow branch first if this is working? This is blocking my work to enable toolchain tests when running our hourly toolchain builds.

packages/Python/lldbsuite/test/lang/swift/implementation_only_imports/library_indirect/TestLibraryIndirect.py
```diff
@@ -12,7 +12,14 @@
...
+# there, import subprocess instead.
+try
```
bgogul: this should be `try:`?

ematejska: For some reason it worked without it but adding it anyway for consistency.

ematejska:
Closing since the work for this was upstreamed here: https://github.com/apple/swift-lldb/pull/2010

PR #2001, opened 2019-09-18
[TestSwiftUninitializedVariable] Re-enable on Linux.

dcci:
There's nothing darwin specific here. Also, clean up unneeded
copyright and Makefile while I'm around.

<rdar://problem/55498143>

### DIFF
--- a/packages/Python/lldbsuite/test/lang/swift/variables/uninitialized/Makefile
+++ b/packages/Python/lldbsuite/test/lang/swift/variables/uninitialized/Makefile
@@ -1,7 +1,3 @@
 LEVEL = ../../../../make
 SWIFT_SOURCES := main.swift
 include $(LEVEL)/Makefile.rules
-
-cleanup:
-	rm -f Makefile *.d
-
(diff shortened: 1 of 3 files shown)


dcci:
@swift-ci test and merge

packages/Python/lldbsuite/test/lang/swift/variables/uninitialized/Makefile
```diff
@@ -1,7 +1,3 @@
 LEVEL = ../../../../make
 SWIFT_SOURCES := main.swift
 include $(LEVEL)/Makefile.rules
-
-cleanup:
```
dcci, line 5: @adrian-prantl I'm under the impression this is not needed anymore after you implemented a general `clean` rule in the top level makefile, but please correct me if I'm wrong.

adrian-prantl: This particular `rm Makefile` is a very odd rule.
The default rule will definitely clean any `.d` files.

adrian-prantl (approved):

PR #2006, opened 2019-09-20
Pass triple to swift link commands

fredriss:
Link commands need to be given the right triple, let's just
pass the full set of SWIFTFLAGS to them.


dcci (approved):
This looks fine -- but is this only an issue for the python tests? what about the lit ones?

fredriss:
> This looks fine -- but is this only an issue for the python tests? what about the lit ones?

This is only an issue when you cross-compile. I don't think our pure lit tests support cross-compilation at the moment?

dcci:
I don't think they do either. LGTM then.

dcci:
Maybe you can amend the commit message adding that this only shows up for cross-compilation? (mainly for archeology purposes).

fredriss:
@swift-ci test and merge

fredriss:
Looks like PR testing is not setup correctly on this branch, I'll just merge it.

adrian-prantl:
This is probably fine.

PR #2007, opened 2019-09-20
[test] Re-enable TestGenericError on Linux, improving coverage for ge…

dcci:
…nerics.

<rdar://problem/55566502>


dcci:
FWIW, I'm going through these before pushing the mirror patch on linux -- because I'm really afraid/nervous the low coverage on linux might mask real issues. In other words, we really can't iterate quickly until we have decent testing in place.

dcci:
@swift-ci test and merge

PR #2010, opened 2019-09-23
Remove importing commands module since it's not used

ematejska:
The commands module has been removed in Python3. Use subprocess instead if it fails to load as recommended by https://docs.python.org/2/library/commands.html.  Future proofing fixes for Python3.

Augmented this pull request to just remove importing the commands module since it's not used.


ematejska:
@swift-ci test

jrose-apple:
Oops. It looks like I'm not even using the import anyway; let's just remove it. (Do you want me to get that?)

ematejska:
I'm happy to augment the pull request to do that :)

ematejska:
@swift-ci please test

jrose-apple (approved):

PR #2012, opened 2019-09-23
Fix build

bulbazord:
This is a merge of llvm/master into upstream-with-swift along with any needed adjustments for swift.


bulbazord:
This was supposed to go into upstream-with-swift. :/

PR #2015, opened 2019-09-24
Prevent assignment to "let"-bound values in Swift expressions.

adrian-prantl:
Due to the way that SIL and IR is generated in the Swift compiler
assigning a new value to a let-binding often does not have the desired
effect since there may be many copies of the value in flight and LLDB
only knows about one of them per PC. Worse, sometimes the value known
to LLDB is only a shadow copy that is only used in the debugger.

rdar://problem/16042273

### DIFF
--- a/include/lldb/Symbol/Variable.h
+++ b/include/lldb/Symbol/Variable.h
@@ -30,11 +30,10 @@ public:
   Variable(lldb::user_id_t uid, const char *name,
            const char
                *mangled, // The mangled or fully qualified name of the variable.
-           const lldb::SymbolFileTypeSP &symfile_type_sp,
-           lldb::ValueType scope, SymbolContextScope *owner_scope,
-           const RangeList &scope_range, Declaration *decl,
-           const DWARFExpression &location, bool external, bool artificial,
-           bool static_member = false);
+           const lldb::SymbolFileTypeSP &symfile_type_sp, lldb::ValueType scope,
+           SymbolContextScope *owner_scope, const RangeList &scope_range,
+           Declaration *decl, const DWARFExpression &location, bool external,
+           bool artificial, bool static_member, bool constant);
 
   virtual ~Variable();
 
@@ -70,6 +69,8 @@ public:
   bool IsArtificial() const { return m_artificial; }
 
   bool IsStaticMember() const { return m_static_member; }
+
+  bool IsConstant() const { return m_constant; }
 
   DWARFExpression &LocationExpression() { return m_location; }
 
@@ -127,6 +128,10 @@ protected:
                                // location
       m_static_member : 1; // Non-zero if variable is static member of a class
                            // or struct.
+  /// Indicates whether the variable is a constant, for example, Swift \c let
+  /// binding.
+  uint8_t m_constant : 1;
+
 private:
   Variable(const Variable &rhs) = delete;
   Variable &operator=(const Variable &rhs) = delete;
--- a/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
+++ b/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
@@ -1156,6 +1156,8 @@ bool SwiftASTManipulator::AddExternalVariables(
         if (!m_wrapper_decl)
           continue;
 
+        // We need to mutate the $__lldb_wrapped_expr_%d member of self later.
+        introducer = swift::VarDecl::Introducer::Var;
         loc = m_wrapper_decl->getBody()->getLBraceLoc();
         containing_function = m_wrapper_decl;
       }
--- a/source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
+++ b/source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
@@ -3678,10 +3678,18 @@ VariableSP SymbolFileDWARF::ParseVariableDIE(const SymbolContext &sc,
                                type_sp->GetType()->GetByteSize().getValueOr(0),
                                die.GetCU()->GetAddressByteSize());
 
+        // Swift let-bindings are marked by a DW_TAG_const_type.
+        bool is_constant = false;
+        if (IsSwiftLanguage(sc.comp_unit->GetLanguage())) {
+          DWARFDIE type_die = die.GetReferencedDIE(llvm::dwarf::DW_AT_type);
+          if (type_die && type_die.Tag() == llvm::dwarf::DW_TAG_const_type)
+            is_constant = true;
+        }
+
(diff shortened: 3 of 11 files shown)


adrian-prantl:
test with https://github.com/apple/swift/pull/27345

adrian-prantl:
test with https://github.com/apple/swift/pull/27345
@swift-ci test

include/lldb/Symbol/Variable.h
```diff
@@ -74,6 +73,8 @@ class Variable : public UserID, public std::enable_shared_from_this<Variable> {
 
   bool IsStaticMember() const { return m_static_member; }
 
+  bool IsConstant() const { return m_constant; }
+
```
adrian-prantl, line 74: There is no reason to not upstream these changes to llvm.org.

adrian-prantl:
test with https://github.com/apple/swift/pull/27345
@swift-ci test

jimingham:
Most of this is actually reformatting of extant code, which seems all fine.  The part where you are getting the constant value and marking it in the Variable is also pretty clear and seems right.

I didn't understand why you changed the type of the return that we gather from Var to Let (SwiftASTManipulator.cpp:420 or thereabouts).  Was that a drive-by cleanup or was it necessary for this patch?

adrian-prantl:
> Was that a drive-by cleanup or was it necessary for this patch?

That's a drive-by cleanup that I now rolled into https://github.com/apple/swift-lldb/pull/2016

jimingham (approved):
LGTM

source/Plugins/SymbolFile/PDB/SymbolFilePDB.cpp
```diff
@@ -957,7 +957,8 @@ VariableSP SymbolFilePDB::ParseVariableForPDBData(
 
   var_sp = std::make_shared<Variable>(
       var_uid, var_name.c_str(), mangled_cstr, type_sp, scope, context_scope,
-      ranges, &decl, location, is_external, is_artificial, is_static_member);
+      ranges, &decl, location, is_external, is_artificial, is_static_member,
```
dcci, line 1023: We should make this an enum.

source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
```diff
@@ -3715,10 +3715,18 @@ VariableSP SymbolFileDWARF::ParseVariableDIE(const SymbolContext &sc,
               type_sp->GetType()->GetByteSize().getValueOr(0),
               die.GetCU()->GetAddressByteSize());
 
+        // Swift let-bindings are marked by a DW_TAG_const_type.
+        bool is_constant = false;
+        if (IsSwiftLanguage(sc.comp_unit->GetLanguage())) {
+          DWARFDIE type_die = die.GetReferencedDIE(llvm::dwarf::DW_AT_type);
```
dcci, line 3684: I don't really like this. It creeps into a generic file. Can you think of a way to make this swift-specific?

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
```diff
@@ -1155,6 +1155,8 @@ bool SwiftASTManipulator::AddExternalVariables(
         if (!m_wrapper_decl)
           continue;
 
+        // We need to mutate the $__lldb_wrapped_expr_%d member of self later.
```
dcci, line 1159: Can you elaborate this?

dcci (changes requested):
Great work! Some comments inline.

source/Plugins/SymbolFile/PDB/SymbolFilePDB.cpp
adrian-prantl: Agreed, but not in swift-lldb. I'll do this when I'm upstreaming this patch.

source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp
adrian-prantl: The code in https://github.com/apple/swift-lldb/blob/5db61e041f7b3d5e84bce27966e32b351c648946/source/Plugins/ExpressionParser/Swift/SwiftASTManipulator.cpp#L196 doesn't compile without making `$__lldb_injected_self` mutable.

source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
adrian-prantl: We don't have a good plugin mechanism for that at the moment. There are already 5 swift-specific extensions in this file in various places (e.g., https://github.com/apple/swift-lldb/blob/5db61e041f7b3d5e84bce27966e32b351c648946/source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp#L3668)

adrian-prantl: We could conceivably add a language-specific `IsConstant()` method here, but which object should it be on?

dcci:
Adrian, we can discuss this more in person but you know that we're going to great lengths to make sure generic files in swift-lldb aren't polluted with swift specific support. I don't want to hold this back for too long because it's really good and sorely needed work, but we both agree that this time is probably the right one to design a plugin-style mechanism for DWARF [or, whatever allows us to hide this].

jimingham:
At present, the boundary for the language specific plugin to DWARF parsing is the DWARFASTParser.  We assumed when we split the Swift/Clang DWARF parsing code up that parsing types would be language specific, and should live in separate files, but parsing block and variable information was generic, and didn't need to be split up into separate plugins.  

I note there are a couple of places where we do Clang specific stuff in SymbolFileDWARF (look for used of DWARFASTParserClang) so that separation is not as clean as it should be.

If this is something that could fit into the DWARFASTParser interface, that would be a natural home for it.  But it's clear that this separation needs a little more work, since we have to violate it from time to time as well...

Jim

source/Plugins/SymbolFile/DWARF/SymbolFileDWARF.cpp
jrose-apple: Arguably doing this for *any* declared-as-const storage is not guaranteed to have the desired effect in most languages, including C/C++.

dcci:
I do understand there was need to violate this in the past. I'm just nervous that we have a project that's supposed to clean these instances and this patch goes in the other direction.

dcci:
We discussed this offline and turns out there's no easy way to prevent swift-specific code to leak. We should re-evaluate this in the future, but for now I think this can go in.

dcci (approved):

PR #2020, opened 2019-09-27
[SwiftASTContext] Remove serialized options fallback.

dcci:
There's a very complex logic in SwiftASTContext to deal with the
case the compiler options aren't serialized that look at
apple_flags in DWARF and tries to reconstruct the compile invocations.

This doesn't work, as it's been shown by every project that
disables serialization of debugging options and reports
that "debugging doesn't work".

We should just nuke this code.

<rdar://problem/55792430>

### DIFF
--- a/source/Symbol/SwiftASTContext.cpp
+++ b/source/Symbol/SwiftASTContext.cpp
@@ -1765,58 +1765,6 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
         swift_ast_sp->SetPlatformSDKPath(serialized_sdk_path);
       }
     }
-
-    if (!got_serialized_options || swift_ast_sp->GetPlatformSDKPath().empty()) {
-      std::string platform_sdk_path;
-      if (sym_vendor->GetCompileOption("-sdk", platform_sdk_path)) {
-        FileSpec sdk_spec(platform_sdk_path.c_str());
-        if (FileSystem::Instance().Exists(sdk_spec)) {
-          swift_ast_sp->SetPlatformSDKPath(platform_sdk_path);
-        }
-
-        if (sym_vendor->GetCompileOption("-target", target_triple)) {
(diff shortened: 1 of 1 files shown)


dcci:
@swift-ci test

dcci:
@adrian-prantl ?

adrian-prantl:
I think there is more code that becomes dead once you remove this, like GetCompileOption.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -1765,58 +1765,6 @@ lldb::TypeSystemSP SwiftASTContext::CreateInstance(lldb::LanguageType language,
         swift_ast_sp->SetPlatformSDKPath(serialized_sdk_path);
       }
     }
-
-    if (!got_serialized_options || swift_ast_sp->GetPlatformSDKPath().empty()) {
```
adrian-prantl, line 1769: We need to at the very least log an error to the types log if (!got_serialized_options). It shouldn't be shown to the user necessarily because they may have compiled without serialized debug options.

dcci: Fair.

adrian-prantl: That's more a question than a request...

dcci:
I think GetCompileOption is not dead yet. there's another use in the Expression parser.

dcci:
```
source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp:          if (symbol_vendor->GetCompileOption("-private-discriminator",
```

I wonder what this is used for.

adrian-prantl:
> ```
> source/Plugins/ExpressionParser/Swift/SwiftExpressionParser.cpp:          if (symbol_vendor->GetCompileOption("-private-discriminator",
> ```
> 
> 
> I wonder what this is used for.

That's the *old* way of communicating the private discriminator for fileprivate types. We now put that into an inline DW_TAG_namespace in the compiler, but the LLDB support apparently hasn't been implemented yet.

PR #2021, opened 2019-09-28
[AST] Ensure extensions are bound.

davidungar:
Companion to https://github.com/apple/swift/pull/27313

In order to find compiler bugs resulting from not binding extensions, the above PR adds some assertion-checking, but since lldb does not go through normal channels, it must do a bit of extra work to ensure that extensions are bound.


davidungar:
@slavapestov To get the new assertion to work, I had to change lldb. What do you think?

davidungar:
@swift-ci please test

davidungar:
https://github.com/apple/swift/pull/27313

@swift-ci please test

PR #2029, opened 2019-10-02
build: normalise the path before use (NFC)

compnerd:
Normalise the path before using it for the inclusion, otherwise Windows
paths can break.


compnerd:
CC: @JDevlieghere

compnerd:

JDevlieghere (approved):

PR #2031, opened 2019-10-04
Convert test to import dispatch to a lit test

bulbazord:
The tests to import Cocoa and Foundation are both lit tests. Additionally, the
current import-dispatch test assumes the location of lldb which fails miserably
when you don't build lldb standalone.


bulbazord:
@swift-ci please test

bulbazord:
cc @JDevlieghere @compnerd @dcci

JDevlieghere (approved):

PR #2032, opened 2019-10-04
[CMake] Introduce LLDB_ENABLE_SWIFT_SUPPORT.

dcci:
Allows `lldb-tblgen` to be built without a swift
checkout.

<rdar://problem/55916729>

### DIFF
--- a/cmake/modules/LLDBConfig.cmake
+++ b/cmake/modules/LLDBConfig.cmake
@@ -52,6 +52,10 @@ option(LLDB_BUILD_FRAMEWORK "Build LLDB.framework (Darwin only)" OFF)
 option(LLDB_NO_INSTALL_DEFAULT_RPATH "Disable default RPATH settings in binaries" OFF)
 option(LLDB_USE_SYSTEM_DEBUGSERVER "Use the system's debugserver for testing (Darwin only)." OFF)
 
+# BEGIN SWIFT MOD
+option(LLDB_ENABLE_SWIFT_SUPPORT "Enable swift support" ON)
+# END SWIFT MOD
+
 # BEGIN SWIFT CODE
 option(LLDB_ALLOW_STATIC_BINDINGS "Enable using static/baked language bindings if swig is not present." OFF)
 
--- a/source/API/CMakeLists.txt
+++ b/source/API/CMakeLists.txt
@@ -1,7 +1,10 @@
 set(LLVM_NO_RTTI 1)
 
 include(AddLLVM)
-include(SwiftAddCustomCommandTarget)
+
+if (LLDB_ENABLE_SWIFT_SUPPORT)
+  include(SwiftAddCustomCommandTarget)
+endif()
 
 if ( CMAKE_SYSTEM_NAME MATCHES "Windows" )
   add_definitions( -DEXPORT_LIBLLDB )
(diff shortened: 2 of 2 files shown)


dcci:
@swift-ci test

cmake/modules/LLDBConfig.cmake
```diff
@@ -52,6 +52,10 @@ option(LLDB_BUILD_FRAMEWORK "Build LLDB.framework (Darwin only)" OFF)
 option(LLDB_NO_INSTALL_DEFAULT_RPATH "Disable default RPATH settings in binaries" OFF)
 option(LLDB_USE_SYSTEM_DEBUGSERVER "Use the system's debugserver for testing (Darwin only)." OFF)
 
+# BEGIN SWIFT MOD
+option(LLDB_ENABLE_SWIFT_SUPPORT "Enable swift support" ON)
```
bulbazord, line 56: It might be a good idea to do some error checking here. Specifically, if LLDB_ENABLE_SWIFT_SUPPORT is ON but we can't find swift, then throw an error.

source/API/CMakeLists.txt
```diff
@@ -199,13 +202,15 @@ if(NOT LLDB_BUILT_STANDALONE)
...
 # Copy the clang resource directory.
+if (LLDB_ENABLE_SWIFT_SUPPORT)
 add_custom_command_target(
     copy-clang-resource-dir
     COMMAND "${CMAKE_COMMAND}" "-E" "copy_directory" "${CLANG_RESOURCE_PATH}" "${lib_dir}/lldb/clang/"
     OUTPUT "${lib_dir}/lldb/clang/"
     VERBATIM
     ALL
     DEPENDS ${clang_headers_target})
+endif()
```
bulbazord: More than just this needs to be guarded. Below is code to copy the swift resource dir as well, so that should be guarded too.

bulbazord (changes requested):
Excellent. I love it. I'm curious though, does this actually work as-is or is this more preparatory?

dcci:
Hi Alex! This currently works for lldb-tblgen [which is really useful to build for cross-compiling lldb itself] but presumably works for other targets too. The goal is to eventually use this to guard the whole swift support, but we're far away from that. Does this make sense?

bulbazord:
Makes sense to me! Thanks for the explanation. Looking forward to seeing where this goes.

dcci:
@swift-ci test

source/API/CMakeLists.txt
```diff
@@ -228,9 +233,10 @@ install(
   DIRECTORY "${lib_dir}/lldb/clang"
   DESTINATION lib${LLVM_LIBDIR_SUFFIX}/lldb/)
 
+if(LLDB_ENABLE_SWIFT_SUPPORT)
```
bulbazord, line 236: nit: you can merge these two conditions

```
if (LLDB_ENABLE_SWIFT_SUPPORT AND LLDB_BUILT_STANDALONE)
```

bulbazord (approved):
Looks fine to me. Looking forward to what this opens up for us. :)

source/API/CMakeLists.txt
compnerd: As long as the entire block is local, that makes sense.  Otherwise, it is nice to leave the upstream code as is.

bulbazord: That block is local

PR #2034, opened 2019-10-07
[debugserver] Don't link against Cocoa, it's not needed.

dcci:
<rdar://problem/55916729>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@373795 91177308-0d34-0410-b5e6-96231b3b80d8


dcci:
@swift-ci test and merge

PR #2036, opened 2019-10-08
[CMake] Set repl_swift's rpath using SWIFT_LIBRARY_DIR

bulbazord:
This is useful because it allows us to not assume the location of the swift
build but rather rely on what SwiftConfig.cmake tells us.


bulbazord:
Please test with the following PRs:
https://github.com/apple/swift/pull/27570

@swift-ci please test

compnerd (approved):

PR #2037, opened 2019-10-08
Display enum bitfields

fredriss:


fredriss:
@swift-ci test and merge

fredriss:
@swift-ci test and merge

source/Symbol/ClangASTContext.cpp
```diff
@@ -9486,6 +9486,32 @@ void ClangASTContext::DumpValue(
...
+      llvm::cast<clang::EnumType>(qual_type.getTypePtr());
+  const clang::EnumDecl *enum_decl = enutype->getDecl();
+  assert(enum_decl);
+  clang::EnumDecl::enumerator_iterator enum_pos, enum_end_pos;
+  const bool is_signed = qual_type->isSignedIntegerOrEnumerationType();
+  lldb::offset_t offset = byte_offset;
+  const int64_t enum_value = data.GetMaxS64Bitfield(
+      &offset, byte_size, bitfield_bit_size, bitfield_bit_offset);
+
+  for (auto enumerator : enum_decl->enumerators()) {
```
adrian-prantl, line 9510: We could do something similar like https://github.com/apple/swift-lldb/blob/stable/source/Plugins/Language/Swift/SwiftOptionSet.cpp to show bitfield enums as `e0 | e1 | e2`.

packages/Python/lldbsuite/test/lang/c/enum_types/TestEnumTypes.py
```diff
@@ -25,11 +25,35 @@ def test(self):
...
+                    patterns=[' = ALL$'])
+        # Test that an enum that doesn't match the heuristic we use in
+        # ClangASTContext::DumpEnumValue, gets printed as a raw integer.
+        self.expect("fr var omega", DATA_TYPES_DISPLAYED_CORRECTLY,
+                    patterns=[' = 7$'])
+        # Test the behavior in case have a variable of a type considered
+        # 'bitfield' by the heuristic, but the value isn't actually fully
+        # covered by the enumerators.
+        self.expect("p (enum bitfield)nonsense", DATA_TYPES_DISPLAYED_CORRECTLY,
+                    patterns=[' = B | C | 0x10$'])
```
adrian-prantl, line 51: ah wait, we do.

source/Symbol/ClangASTContext.cpp
adrian-prantl: (I was looking at the first commit only when I wrote this.)

source/Symbol/ClangASTContext.cpp
```diff
@@ -9486,6 +9486,82 @@ void ClangASTContext::DumpValue(
...
+  std::vector<std::pair<uint64_t, llvm::StringRef>> values;
+  values.reserve(num_enumerators);
+  for (auto enumerator : enum_decl->enumerators())
+    if (auto val = enumerator->getInitVal().getZExtValue())
+      values.emplace_back(val, enumerator->getName());
+
+  // Sort in reverse order of the number of the population count,  so that in
+  // `enum {A, B, ALL = A|B }` we visit ALL first. Use a stable sort so that
+  // A | C where A is declared before C is displayed in this order.
+  std::stable_sort(values.begin(), values.end(), [](const auto &a, const auto &b) {
```
adrian-prantl, line 9545: This actually looks better than what SwiftOptionSet does. Could we reuse it there?

fredriss: No idea. Feel free to steal it :-)

dcci: Yes, OptionSet should use something like this, if possible at all. The formatter is really old [predates our time on lldb] so maybe it's time for a revamp.

PR #2039, opened 2019-10-08
[Testsuite] Get rid of most of the recursive shared library Makefiles

fredriss:
Most of the secondary Makefiles we have are just a couple variable
definitions and then an include of Makefile.rules. This patch removes
most of the secondary Makefiles and replaces them with a direct
invocation of Makefile.rules in the main Makefile. The specificities
of each sub-build are listed right there on the recursive $(MAKE)
call. All the variables that matter are being passed automagically by
make as they have been passed on the command line. The only things you
need to specify are the variables customizating the Makefile.rules
logic for each image.

This patch also removes most of the clean logic from those Makefiles
and from Makefile.rules. The clean rule is not required anymore now
that we run the testsuite in a separate build directory that is wiped
with each run. The patch leaves a very crude version of clean in
Makefile.rules which removes everything inside of $(BUILDDIR). It does
this only when the $(BUILDDIR) looks like a sub-directory of our
standard testsuite build directory to be extra safe.

Reviewers: aprantl, labath

Subscribers: lldb-commits

Tags: #lldb

Differential Revision: https://reviews.llvm.org/D68558

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@374076 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit a105d87cb9ab09a24b73edfc1bf23505ebf03f65)


fredriss:
@swift-ci test and merge

PR #2044, opened 2019-10-09
Fix Merge of rG41ff39605ea1: Add pretty printin...

gmittert:
Upstream lldb doesn't have the base class param so it wasn't in the call


JDevlieghere:
Thanks!

PR #2046, opened 2019-10-10
Change debugserver to use the brk #0 for breakpoints.

jasonmolenda:
Change debugserver to use the brk #0 for breakpoints.
debugserver had been using an instruction that would work
for armv7 or aarch64 processes, but we don't have armv7 code
running on arm64 devices any more so this is unnecessary.

<rdar://problem/56133118>

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@374264 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit fdebaf769010c566ad3050b20894e9a3b0d4510d)


jasonmolenda:
@swift-ci test and merge

PR #2049, opened 2019-10-10
[lldb] Get the TargetAPI lock in SBProcess::IsInstrumentationRuntimePresent (backport of D67831)

Teemperor:
We should get the TargetAPI lock here to prevent the process of being destroyed while we are in the function. Thanks Jim for explaining what's going on.

Fixes rdar://54424754


Teemperor:
@swift-ci test

dcci (approved):

PR #2050, opened 2019-10-10
[lldb][NFC] Remove strange bool parameter from Searcher::SearchCallba…

Teemperor:
…ck (testing)

Just testing what needs to be updated in Swift when this lands.


Teemperor:
@swift-ci test

PR #2056, opened 2019-10-11
Improve documentation of module loaders.

adrian-prantl:


adrian-prantl:
@swift-ci test and merge

varungandhi-apple (changes requested):
I think I made a mistake in recollecting what swiftc does when we were talking. If you look [here](https://github.com/apple/swift/blob/db902a19cdc84c47014845b015deaa7eea4fcb09/lib/Frontend/Frontend.cpp#L356-L383), we use the same ordering that swift-lldb does -- first the ModuleInterfaceLoader, then the SerializedModuleLoader, then the ClangImporter. I will add a doc comment to swiftc's code. You can probably note here that the ordering is the _same_ as swiftc.

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3432,11 +3433,13 @@ swift::ASTContext *SwiftASTContext::GetASTContext() {
 
   // 2. Create and install the module interface loader.
   //
-  // TODO: It may be nice to reverse the order between MIL and SML in
-  //       LLDB, since binary swift modules likely contain private
-  //       types that the module interfaces are missing. On the
-  //       other hand if we need to go looking for a module on disk,
-  //       something is already screwed up in the debug info.
+  // This order is different from the Swift compiler, where the
+  // serialized module loader comes before the module interface
```
harlanhaskins: Hm, in swiftc the module interface loader _does_ come before the serialized loader...

adrian-prantl:
@swift-ci test and merge

source/Symbol/SwiftASTContext.cpp
```diff
@@ -3432,11 +3433,14 @@ swift::ASTContext *SwiftASTContext::GetASTContext() {
 
   // 2. Create and install the module interface loader.
   //
-  // TODO: It may be nice to reverse the order between MIL and SML in
-  //       LLDB, since binary swift modules likely contain private
-  //       types that the module interfaces are missing. On the
-  //       other hand if we need to go looking for a module on disk,
-  //       something is already screwed up in the debug info.
+  // The ordering of 2-4 is the same as the Swift compiler's 1-3,
```
varungandhi-apple, line 3436: For swiftc, ModuleInterfaceLoader is number 3 and ClangImporter is number 5. We might have the SourceLoader and (MemoryBufferSerializedModuleLoader for lldb) come before.

PR #2060, opened 2019-10-14
Extract additional triple details from serialized Swift ASTs.

adrian-prantl:
Unfortunately things like the vendor name can affect Swift's ability
to find the correct stdlib.

<rdar://problem/45766181>


adrian-prantl:
@swift-ci test and merge

PR #2065, opened 2019-10-16
[CMake] Use variables exported by SwiftConfig to compose LLDB variables.

JDevlieghere:
The SwiftConfig.cmake file exports variables that point to the swift
binary and library directory. Use those instead of going through the
LLVM_LIBRARY_OUTPUT_INTDIR indirection, which might be incorrect, for
example for out-of-tree builds with the Xcode generator.


JDevlieghere:

dcci:
Yeah, I think this one is more correct. Good catch.

dcci (approved):

bulbazord:
Oh I didn't even notice we were doing this. Thanks for cleaning this up! :)

PR #2068, opened 2019-10-16
[CMake] Make it possible to set the RPATH directly in add_lldb_exectable

JDevlieghere:
Make it possible to pass a build and install RPATH to
add_lldb_executable instead of having to call lldb_setup_rpaths after
the fact.

This fixes a real issue where setting an install RPATH with
lldb_setup_rpaths would only affect the symroot installation component.
Given that lldb_setup_rpaths sets a target property I would expect this
to be orthogonal to installation components, but maybe I'm wrong.


JDevlieghere:

PR #2069, opened 2019-10-17
[CMake] Make it possible to set the RPATH in add_lldb_exectable.

JDevlieghere:
Make it possible to pass a build and install RPATH to
add_lldb_executable instead of having to call lldb_setup_rpaths after
the fact.

This fixes a real issue where setting an install RPATH with
lldb_setup_rpaths would only affect the symroot installation component.
Given that lldb_setup_rpaths sets a target property I would expect this
to be orthogonal to installation components. Regardless, it makes sense
to integrate this functionality in add_lldb_exectable.

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@375068 91177308-0d34-0410-b5e6-96231b3b80d8
(cherry picked from commit 0f53352cb22da3e0bd5e4d4272e9849e9883f11e)

### DIFF
--- a/cmake/modules/AddLLDB.cmake
+++ b/cmake/modules/AddLLDB.cmake
@@ -142,7 +142,7 @@ function(add_lldb_executable name)
   cmake_parse_arguments(ARG
     "GENERATE_INSTALL"
     "INSTALL_PREFIX;ENTITLEMENTS"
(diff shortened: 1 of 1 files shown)


JDevlieghere:

PR #2070, opened 2019-10-17
[CMake] Support unified builds

smeenai:
When we're doing a unified build (i.e. building Swift as an LLVM
subproject) with the monorepo, Swift won't necessarily live beside lldb.
Use LLVM_EXTERNAL_SWIFT_SOURCE_DIR to determine its location, but fall
back to checking for a sibling directory for non-unified builds.


smeenai:
CC @compnerd @dcci @JDevlieghere @xiaobai 

I'll need someone to merge this for me.

PR #2071, opened 2019-10-17
Backport "[lldb] Don't emit artificial constructor declarations as global functions"

Teemperor:
Summary:
When we have a artificial constructor DIE, we currently create from that a global function with the name of that class.
That ends up causing a bunch of funny errors such as "must use 'struct' tag to refer to type 'Foo' in this scope" when
doing `Foo f`. Also causes that constructing a class via `Foo()` actually just calls that global function.

The fix is that when we have an artificial method decl, we always treat it as handled even if we don't create a CXXMethodDecl
for it (which we never do for artificial methods at the moment).

Fixes rdar://55757491 and probably some other radars.

Reviewers: aprantl, vsk, shafik

Reviewed By: aprantl

Subscribers: jingham, shafik, labath, JDevlieghere, lldb-commits

Tags: #lldb

Differential Revision: https://reviews.llvm.org/D68130

git-svn-id: https://llvm.org/svn/llvm-project/lldb/trunk@375151 91177308-0d34-0410-b5e6-96231b3b80d8


Teemperor:
@swift-ci test

dcci (approved):

PR #2072, opened 2019-10-17
merge llvm.org into upstream-with-swift

adrian-prantl:


hyp:
Thanks, I merged it in manually

